PR #4, opened 2021-11-19
Start to adapt the project to a more common/OSS friendly setup.

stellaraccident:
* Adds a configure.py script to get everything right.
* Enables an optional dependency on iree-dialects (will register dialects and passes in the -opt binary and adds Python bindings).
* Adds a .env file that can be sourced to use built Python libs (also works for IDEs).
* Moves a couple of files which were in the wrong place.
* Disables the in-tree LinalgExt, which has no build support.
* Adds a basic CI action.


README.md
```diff
@@ -20,26 +20,24 @@ for more information.
 
 This project builds as part of the LLVM External Projects facility (see
 documentation for the `LLVM_EXTERNAL_PROJECTS` config setting). There are many
-ways to set this up but the following is recommended. It is left to the reader
-to adapt paths if deviating. We assume below that projects are checked out to
-`$HOME/src`.
+ways to set this up. We recommend using our `configure.py` script below.
+
+It is left to the reader to adapt paths if deviating. We assume below that
+projects are checked out to `$HOME/src`.
```
nicolasvasilache, line 26: Could we maybe invert the dep. chain here?

Basically ask the user to define the `LLVM_SOURCE_DIR`, `IREE_LLVM_SANDBOX_SOURCE_DIR` and `IREE_LLVM_SANDBOX_BUILD_DIR` and merely give as an example the following:

*   `LLVM_SOURCE_DIR`: $HOME/src/llvm-project
*   `IREE_LLVM_SANDBOX_SOURCE_DIR`: $HOME/src/iree-llvm-sandbox
*   `IREE_LLVM_SANDBOX_BUILD_DIR`: $HOME/src/iree-llvm-sandbox/build

Then our instructions should never mention the `$HOME` and `src magic constants anymore.

README.md
```diff
@@ -51,53 +49,50 @@ which python
...
+accessing IREE specific IR and evaluating on IREE compatible targets):
+
+### Building with IREE.
+
+Checkout the [IREE](https://github.com/google/iree) GitHub repo next to this
+directory and initialize submodules:
 
 ```
-pip3 install --pre torch torchvision torchaudio -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+(cd .. && git clone https://github.com/google/iree --recurse-submodules=third_party/llvm-project)
```
nicolasvasilache, line 66: Absolute paths based on the 3 env vars above would be nicer.

README.md
```diff
@@ -51,53 +49,50 @@ which python
...
-  -DLLVM_TARGETS_TO_BUILD="X86;NVPTX" \
-  -DMLIR_INCLUDE_INTEGRATION_TESTS=ON \
-  -DLLVM_ENABLE_ASSERTIONS=ON \
-  -DLLVM_INCLUDE_UTILS=ON \
-  -DLLVM_INSTALL_UTILS=ON \
-  -DLLVM_BUILD_EXAMPLES=ON \
-  -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
-  -DPython3_EXECUTABLE=$(which python) \
-  -DCMAKE_BUILD_TYPE=Release
+python configure.py --iree-path=../iree
```
nicolasvasilache, line 72: Absolute paths based on the 3 env vars above would be nicer.

README.md
```diff
@@ -51,53 +49,50 @@ which python
...
-Useful build commands (from the `${IREE_LLVM_SANDBOX_BUILD_DIR}` directory):
+You must checkout [llvm-project](https://github.com/llvm/llvm-project) at a
+compatible commit next to this directory.
 
-*   `ninja check-mlir`: Run MLIR tests.
-*   `ninja all`: Build everything.
-*   `ninja IREELLVMSandboxPythonModules`: Just build the python packages
-    (smaller/faster than everything).
+```
+(cd .. && git clone https://github.com/llvm/llvm-project.git)
```
nicolasvasilache, line 83: Absolute paths based on the 3 env vars above would be nicer.

README.md
```diff
@@ -51,53 +49,50 @@ which python
...
+
+```
+python configure.py
+```
 
 ## Using the Python API
 
 ```
-export PYTHONPATH=${IREE_LLVM_SANDBOX_BUILD_DIR}/tools/iree_llvm_sandbox/python_package
+source .env && export PYTHONPATH
```
nicolasvasilache, line 95: Absolute paths based on the 3 env vars above would be nicer.

.github/workflows/buildAndTest.yml
```diff
@@ -0,0 +1,59 @@
...
+      uses: llvm/actions/install-ninja@55d844821959226fab4911f96f37071c1d4c3268
+    - name: Ccache for C++ compilation
+      uses: hendrikmuhs/ccache-action@4687d037e4d7cf725512d9b819137a3af34d39b3
+      with:
+        key: ${{ runner.os }}-buildtestasserts
+
+    - name: Build
+      run: |
+        cd $GITHUB_WORKSPACE/sandbox
+        python configure.py --iree-path=../iree --llvm-path=../llvm-project
```
nicolasvasilache, line 59: export the recommended paths explicitly and use them here please?

nicolasvasilache: so this can also serve as a minimal repro for anyone who lands here

tools/mlir-proto-opt/mlir-proto-opt.cpp
```diff
@@ -10,9 +10,9 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include "./lib/CAPI.h"
-#include "include/LinalgExt/LinalgExtDialect.h"
-#include "include/LinalgExt/Passes.h"
+#include "CAPI.h"
+//#include "include/LinalgExt/LinalgExtDialect.h"
```
nicolasvasilache, line 14: QQ as to why this is needed ?
Not a blocker, we can iterating after landing this.

configure.py
```diff
@@ -0,0 +1,137 @@
...
+  if lld_path:
+    if has_clang:
+      print(f"-- Using LLD: {lld_path}")
+      llvm_configure_args.append(f"-DLLVM_USE_LINKER={lld_path}")
+    else:
+      print("WARNING: Can only use lld with clang")
+
+  # Detect ccache.
+  ccache_path = shutil.which("ccache")
+  if ccache_path and args.enable_ccache:
```
nicolasvasilache, line 97: should we warn if building without LLD or CCACHE?
I wasn't personally versed with those and now I'd almost consider it a bug not using it.

I usually just go locally with the following in the cmake config:
`-DLLVM_ENABLE_LLD=ON -DLLVM_CCACHE_BUILD=ON`

do we envision people need to bring their own ?

configure.py
```diff
@@ -0,0 +1,137 @@
...
+    llvm_configure_args.append(f"-DCMAKE_C_COMPILER={clang_path}")
+    llvm_configure_args.append(f"-DCMAKE_CXX_COMPILER={clangpp_path}")
+    has_clang = True
+  else:
+    print(
+        "WARNING: Could not find clang. Building with default system compiler")
+
+  # Detect lld.
+  lld_path = shutil.which("ld.lld")
+  if lld_path:
```
nicolasvasilache, line 88: should we warn if building without LLD or CCACHE?
I wasn't personally versed with those and now I'd almost consider it a bug not using it.

I usually just go locally with the following in the cmake config:
`-DLLVM_ENABLE_LLD=ON -DLLVM_CCACHE_BUILD=ON`

do we envision people need to bring their own ?

configure.py
```diff
@@ -0,0 +1,137 @@
...
+
+  # CMake configure.
+  build_dir = os.path.abspath(args.build_dir)
+  os.makedirs(build_dir, exist_ok=True)
+  cmake_args = [
+      "cmake",
+      "-GNinja",
+      f"-B{build_dir}",
+      f"-S{os.path.join(llvm_path, 'llvm')}",
+      "-DLLVM_ENABLE_PROJECTS=mlir",
```
nicolasvasilache, line 109: May be going a bit overboard but to properly use the `llvm-mca.py` I added and really get to the assembly, I am a bit more aggressive here: `-DLLVM_ENABLE_PROJECTS="mlir;clang;clang-tools-extra"`.

configure.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python
```
nicolasvasilache, line 1: Thanks, this should improve the situation a lot!

nicolasvasilache (approved):

tools/mlir-proto-opt/mlir-proto-opt.cpp
stellaraccident: It was missing build support so I commented it out. That has now been resolved, so re-enabled.

configure.py
stellaraccident: Done - I now print a warning if LLD is not found but enable it regardless. For ccache, it will warn if not found.

configure.py
stellaraccident: There are... strong opinions and variation in the field. For a project like this, probably better to err on the side of user friendly defaults.

stellaraccident:
I'm going to go ahead and commit this to get the CI priming and then followup with additional fixes.

configure.py
```diff
@@ -0,0 +1,140 @@
...
+  ] + llvm_configure_args
+  print(f"-- Running cmake:\n  {' '.join(cmake_args)}")
+  subprocess.check_call(cmake_args, cwd=build_dir)
+
+  # Write out .env.
+  with open(f"{os.path.join(args.repo_root, '.env')}", "wt") as f:
+    f.write(f"PYTHONPATH={os.path.join(build_dir, 'tools', 'sandbox', 'python_package')}")
+
+  # Do initial build.
+  cmake_args = ["cmake", "--build", build_dir, "--target", "tools/sandbox/all"]
```
nicolasvasilache, line 132: I am not clear where the target `tools/sandbox/all` comes from? 
I had to do a bunch of changes to the setup to work on my machine, atm it is failing the iree build:
https://github.com/google/iree-llvm-sandbox/pull/17

I'll try one more time and if it does not go through I'll have to temporarily disable it until you make another pass at it.

stellaraccident: Not a problem to disable.

That is just cmake/ninja shortcuts. You will find tools/sandbox as a directory in your build dir. And cmake generates an "all" target for each directory. Building subsets this way is usually how i work in large projects.

Llvm puts all external projects under "tools". That is where mlir is too. So tools/mlir/all works as well.

PR #12, opened 2021-11-22
Fix cmake post file moves to mirror MLIR core layout

nicolasvasilache:


lib/CMakeLists.txt
```diff
@@ -1,13 +1,16 @@
-add_subdirectory(LinalgExt)
-add_subdirectory(VectorExt)
+add_subdirectory(Dialects)
+add_subdirectory(Transforms)
+
+add_mlir_public_c_api_library(IREELinalgTensorSandboxCAPI
```
gysit, line 4: Is there a reason why IREELinalgTensorSandboxCAPI moved in front of IREELinalgTensorSandbox? I would have assumed the order should be inverse due to the dependency? But maybe cmake does not care?

stellaraccident (approved):

stellaraccident (approved):

PR #16, opened 2021-11-23
Cleanup to LinalgExt tiling algorithm and tests.

MaheshRavishankar:


stellaraccident (approved):

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -137,11 +137,11 @@ Operation *ReverseOp::getTiledImplementation(OpBuilder &builder,
   SmallVector<Type, 4> resultTypes;
   if (hasTensorSemantics()) {
     tiledOperands.emplace_back(
-        getSlice(builder, loc, output(), mirrorOffsets, sizes, strides));
+        getSlice(builder, loc, outputs[0], mirrorOffsets, sizes, strides));
```
nicolasvasilache, line 140: Why this change ?
It's not like the algorithm as written supports multi-output, so I'd expect either `output()` or an early exit somewhere that there is a single `output` in this instance of the op (i.e. `outputs.empty()`).

lib/Dialects/LinalgExt/Transforms/Tiling.cpp
```diff
@@ -39,24 +39,83 @@ static Value insertSliceIntoTensor(OpBuilder &b, Location loc,
...
-struct OpTilingPattern : public OpRewritePattern<TiledOp> {
+/// The tiling here works by two steps. The first step is to create a loop based
+/// on the loop bounds of the operation obtained from `TilingInterface`.
+///
+/// ```mlir
+///   %1 = <tiling interface op> ins(...) outs(%0 : ...)
+///   ... <use_op> ... %1 ...
+/// ```
+///
+/// is rewritten as
```
nicolasvasilache, line 50: "is rewritten using a "noop" subtensor extract/insert pair"

test/LinalgExt/tiling.mlir
```diff
@@ -1,6 +1,5 @@
-// RUN: mlir-proto-opt -linalg-ext-tiling="tile-sizes=2,4" %s | FileCheck %s
+// RUN: mlir-proto-opt -linalg-ext-tiling="tile-sizes=10,20" -split-input-file -canonicalize -cse %s | FileCheck %s
```
nicolasvasilache, line 1: Core has good hygiene here in that the base patterns do not involve multiple passes.
Let's please avoid this here too.

test/LinalgExt/tiling.mlir
```diff
@@ -13,3 +12,30 @@ func @reverse_1d_tensor(%arg0 : tensor<?x?xf32>) -> tensor<?x?xf32> {
...
+//      CHECK:   %[[RESULT:.+]] = scf.for %[[IV0:[a-zA-Z0-9]+]] = %[[C0]] to %[[D0]] step %[[C10]]
+// CHECK-SAME:       iter_args(%[[ARG1:[a-zA-Z0-9]+]] = %[[INIT]])
+//      CHECK:     %[[YIELD:.+]] = scf.for %[[IV1:[a-zA-Z0-9]+]] = %[[C0]] to %[[D1]] step %[[C20]]
+// CHECK-SAME:         iter_args(%[[ARG2:[a-zA-Z0-9]+]] = %[[ARG1]])
+//  CHECK-DAG:       %[[T0:.+]] = affine.min #[[MAP0]](%[[IV0]])[%[[D0]]]
+//  CHECK-DAG:       %[[T1:.+]] = affine.min #[[MAP1]](%[[IV1]])[%[[D1]]]
+//      CHECK:       %[[INPUT_SLICE:.+]] = tensor.extract_slice %[[ARG0]][%[[IV0]], %[[IV1]]] [%[[T0]], %[[T1]]]
+//      CHECK:       %[[OUT_SLICE:.+]] = linalg.init_tensor [%[[T0]], %[[T1]]]
+//      CHECK:       %[[RESULT_SLICE:.+]] = linalg_ext.reverse
+// CHECK-SAME:           __internal_linalg_transform__ = "tiled"
```
nicolasvasilache, line 35: The our base pattern that we test should not leak such internal details.
This is an implementation detail of using a staged pattern rewriter that should be cleanly layered on top.
Atm we have a special pattern specifically for this purpose, in the future we should have a new driver for this.

nicolasvasilache (changes requested):

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
MaheshRavishankar: `output()` would refer to the destination of the untiled operation. I need to rework this a bit to be the `iter_args` of the inner loop. This is WIP, so technically it is correct, but does not work for reductions. Hoping to land this clean up before changing it to handle reduction cases as well.

test/LinalgExt/tiling.mlir
MaheshRavishankar: Done.

test/LinalgExt/tiling.mlir
MaheshRavishankar: Removed.

nicolasvasilache (approved):

PR #17, opened 2021-11-23
Configure and README improvements

nicolasvasilache:


gysit (approved):
LGTM!

configure.py
```diff
@@ -7,44 +7,55 @@
...
   return parser.parse_args()
 
 
 def main(args):
+  # Sanity checks so we are all on the same page.
+  assert "IREE_LLVM_SANDBOX_SOURCE_DIR" in os.environ, \
+    "env var IREE_LLVM_SANDBOX_SOURCE_DIR must be set"
+  assert "IREE_LLVM_SANDBOX_SOURCE_DIR" in os.environ and \
+    os.getenv("IREE_LLVM_SANDBOX_SOURCE_DIR") == os.path.abspath(os.path.dirname(__file__)), \
+    f"env var IREE_LLVM_SANDBOX_SOURCE_DIR must be set to {os.path.abspath(os.path.dirname(__file__))}"
```
giuseros, line 33: Minor issue: if the env var has a trailing `\' this will error

configure.py
```diff
@@ -60,55 +71,53 @@ def main(args):
...
-    print(f"-- Using explicit llvm-project path: {llvm_path}")
-  elif llvm_path:
-    print(f"-- Using inferred llvm-project path: {llvm_path}")
   else:
-    llvm_path = os.path.join(args.repo_root, "..", "llvm-project")
+    # Detect LLVM.
+    # As per "Building without IREE" instructions:
+    # You **must** checkout [llvm-project](https://github.com/llvm/llvm-project) at a
+    # compatible commit next to this directory.
+    llvm_path = os.path.join(repo_root, "..", "llvm-project")
```
giuseros, line 79: This is problematic: I have different clones of llvm, and llvm-project is not the one I am using for this. Can we use an env var also for the path to LLVM? Or at least have the path to be specified as an option?

giuseros (commented):
It works almost out-of-the-box for me. I have highlighted few issues I still had to build on AArch64.

configure.py
nicolasvasilache: ok, but is this something to act on?
I don't plan on handling error cases here, just don't put a trailing char?

configure.py
nicolasvasilache: I'll have to defer to @stellaraccident on this as the README and the script were inconsistent.
This has to do with how external projects build with LLVM and I think there were changes recently.
An easy way would be for you to put a symlink and call it a day.

configure.py
```diff
@@ -60,55 +71,53 @@ def main(args):
...
-  build_dir = os.path.abspath(args.build_dir)
+  build_dir = os.path.abspath(os.getenv("IREE_LLVM_SANDBOX_BUILD_DIR"))
   os.makedirs(build_dir, exist_ok=True)
   cmake_args = [
       "cmake",
       "-GNinja",
       f"-B{build_dir}",
       f"-S{os.path.join(llvm_path, 'llvm')}",
-      "-DLLVM_ENABLE_LLD=ON",
       f"-DLLVM_ENABLE_PROJECTS={';'.join(llvm_builtin_projects)}",
```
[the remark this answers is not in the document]
nicolasvasilache: this was already here before and it works for me; does it break for you?

configure.py
```diff
@@ -60,55 +71,53 @@ def main(args):
...
       "-DLLVM_TARGETS_TO_BUILD=X86;NVPTX",
```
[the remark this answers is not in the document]
nicolasvasilache: not a problem on my machine either but ok

configure.py
```diff
@@ -125,11 +134,14 @@ def main(args):
   subprocess.check_call(cmake_args, cwd=build_dir)
```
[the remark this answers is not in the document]
nicolasvasilache: works for me, updating.

configure.py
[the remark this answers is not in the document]
nicolasvasilache: actually no, double quotes yiekd this error:

```
CMake Error at CMakeLists.txt:756 (message):
  The target `"X86"' is experimental and must be passed via
  LLVM_EXPERIMENTAL_TARGETS_TO_BUILD.
```

configure.py
stellaraccident: It's your project to decide. I personally prefer having options to specify the paths like the script originally had. I don't have a strong reason -- just a lot of time making scripts like this more general for weird situations. If it were me, I would use the side by side checkout thing as a default but slow override with options.

configure.py
```diff
@@ -7,44 +7,58 @@
...
-                      type=str,
-                      default="build")
   return parser.parse_args()
 
 
 def main(args):
+  # Sanity checks so we are all on the same page.
+  assert "IREE_LLVM_SANDBOX_SOURCE_DIR" in os.environ, \
+    "env var IREE_LLVM_SANDBOX_SOURCE_DIR must be set"
+  assert "IREE_LLVM_SANDBOX_SOURCE_DIR" in os.environ and \
```
stellaraccident, line 34: Do you really need this? It is redundant with the location of the script: os.path.dirname(__file__)

configure.py
```diff
@@ -7,44 +7,58 @@
...
+    os.getenv("IREE_LLVM_SANDBOX_SOURCE_DIR") == os.path.abspath(os.path.dirname(__file__)), \
+    f"env var IREE_LLVM_SANDBOX_SOURCE_DIR must be set to {os.path.abspath(os.path.dirname(__file__))}"
+  assert "IREE_LLVM_SANDBOX_BUILD_DIR" in os.environ, \
```
stellaraccident, line 37: I generally prefer better defaults (ie. Build for in source tree) that allows override but up to you.

configure.py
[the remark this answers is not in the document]
stellaraccident: I think it gets escaped properly down the line by subprocess. Generally, you can count on each argument being an actual argument in the executable argv

configure.py
nicolasvasilache: I don't have a strong opinion on this, I would just like consistency between LLVM, the README, what configure.py does and my ability to build locally. 

The current proposed changes are my iteration on it as I was not able to reconcile things otherwise.
I am happy to revert to a prior version if all things agree.

It is my understanding we have the constraint that iree_llvm_sandbox, llvm_project and iree must be silblings under a common parent root (or we had that constraint until recently). This seems to directly contradict being able to use iree/llvm-project, so this can't be true.

Anyway, I am looking for advice / what to revert while still being able to build and run the project.

configure.py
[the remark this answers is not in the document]
giuseros: I was manually adding quotes before :). To me it breaks with: 
```
/bin/sh: clang: command not found
/bin/sh: clang-tools-extra: command not found
```
Might be because of a different shell?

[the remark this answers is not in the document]
stellaraccident: Weird. If the quotes help and don't break the CI, fine with me (I've seen cases where adding quotes to this API does break things).

configure.py
giuseros: The main problem is that in the same folder I have at least three llvm clones, and one of them is named llvm-project, but it is quite old, so I was getting weird errors (albeit the `configure.py` was still detecting the `llvm-project` folder) . I agree with @stellaraccident , having an additional option to override the default llvm-project path should be simple enough and unharmful. I don't want to block this PR though. I can add it later in a different PR

stellaraccident: My guidance would be: the side by side constraint is a useful default, but explicit paths/options always gets used eventually for advanced cases.

The property that you really want is: check things out like this, run script without options, success. Having knobs to control it for advanced cases usually is required eventually, and there is nothing in the build setup that *requires* these projects live side by side.

configure.py
nicolasvasilache: It is redundant but then the README is off.
I'd happily drop from the README `IREE_LLVM_SANDBOX_SOURCE_DIR`, the fewer chances for inconsistencies the better.

configure.py
nicolasvasilache: `there is nothing in the build setup that requires these projects live side by side` ok I think this is the information I was hung up on, I should then be able to update more of the README and less of the configure script. I'll start a new PR.

configure.py
giuseros: Well, it should be sufficient to replace it with:
```
os.getenv("IREE_LLVM_SANDBOX_SOURCE_DIR").rstrip('/') == os.path.abspath(os.path.dirname(__file__))
```

configure.py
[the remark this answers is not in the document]
giuseros: Oh. What shell are you using ? On mine `-DLLVM_TARGETS_TO_BUILD="X86"` works fine

nicolasvasilache:
@stellaraccident @giuseros I think I can retire this in favor of a simpler #19 sorry for the noise..

PR #19, opened 2021-11-23
Update and improve README + config script

nicolasvasilache:


nicolasvasilache:
Also @giuseros, my shell is:
```
ps -p $$ 
    PID TTY          TIME CMD
2386742 pts/19   00:00:00 bash
```

stellaraccident (approved):
Fwiw - in these kinds of early states, I am fine with post review on basic bootstrapping stuff.

stellaraccident:
Also, during bootstrapping, I'm fine landing before the CI is finished --
as long as it gets to the part where it is compiling. It takes an hour or
so when the llvm version bumps. During regular dev, it hasn't been so bad,
but a real pita if iterating on the config.

nicolasvasilache:
SG, thanks for the guidance @stellaraccident !

configure.py
```diff
@@ -110,7 +121,7 @@ def main(args):
       f"-S{os.path.join(llvm_path, 'llvm')}",
       "-DLLVM_ENABLE_LLD=ON",
```
giuseros, line 122: @nicolasvasilache could you remove this since you introduced an option to specify `DLLVM_ENABLE_LLD` ?

giuseros (commented):
I don't know exactly what you did, but everything seems to work now! I am happy with this bar the `lld` thing which is still enabled by default (see comments).

configure.py
nicolasvasilache: ah yes, sorry, see #21 
you should also have an invite in  https://github.com/google/iree-llvm-sandbox/invitations so I can add you to reviews
please also send me the github logins of a few folks I should add on your end

PR #23, opened 2021-11-24
Fix timing facility for reference benchmarks

ftynse:
A recent move to C++-driven time measurement changed the timing facility
in a way that is not suitable for reference NumPy and PyTorch
benchmarks: they are not returning a list of times like the generated
code. Put back the Python timing facility for these caess.


nicolasvasilache:
forgot to approve and just merged, sorry 😛

PR #24, opened 2021-11-24
Add tests/benchmarks for 2D column reduction.

pifon2a:


python/reduction/column_reduction_2d_bench.py
```diff
@@ -0,0 +1,93 @@
...
+          # Little trick avoids tiling small dimensions and otherwise tile by 128.
+          sizes=[4, 128] if problem_sizes[1] > 256 else [4],
+          interchange=[],
+          peel=[],
+          pad=False,
+          pack_paddings=[],
+          hoist_paddings=[],
+          # kwargs passed down to LowerVectors.
+          # TODO: better composition of experts.
+          multi_reduction_lowering='innerparallel',
```
nicolasvasilache, line 48: you likely want something else here

python/reduction/column_reduction_2d_bench.py
```diff
@@ -0,0 +1,93 @@
...
+    TransformationList.__init__(self, **kwargs)
+
+
+def all_experts(problem_sizes: List[int]):
+  return [
+      SingleTilingExpert(
+          fun_name=fun_name,
+          op_name=op_name,
+          # Little trick avoids tiling small dimensions and otherwise tile by 128.
+          sizes=[4, 128] if problem_sizes[1] > 256 else [4],
```
nicolasvasilache, line 40: Had a bunch of those in the transpose, ideally you'd use something with a good name and put it in utils.
But this is so sall that I don't want to slow you down :)

nicolasvasilache (approved):

PR #25, opened 2021-11-24
Include pad op vectorization patterns with other vectorization patterns.

MaheshRavishankar:


nicolasvasilache (approved):
LGTM but this seems to break the build, please fix before landing.

MaheshRavishankar:
> LGTM but this seems to break the build, please fix before landing.

It's needs LLVM hash that CI uses to be updated.

MaheshRavishankar:
Actually if the test is re-triggered the builds should pass (the hash on IREE side is updated to the required point). I dont know how to do that though (@stellaraccident ?)

PR #26, opened 2021-11-24
Move expert definition to C++ passes

MaheshRavishankar:
Turn the python expert into a thin wrapper around the C++ expert

### DIFF
--- a/include/Transforms/Passes.h
+++ b/include/Transforms/Passes.h
@@ -28,7 +28,7 @@ std::unique_ptr<OperationPass<FuncOp>>
 createLinalgFuseOutputIntoReductionPass();
 
 /// Creates a pass to drive transformations on Linalg on tensors.
-std::unique_ptr<OperationPass<FuncOp>> createLinalgTensorCodegenDriverPass();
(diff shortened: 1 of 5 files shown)


MaheshRavishankar:
Only the last two commits are part of this PR. The first commit is part of #25

nicolasvasilache (approved):
Please add a more descriptive comment in line with what we discussed offline: 

```
Move expert definition to C++ passes 
Turn the python expert into a thin wrapper around the C++ expert
```

LGTM but this seems to break the build, please fix before landing.

MaheshRavishankar:
> Please add a more descriptive comment in line with what we discussed offline: 
> 
> ```
> Move expert definition to C++ passes 
> Turn the python expert into a thin wrapper around the C++ expert
> ```
> 
> LGTM but this seems to break the build, please fix before landing. 

The breakage is because CI needs to use a different LLVM commit

nicolasvasilache:
ah yes, I forgot the build automatically pulls down IREE's pinned LLVM version.
Ok if you've tested let's merge then.

PR #32, opened 2021-11-25
Add missing dependencies to CMake

ftynse:
Numerous dependencies were missing from various CMake files. The entire
build probably worked because it was never attempted with
`-DBUILD_SHARED_LIBS=On`, but there is no reason it won't break in a
different environment.


nicolasvasilache (approved):
Thanks much !

PR #36, opened 2021-11-25
Add in_parallel, perform_concurrently and parallel_insert_slice operations

apaszke:
Including printers and parsers that roundtrip. They don't do anything interesting yet.

cc @nicolasvasilache


include/Dialects/LinalgExt/LinalgExtOps.td
```diff
@@ -191,4 +191,129 @@ def LinalgExt_TileYieldOp : LinalgExt_PureOp<"tile_yield", [
   let assemblyFormat = "attr-dict ($operands^ `:` type($operands))?";
 }
 
+def LinalgExt_InParallelOp : LinalgExt_PureOp<"in_parallel", [
```
nicolasvasilache, line 194: Maybe add some notes that this is very similar to an scf.for_op but:
- without iter_args 
- a special terminator (that would really like to be a dependent-attribute that can refer to SSA values)
- with a single thread id in the range 0:num_threads:1 instead of an induction variable.

include/Dialects/LinalgExt/LinalgExtOps.td
```diff
@@ -191,4 +191,129 @@ def LinalgExt_TileYieldOp : LinalgExt_PureOp<"tile_yield", [
...
+def LinalgExt_EndPerformConcurrentlyOp : LinalgExt_PureOp<"end_perform_concurrently", [
+       NoSideEffect, Terminator]> {
+  let summary = "terminates a `perform_concurrently` block";
+  let description = [{
+    A designated terminator for `perform_concurrently`. It's not expected to appear
+    in the textual form of the IR.
+  }];
+}
+
+// Note that we cannot mark this operation as pure (NoSideEffects), even
```
nicolasvasilache, line 280: I'd even that this in the actual op description, this is part of the semantics.

include/Dialects/LinalgExt/LinalgExtOps.td
```diff
@@ -191,4 +191,129 @@ def LinalgExt_TileYieldOp : LinalgExt_PureOp<"tile_yield", [
...
+    If an index is updated by exactly one updates, the value contained at that index
+    in the resulting tensor will be equal to the value at a corresponding index of a
+    slice that was used for the updated. If an index is not updated at all, its value
+    will be equal to the one in the original tensor.
+  }];
+
+  let arguments = (ins
+    AnyRankedTensor:$source,
+    AnyRankedTensor:$dest,
+    Variadic<Index>:$offsets,
```
nicolasvasilache, line 305: Note: There is a special interface for this kind of ops that comes with canonicalizations etc.

The problem though is that the interface wants return values that you omit here.
We should still consider refactoring the OpInterface and cleanup.

test/LinalgExt/invalid.mlir
```diff
@@ -0,0 +1,44 @@
+// RUN: mlir-opt %s -split-input-file -verify-diagnostics
+
+// -----
+
+func @not_enough_results() -> () {
+  %num_threads = arith.constant 100 : index
+  // expected-error@+1 {{'linalg_ext.in_parallel' op has 1 results, but its terminator produces 0 values}}
```
nicolasvasilache, line 7: `s/produces/has` of `s/produces/yields` here, below and in the .cpp ?

I'd expect to see a value returned if it were produced but here the terminator has no results.

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -208,5 +208,142 @@ static ParseResult parseTileOp(OpAsmParser &parser, OperationState &result) {
...
+      parser.resolveOperand(numThreads, indexType, result.operands))
+    return failure();
+  if (parser.parseArrowTypeList(result.types)) return failure();
+
+  SmallVector<OpAsmParser::OperandType, 8> regionOperands;
+  SmallVector<Type, 8> regionTypes;
+  std::unique_ptr<Region> region = std::make_unique<Region>();
+  if (parser.parseRegion(*region, regionOperands, regionTypes))
+    return failure();
+  InParallelOp::ensureTerminator(*region, builder, result.location);
```
nicolasvasilache, line 272: Somewhere the BB produced by this call is missing the index barg and other things fail downstream.

Please send a fix and a proper test soon after landing this.

nicolasvasilache (approved):

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
apaszke: Yes, that's expected. `ensureTerminator` only ensures that the block exists and is terminated, but it's not responsible for adding any necessary args. For example, the implementation of `ensureTerminator` for `scf.for` doesn't insert bbargs for the induction variable and its `iter_args`.

include/Dialects/LinalgExt/LinalgExtOps.td
apaszke: Not sure if I follow. Is there anything you'd like to change before this lands, or is it more of a note for the future?

nicolasvasilache:
It's a suggestion for change it is not a blocker, exercise your judgement.

test/LinalgExt/invalid.mlir
apaszke: Done!

include/Dialects/LinalgExt/LinalgExtOps.td
apaszke: Done.

include/Dialects/LinalgExt/LinalgExtOps.td
apaszke: Done.

PR #41, opened 2021-11-26
Fix github actions workflow

nicolasvasilache:


nicolasvasilache:
Submitting to accelerate progress.

PR #42, opened 2021-11-26
Add tests/benchmarks for 1d reduction.

pifon2a:


python/reduction/definitions.py
```diff
@@ -15,6 +15,126 @@
...
+################################################################################
+#   Op def: ( m )
+#    Iters: ({Red()})
+#             I   O
+#   Layout: {{m}, {}}
+
+
+# TODO: fold OpDSL definition and inferences into ProblemDefinition.
+@linalg_structured_op
+def reduction_1d(A=TensorDef(T, S.M), B=TensorDef(T, output=True)):
```
nicolasvasilache, line 29: If you consider adding another one, it would prob time to generalize this on the model of what @ftynse did on other things.

nicolasvasilache (approved):

PR #51, opened 2021-11-26
Make experts compose

ftynse:
This is the first step towards more composable `TransformationList` (aka
expert) API. It enables a construction of a `TransformationList` from a
list of Transformation subclasses without instantiating them as well as
a direct concatenation of two `TransformationList`s operating on the
same function/opname pair. Thanks to a metaclass, such
`TransformationList`s are concrete subclasses with a flat list of
variables amenable to the existing search mechanism. Each Transformation
now has its own list of variables that is forwarded down from the
`TransformationList`.

Further work is necessary to:

  - better compose TransformationList objects in addition to classes
    (these objects are not subject to search);
  - allow for direct concatenation of TransformationLists that operate o
    different function/opname pairs.


python/core/experts.py
```diff
@@ -47,234 +52,128 @@ def __call__(self, entry_point_name: str, module: Module):
...
+VectorLowering = TransformListFactory('VectorLowering',
+                                      [LowerVectorFactory(i) for i in range(7)])
+
+# TODO: After DecomposeToLowerDimensionalNamedOp the op_name to anchor on
+# changes: we need a better control mechanism.
+LoweringOnlyExpert = Bufferization.then(VectorLowering).then(LLVMLowering)
+SingleTilingExpert = Tiling.then(Decomposition).then(Vectorization).then(
+    LoweringOnlyExpert)
+DoubleTilingExpert = Tiling.then(SingleTilingExpert)
+TripleTilingExpert = Tiling.then(DoubleTilingExpert)
```
nicolasvasilache, line 176: 🚀

nicolasvasilache (approved):

PR #52, opened 2021-11-26
Add a rewrite from linalg_ext::in_parallel to scf::for

nicolasvasilache:


include/Dialects/LinalgExt/LinalgExtOps.td
```diff
@@ -264,20 +264,9 @@ def LinalgExt_PerformConcurrentlyOp : LinalgExt_PureOp<"perform_concurrently", [
...
-        }
-        if (auto endPerformOp = llvm::dyn_cast<EndPerformConcurrentlyOp>(op)) {
-          continue;
-        }
-        llvm_unreachable("Unexpected operation in perform_concurrently");
-      }
-      return ret;
-    };
+    SmallVector<Type> yieldedTypes();
+    SmallVector<Value> yieldedValues();
```
apaszke, line 268: I'm not sure if yielded values are even well defined, so I'd rather not add it unless we have to.

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -356,9 +360,58 @@ static ParseResult parsePerformConcurrentlyOp(OpAsmParser &parser,
...
+    }
+    llvm_unreachable("Unexpected operation in perform_concurrently");
+  }
+  return ret;
+}
+
+SmallVector<ParallelInsertSliceOp> PerformConcurrentlyOp::yieldingOps() {
+  SmallVector<ParallelInsertSliceOp> ret;
+  for (Operation &op : *getBody()) {
+    // TODO: llvmTypeSwitch when this grows up.
```
apaszke, line 403: Or perhaps an interface instead?

lib/Dialects/LinalgExt/Transforms/InParallelToSequentialFor.cpp
```diff
@@ -0,0 +1,115 @@
...
+    // As a consequence, `ensureTerminator` is not called and the body has no
+    // terminator.
+    PerformConcurrentlyOp performConcurrentlyOp = inParallelOp.getTerminator();
+    scf::ForOp forOp = rewriter.create<scf::ForOp>(
+        loc, zero, numThreads, one, performConcurrentlyOp.yieldedValues());
+
+    rewriter.setInsertionPointToStart(forOp.getBody());
+
+    BlockAndValueMapping bvm;
+    bvm.map(performConcurrentlyOp.yieldedValues(), forOp.getRegionIterArgs());
```
apaszke, line 56: Perhaps move this after `mergeBlock`? It's placement is a little confusing, because it seems like this mapping might influence block merging but it really shouldn't.

lib/Dialects/LinalgExt/Transforms/InParallelToSequentialFor.cpp
```diff
@@ -0,0 +1,115 @@
...
+    rewriter.setInsertionPoint(performConcurrentlyOp);
+    for (ParallelInsertSliceOp op : performConcurrentlyOp.yieldingOps()) {
+      toYield.push_back(rewriter.createOrFold<tensor::InsertSliceOp>(
+          loc, op.source(), bvm.lookup(op.dest()), op.offsets(), op.sizes(),
+          op.strides()));
+    }
+
+    // performConcurrentlyOp.yieldedValues come from above, not from bbArgs.
+    // There is no rewriter method to make mergeBlocks update non-bbArgs.
+    // Need to manually clone + bvm all uses that are now nested under forOp.
```
apaszke, line 73: I actually think that this is incorrect. If a loop iteration uses a "yielded value" then it should see the tensor value from before the `in_parallel` block runs, not a partially updated one.

apaszke (changes requested):
Mostly LGTM, except for the one replacement of uses of the values updated inside `in_parallel`.

It would be also good to discuss yielded values, because I don't think this concept even makes sense if we add e.g. associative reductions.

lib/Dialects/LinalgExt/Transforms/InParallelToSequentialFor.cpp
nicolasvasilache: I am not sure I follow, the scf.for does an incremental partial update at each sequential iteration.
This bufferizes to partial updates to the memory region at each iteration.
How would you write the following in the sequential case ?

```
func @static_tile(%arg0: index, %arg1: tensor<?xf32>, %arg2: tensor<?xf32>) -> ... {
  %2 = scf.for %arg3 = %c0 to %1 step %c1 iter_args(%arg4 = %arg2) -> (tensor<?xf32>) {
    %6 = tensor.extract_slice %arg4[%3] [%5] [%c1] : tensor<?xf32> to tensor<?xf32>
    %7 = tensor.extract_slice %arg1[%3] [%5] [1] : tensor<?xf32> to tensor<?xf32>
    %8 = linalg.generic {indexing_maps = [#map4, #map4], iterator_types = ["parallel"]} ins(%7 : tensor<?xf32>) outs(%6 : tensor<?xf32>) {...}
    %9 = tensor.insert_slice %8 into %arg4[%3] [%5] [%c1] : tensor<?xf32> into tensor<?xf32>
    scf.yield %9 : tensor<?xf32>
  }
}
```

nicolasvasilache: Adding a scary comment as per our offline chat.

nicolasvasilache:
Adressed feedback, let's iterate with post-review if needed.

PR #53, opened 2021-11-26
Make `Transformation`s classes chainable to form `TransformationList`s

ftynse:
This removes some boilerplate in construction of singleton
transformation lists for the purpose of further composition.

### DIFF
--- a/python/matmul/test.py
+++ b/python/matmul/test.py
@@ -42,14 +42,15 @@ expert_tile_and_interchange_1 = SingleTilingExpert('matmul_on_tensors',
                                                    peel=[])
 
 # 1 level of tiling and then generalize and interchange.
-expert_tile_1_and_generalize_interchange = Tiling.then(Generalization).then(
-    Vectorization).then(LoweringOnlyExpert)('matmul_on_tensors',
-                                            'linalg.generic',
-                                            tile_sizes=[8, 8, 24],
-                                            tile_interchange=[2, 0, 1],
-                                            pad=False,
-                                            peel=[],
-                                            iterator_interchange=[0, 2, 1])
+expert_tile_1_and_generalize_interchange = \
+    Tile.then(Generalize).then(Vectorize).then(LoweringOnlyExpert)(\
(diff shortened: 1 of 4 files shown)


python/matmul/test.py
```diff
@@ -42,14 +42,14 @@ def __init__(self, tiling_transforms):
...
 # 1 level of tiling and then generalize and interchange.
-expert_tile_1_and_generalize_interchange = Tiling.then(Generalization).then(
-    Vectorization).then(LoweringOnlyExpert)('matmul_on_tensors',
-                                            'linalg.generic',
-                                            tile_sizes=[8, 8, 24],
-                                            tile_interchange=[2, 0, 1],
-                                            pad=False,
-                                            peel=[],
-                                            iterator_interchange=[0, 2, 1])
+expert_tile_1_and_generalize_interchange = Tile.then(Generalize).then(
```
nicolasvasilache, line 45: Nite: can we make this resemble:

```
expert_tile_1_and_generalize_interchange = \
  Tile('matmul_on_tensors',                              \
         'linalg.generic',                                          \
         tile_sizes=[8, 8, 24],                                 \
         tile_interchange=[2, 0, 1],                        \
         pad=False,                                                 \
         peel=[],                                                      \
         iterator_interchange=[0, 2, 1])                \
  .then(Generalize)                                             \
  .then(Vectorize)                                               \
  .then(LoweringOnlyExpert)
```

or is python going to complain?

nicolasvasilache: yes, I insist on the backslashes being at exactly those positions :) ... sigh webUI

nicolasvasilache (approved):

python/matmul/test.py
ftynse: We can't (yet) have parameterized transforms followed by unparameterized transforms, so no options directly to the first `Tile`. Added the backslashes though, I would have hoped for the formatter to put them in.

nicolasvasilache: sg, thanks!

PR #55, opened 2021-11-29
Update fusion to use the new tileAndFuse implemented by CodegenStrategy

gysit:
The revision implements the following changes:
- update the fusion test to use the new way of defining transformations.
- replace the sandbox fusion implementation by the tile and fusion implemented by CodegenStrategy.
- add padding and vectorization to the fuse transform since they are now supported.
- add helper functions to parse tiling and padding options.

The commits depend on the following upstream revisions: 
- https://reviews.llvm.org/D114690
- https://reviews.llvm.org/D114689
- https://reviews.llvm.org/D114585

It thus bumps the llvm version to a commit that includes all revsion.


python/core/transforms.py
```diff
@@ -63,32 +63,63 @@ class Fuse(Transform):
...
       dims = [str(ic) for ic in self.tile_interchange]
       interchange_str = f'tile-interchange={",".join(dims)}'
+    if self.pad:
+      packing_flags = [str(pp) for pp in self.pack_paddings]
+      hoisting_depths = [str(hd) for hd in self.hoist_paddings]
+      pad_str = f'pad'
+      if packing_flags:
+        pad_str = pad_str + f' pack-paddings={",".join(packing_flags)}'
+      if hoisting_depths:
+        pad_str = pad_str + f' hoist-paddings={",".join(hoisting_depths)}'
```
ftynse, line 111: This looks similar to what `Tile.__init__` does below, any chance of generalizing and factoring out?

python/core/transforms.py
```diff
@@ -6,6 +6,36 @@
 import mlir.all_passes_registration
 
 
+def _get_tile_sizes_str(transform: Transform):
```
ftynse, line 9: Nit: trailing `-> str` for more type hints plz.

ftynse (approved):

MaheshRavishankar (approved):
LGTM!

hanhanW (approved):
LGTM!

The usage of tileAndFuse in CodegenStrategy basically covers the uses in IREE. In addition, it has more features like interchange support. I'm looking forward to integrate these things into IREE.

PR #59, opened 2021-11-29
Make Transformation instances compose similarly to classes

ftynse:
A previous commit provided a fluent-style API for `Transformation` and
`TransformationList` subclasses to compose and create new
`TransformationList` subclasses parameterizable by a union of the
options of individual transformations. Implement the same for objects of
said classes. This makes the nicely chainable with the `then` function
that clearly indicates the order of transformations.

This makes the construction of custom transformation chains slightly
more verbose, although arguably more readable, by having each
transformation that takes parameters be listed instead of taking one
list of kwargs into a transformation list and dispatching them to
individual transformations. Further improvements are necessary in the
specification of transformation "variables", currently tied into
autotuning, to allow for a more concise specification of parameters.

### DIFF
--- a/python/conv/conv_1d_bench.py
+++ b/python/conv/conv_1d_bench.py
@@ -18,20 +18,22 @@ op_name = 'linalg.conv_1d_nwc_wcf'
 ################################################################################
 
 all_experts = [
-    SingleTilingExpert(
-        fun_name=fun_name,
-        op_name=op_name,
-        #           N  W   C  KW  F
-        tile_sizes=[1, 8, 32, 1, 8],
-        tile_interchange=[],
-        peel=[],
-        pad=False,
-        pack_paddings=[],
-        hoist_paddings=[],
-        # kwargs passed down to LowerVectors.
-        # TODO: better composition of experts.
-        transpose_lowering='shuffle',
-        print_ir_after_all=False)
+    e.print_ir(after_all=False) for e in [
+        TileAndDecompose(
+            fun_name,
+            op_name,
+            #           N  W   C  KW  F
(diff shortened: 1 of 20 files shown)


gysit (approved):
nice! There may be a merge conflict since I changed the printer thing. Just use your version!

python/core/transform.py
```diff
@@ -7,6 +8,61 @@
...
+                      else [objtype] + other_cls._transform_classes)
+        return TransformListMetaclass(name, (TransformationList,), {},
+                                      transforms=transforms)
+
+      return then_class
+
+    # Calling as instance method.
+    assert isinstance(obj, Transform)
+
+    def then_instnace(
```
gysit, line 47: nit: then_instance

python/conv/conv_1d_bench.py
```diff
@@ -18,20 +18,22 @@
...
-        # TODO: better composition of experts.
-        transpose_lowering='shuffle',
-        print_ir_after_all=False)
+    e.print_ir(after_all=False) for e in [
+        TileAndDecompose(
+            fun_name,
+            op_name,
+            #           N  W   C  KW  F
+            tile_sizes=[1, 8, 32, 1, 8],
+            tile_interchange=[],
```
nicolasvasilache, line 27: Can we drop the unused/default entries?
We should be able to write this more concisely as:
```
TileAndDecompose(fun_name, op_name, tile_sizes=[1, 8, 32, 1, 8]) ... the rest.
```

Not for this PR but maybe as a followup when everything uses your nice fluent API as it gets close to the finish line.

PR #60, opened 2021-11-29
Always use transforms for printing IR

ftynse:
`TransformationList`s used to support two ways of printing IR: using the
`Print` pseudo-transformation and using the various flags passed as
kwarg when creating the `TransformationList`. The latter is poorly
composable (e.g., transformations with different "print after all" flag
values have no natural composition). Drop it in favor of using the
`Print` transformation and the chainable `print_ir` function that
simplifies the insertion of `Print`s into the `TransformationList`.

### DIFF
--- a/python/core/transform.py
+++ b/python/core/transform.py
@@ -56,10 +56,7 @@ class _TransformThenDescriptor:
       if isinstance(other, Transform):
         return TransformationList(transforms=[obj, other])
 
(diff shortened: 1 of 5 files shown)


gysit (approved):

gysit (approved):

nicolasvasilache (commented):
Nice cleanup!

PR #62, opened 2021-11-30
Add default values to Transform variables

ftynse:
Previously, default values were provided in the parsing function call
inside the constructor of each Transform. This indirection made the
overall flow less readale. Even before that, the default values were not
available at all so multiple benchmarks had to copy-paste them when
constructing transforms. Put the default values next to the variables
and remove their assignments equal to defaults from the transform
instantiations in the benchmarks.


python/core/transform.py
```diff
@@ -78,21 +80,18 @@ def __call__(self, module: Module, fun_name: str):
...
-                                 kwargs: tp.Mapping[str, tp.Any],
-                                 defaults: tp.Mapping[str, tp.Any] = dict()):
+  def _parse_variables_in_kwargs(self, kwargs: tp.Mapping[str, tp.Any]):
     """Set up instance fields that correspond to known variables from kwargs.
-    
-    Use the values if `defaults` if `kwargs` does not have one for the given
-    variable. Either `kwargs` or `defaults` must contain a value for all known
-    variables. 
+
+    The values that are missing the default value in `variables` of the current
```
gysit, line 86: nit: missing *a* default value?

gysit (approved):
I like it!

nicolasvasilache (commented):
Beautiful ✨  please ignore my previous comment on your previous CL then 😄

PR #63, opened 2021-11-30
Add clang-format config file to sandbox

dcaballe:
The config file is the same as in MLIR core.


.clang-format
```diff
@@ -0,0 +1,2 @@
+BasedOnStyle: LLVM
+AlwaysBreakTemplateDeclarations: Yes
```
ftynse, line 2: Please add newline to the end of file :)

ftynse (approved):

nicolasvasilache (commented):
thanks @dcaballe !

PR #64, opened 2021-11-30
Add utility to vector-predicate a tiled loop

dcaballe:
This patch introduces a utility that applies vector predication to a tiled
loop. The tiled loop is assumed to be vectorized after predication using
the step of the tiled loop as vectorization factor. The transformation
is very simple for now: a new vector.predicate operation is added, which
encloses the loop body. The predicate ensures that the tiled loop will
be executed in a VF-way vector fashion without going beyond the upper bound
of the loop, even when the number of iterations is not multiple of VF.


lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+    return;
+  dest->getBlock()->getOperations().splice(
+      Block::iterator(dest), opRange.begin()->getBlock()->getOperations(),
+      opRange.begin(), opRange.end());
+}
+
+/// Iterate over all the definitions produced by the operation range `opRange`
+/// and gather those with users outside the range.
+static void
+getDefsWithExternalUses(llvm::iterator_range<Block::iterator> opRange,
```
nicolasvasilache, line 39: I think we consistently use the tern "outside" in MLIR.
So `getDesfsWithUsesOutside` ?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+                        SmallVectorImpl<Value> &defsWithExternalUses,
+                        SmallVectorImpl<Type> &defTypes) {
+  SmallPtrSet<Operation *, 8> rangeSet;
+  for (Operation &op : opRange)
+    rangeSet.insert(&op);
+
+  for (Operation &op : llvm::make_early_inc_range(opRange))
```
nicolasvasilache, line 46: do you need the early_inc ?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+        if (rangeSet.count(user) == 0) {
+          defsWithExternalUses.push_back(res);
+          defTypes.push_back(res.getType());
+          break;
+        }
+}
+
+/// Introduce a vector.predicate op that encloses the loop body of the tiled
+/// loop. The predicate is computed based on the provided vectorization factor.
+static void createVectorPredicateOp(TiledLoopOp loopOp, int64_t vecFactor) {
```
nicolasvasilache, line 58: Can we add a level of indirection from the start here?
We shouldn't anchor anything on TiledLoopOp as it has too many lose ends that will not connect.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+  SmallVector<Value, 8> defsWithExternalUses;
+  SmallVector<Type, 8> defTypes;
+  getDefsWithExternalUses(opsToMove, defsWithExternalUses, defTypes);
+
+  // Initialize a builder to insert vector.predicate at the beginning of the
+  // loop body.
+  Location loc = loopOp.getLoc();
+  MLIRContext *context = loopOp.getContext();
+  OpBuilder builder(loopOp);
```
nicolasvasilache, line 73: Passing the builder as a parameter and using RAII InsertionGuard + setting insertion points is preferred.
This is a prerequisite for proper composition of transformations and reuse in patterns (if applicable).
We might as well put it in this form from the beginning and avoid artificial frictions on composability.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+  // Generate the predicate to used by the vector.predicate operation:
+  //   %min = affine.min affine_map<(d0)[s0] -> (vecFactor, -d0 + s0)>(%iv)[%ub]
+  //   %vec_pred = vector.create_mask %min : vector<8xi1>
+  // TODO: This is probably not the most efficient way to generate the vector
+  // predicate. Consider using the vector IV.
+  AffineExpr i, j;
+  bindDims(context, i, j);
+  SmallVector<AffineMap, 4> maps = AffineMap::inferFromExprList(
+      ExprList{{builder.getAffineConstantExpr(vecFactor), i - j}});
+  auto minOp = builder.create<AffineMinOp>(
```
nicolasvasilache, line 85: use/refactor/extend the AffineBuilder utils?
e.g. https://github.com/google/iree-llvm-sandbox/blob/main/lib/Dialects/LinalgExt/Transforms/TileToSequentialFor.cpp#L73

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+  auto maskType = VectorType::get({vecFactor}, builder.getI1Type());
+  auto predicate =
+      builder.create<CreateMaskOp>(loc, maskType, ValueRange{minOp});
+
+  // Generate the vector.predicate operation and move the range [`startOp`,
+  // `endOp`] of operations within its truePredicateRegion. We have to rewire
+  // the def-use chain for those definitions within the range that have external
+  // uses. Those uses are rewired to the results of the vector.predicate.
+  auto vecPredOp = builder.create<PredicateOp>(loc, defTypes, predicate);
```
nicolasvasilache, line 97: Some of these may look quite different with a PatternRewriter, can this be isolated in a helper that takes the builder so it becomes easier to both 1) distinguish the mechanical stuff and 2) evolve as needed ?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,138 @@
...
+    builder.setInsertionPoint(truePredTerminator);
+    builder.create<vector_ext::YieldOp>(loc, defsWithExternalUses);
+    truePredTerminator->erase();
+  }
+}
+
+/// Utility that predicates a tiled loop with a vector.predicate operation. The
+/// vectorization factor used for predication is assumed to be the step of the
+/// tiled loop.
+LogicalResult mlir::vector_ext::predicateTiledLoop(TiledLoopOp loopOp) {
```
nicolasvasilache, line 120: Same comment re anchoring on `TiledLoopOp`.
Actually, can the general impl. of this be achieved by outlining inside an scf.execute op + region ?

test/VectorExt/tiled_loop_masking.mlir
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// CHECK-LABEL:   func @tiled_loop_1d(
+// CHECK:           %[[ZERO:.*]] = arith.constant 0 : index
+// CHECK:           %[[STEP:.*]] = arith.constant 8 : index
+// CHECK:           %[[UB:.*]] = tensor.dim
+// CHECK:           linalg.tiled_loop (%[[IV:.*]]) = (%[[ZERO]]) to (%[[UB]]) step (%[[STEP]])
+// CHECK-NEXT:        %[[PRED_MIN:.*]] = affine.min #map0(%[[UB]], %[[IV]])
+// CHECK-NEXT:        %[[PRED:.*]] = vector.create_mask %[[PRED_MIN]] : vector<8xi1>
+// CHECK-NEXT:        %[[PRED_RES:.*]] = vector_ext.predicate(%[[PRED]]: vector<8xi1>) -> (tensor<?xf32>) {
```
nicolasvasilache, line 36: Nice!

@ThomasRaoux here is some IR that would look similar to what we discussed yesterday for vector distribution.
Maybe the `vector.predicate` itself can be extended to also serve the GPU programming model purpose?

nicolasvasilache (commented):
Thanks @dcaballe !
Generally this looks great to me, mostly minor refactorings.

One extra Q, as I see your code moving ops around in regions, it is reminiscent to me of older stuff I experimented with. At the time I came to the conclusion that outlining through and `scf.execute` op was a good solution.

I am wondering whether parts of what you have here should be outlined inside an `scf.execute` op + region as an interim state (and inlined back immediately after, it should not leak into the result IR). This way, the type of logic you wrote would become much more reusable and we could have a good way of generalizing such rewrites with a few util/helper functions?

If that sounds interesting, I should be able to dig up some of my older stuff and we can iterate on better tooling to support such cases now and in the future.

dcaballe:
> I am wondering whether parts of what you have here should be outlined inside an scf.execute op + region as an interim state (and inlined back immediately after, it should not leak into the result IR)

(Sorry, I now understand the discussion from this morning). The `vector.predicate` op is expected to have a well-defined role and be semantically charged. It is expected to leak into the IR and stay around until we decide to lower it (probably after linalg) and materialize all the masked ops. It is also expected to have a second region (TODO) to model those ops masked by the complementary predicate. In essence, it is like an `scf.if` but with different semantics. Actually, some `scf.if` will be converted to `vector.predicate` in some scenarios. For these reasons, I don't think we can resuse the `scf.execute`. However, I do agree that we should have a bit more amenable utilities to deal with this scenarios (i.e,. introduce op with region(s) that will enclose a bunch of existing ops).

I guess what is missing from my side is to provide a bit more context about the op and where this is going but I wanted to have some PoC working before spending time on design discussions.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: Good catch! It was needed initially but not anymore. Thanks!

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: I hoped to do that when introducing support for `scf.if` but let me try now...

nicolasvasilache:
> For these reasons, I don't think we can resuse the `scf.execute`.

Just to be clear, I am not suggesting we use `scf.execute` in the `vector.predicate` IR but just as an implementation detail that everyone could funnel on and that gets folded away immediately.

The part that I find interesting with this `scf.execute` construct is that it is "half of a function".

It seems having the minimal utilities to:
1. outline ops to an scf.execute 
2. plug return values properly with yields
3. do more stuff inside that scf.execute
4. outline it back at the place of a user's choosing
5. all this while allowing BlockAndValueMapping

would allow having a single place where such manipulations of "moving ops between regions and doing BlockAndValueMapping + result types".

I'd happily reuse such a single place as my experience is that everyone ends up reimplementing some flavor of this for their own ops and it is far from trivial (in particular the self-use, infinite recursion case that can hit hard).

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: Done. At least in LLVM, the builder is intentionally lightweight to encourage its arbitrary creation/destruction as needed and avoid to carry a single builder all over the place. I guess you are referring to being able to pass either a builder or a rewriter, right?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: Good point. I didn't know about that builder! However, AffineBuilder is part of the sandbox so I think it's better to avoid dependences with other pieces of the sandbox so that we can facilitate the graduation/reversion of the experimental code (as we described in the sandbox guidelines). Does that make sense? Or  do you plan to move that to core in the upcoming days?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: Let me know what you think about the new version of the code. I changed the utility so that it returns this newly created op. Then, an external rewriter could use it to replace another op, for example. Would that be enough?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: It should be better now.

test/VectorExt/tiled_loop_masking.mlir
dcaballe: It would be great to know more about the vector distribution! I think the vector.predicate op would have applicability in the GPU programming model but I'm not sure about the semantics you are looking for. Should we schedule a 1:1 to discuss this further?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/// Introduce a vector.predicate op that encloses all the operations in
+/// `regionToPredicate` (except its terminator). `createPredicate` is used to
+/// create the operations that generate the predicate used by vector.predicate.
+/// These operations will be inserted at the beginning of `regionToPredicate`.
+/// The new vector.predicate will be inserted right after the operations
+/// generating the predicate.
+Optional<PredicateOp> mlir::vector_ext::predicateOp(
+    Operation *op, Region *regionToPredicate, OpBuilder &builder,
```
nicolasvasilache, line 62: We consistently pass the builder as the first param in core MLIR.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
nicolasvasilache: yes, looks great, the idea was to avoid cloning the builder internally which would create a mess when mixed with PatterRewriter.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+  // Retrieve vectorization factor from the step of the tiled loop.
+  Value step = loopOp.step()[0];
+  auto maybeVecFactor = getConstantIntValue(step);
+  if (!maybeVecFactor)
+    return failure();
+  int64_t vecFactor = *maybeVecFactor;
+
+  auto createPredicate = [&](OpBuilder &builder) -> Value {
+    // Generate the predicate to used by the vector.predicate operation:
```
nicolasvasilache, line 127: nit: wording

test/VectorExt/tiled_loop_masking.mlir
nicolasvasilache: Yes, this is also related to discussions around scalable vectors: https://llvm.discourse.group/t/rfc-add-built-in-support-for-scalable-vector-types/4484/13
Let's plan a meeting with @ThomasRaoux too on PST time and see how we can split things up here.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+  return vecPredOp;
+}
+
+/// Utility that predicates the body a tiled loop with a vector.predicate
+/// operation. The vectorization factor used for predication is assumed to be
+/// the step of the tiled loop.
+LogicalResult mlir::vector_ext::predicateTiledLoop(TiledLoopOp loopOp) {
+  if (loopOp.lowerBound().size() > 1)
+    // TODO: Support multi-dim tiled loops.
```
nicolasvasilache, line 116: Please don't go there too deeply without helpers that are completely independent of linalg.tiled_loop and that we later turn into interfaces, the multi-dim part is quite unlikely to survive at the level you'd create this op.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -0,0 +1,156 @@
...
+    Location loc = loopOp.getLoc();
+    auto minOp = builder.create<AffineMinOp>(
+        loc, loopOp.step()[0].getType(), maps[0],
+        ValueRange{loopOp.upperBound()[0], loopOp.getInductionVars()[0]});
+
+    auto maskType = VectorType::get({vecFactor}, builder.getI1Type());
+    return builder.create<CreateMaskOp>(loc, maskType, ValueRange{minOp});
+  };
+
+  OpBuilder builder(loopOp);
```
nicolasvasilache, line 146: Builder creation within the op will make it hard to compose, please pass from above.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
nicolasvasilache: Hmm I still see this anchoring on `TiledLoopOp`.
Can this be implemented on Operation* with some helper function that would later turn into an interface on the right ops?

nicolasvasilache (commented):
I'm fine iterating with the current form to build experience.

Still I am wondering whether this vector-predicating should directly transform the linalg.generic instead (in some future CL)?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
nicolasvasilache: yes, it's to avoid creating a builder internally because that would mess up a PatternRewriter user of the API.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
nicolasvasilache: There is an ArithBuilder with similar purpose.
We can move it next to it.

Ideally I'd wait for affine_apply / min / max to move to arith but it will likely take time: https://llvm.discourse.group/t/rfc-arith-dialect-versions-of-affine-apply-min-max/4656

But if my choice is between using verbose APIs and moving to core before it is in the right place, I'd bias towards moving to core :)

nicolasvasilache (approved):

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: Look at `predicateOp` above. That's the generic implementation on `Operation*`. `predicateTiledLoop` is just extracting the information needed from the TileLoopOp and the invoking the generic implementation (line 147).

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: multi-dim support should be transparent to the generic `predicateOp` implementation. It should only impact the way we create the predicate (the lambda function below).

dcaballe (commented):
> Still I am wondering whether this vector-predicating should directly transform the linalg.generic instead (in some future CL)?

We have to predicate the whole loop body because predication will allow the simplification of the affine.min and insert/extract, which is required to enable vectorization.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
nicolasvasilache: nice, thanks!

PR #65, opened 2021-11-30
Make LowerVectors a single multi-stage Transform

ftynse:
Historically, `LowerVectors` has been parameterized by the vector
lowering stage (one out of 7). This leads to unnecessary complexity when
composing transformations that requires factories and non-trivial Python
constructs just to reuse one three-line method from the base `Transform`
class. Reimplement this method to run the pass pipeline as many times as
there are stages instead. `LowerVectors` is now directly composable with
other `Transform`s and can run an arbitrary subset of stages.

### DIFF
--- a/python/conv/conv_1d_bench.py
+++ b/python/conv/conv_1d_bench.py
@@ -26,7 +26,7 @@ all_experts = [
             tile_sizes=[1, 8, 32, 1, 8]) +\
           Vectorize(fun_name, op_name) +\
           Bufferize() +\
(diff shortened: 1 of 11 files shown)


gysit (approved):

nicolasvasilache (commented):
Just to be sure, this still allows printing IR after each stage, right ?
This is the reason I added those stages in the first place to have a clear picture of what happens when these various levels of pattern staging are applied one by one.

ftynse:
This still allows one to create `LowerVectors(0)`, `LowerVectors(1)`, etc. and have printing in between those.

PR #66, opened 2021-11-30
Move more Transform configuration knobs to be Variables

ftynse:
This exposes them to search and also makes composition cleaner. It
required to introduce a new class for choice variables.


gysit (approved):

nicolasvasilache (commented):
Thanks!

PR #73, opened 2021-12-02
Fix missing vector dialect registration from LinalgFuse driver.

MaheshRavishankar:
Also add some unit tests for the single tiling expert driver and
linalg fuse driver.

### DIFF
--- a/include/Transforms/Passes.td
+++ b/include/Transforms/Passes.td
@@ -44,7 +44,8 @@ def LinalgFuse : Pass<"linalg-fuse", "FuncOp"> {
   let dependentDialects = [
     "::mlir::arith::ArithmeticDialect", "::mlir::AffineDialect",
     "::mlir::linalg::LinalgDialect", "::mlir::scf::SCFDialect",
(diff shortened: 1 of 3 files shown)


gysit (approved):
nice! need to add these to the test script somehow...

MaheshRavishankar:
That'd be great. Any idea how to do that?

gysit:
> That'd be great. Any idea how to do that?

I will give it a try. There is probably a right way to do it or I can just call lit from python :)

MaheshRavishankar:
IREE has some cmake macro for the tests, but might be too involved for simple testing of this sort

gysit:
yes, I think it would be better to include it in the run_tests.py which is already run by a github action. Although cmake may also be a good option.

MaheshRavishankar:
> yes, I think it would be better to include it in the run_tests.py which is already run by a github action. Although cmake may also be a good option.

If we could get lit testing working with python that would be great too. The command line is pretty simple, last I looked at it seemed very complicated to do internally. Maybe we can sidestep all of that in OSS. If you add this to `run_test.py`. I can figure out how to add it for python

PR #74, opened 2021-12-02
Reorganize python files

nicolasvasilache:
This commit is in preparation for adding python bindings for our experimental dialects
so we can prototype the e2e execution of the linalg_ext ops from python.


README.md
```diff
@@ -117,89 +117,3 @@ TODOs:
 
 1.  hook up a lit test target.
 2.  re-add npcomp instructions once it is upgraded to use the same build setup.
-
```
ftynse, line 120: Are the related files removed completely? If not, may be move this to a README.md in the directory with those files.

nicolasvasilache: Yes they are removed as per Deny's suggestion (internally).

ftynse (approved):

PR #82, opened 2021-12-03
Add a TilingToTileOp pass.

nicolasvasilache:


MaheshRavishankar:
@nicolasvasilache I pushed a couple of more commits to this branch (and hence the PR is also updated). The first commit fixes the verification issue you were hitting (was a verification error on `tensor.extract_slice` that is fixed now).

The second commit adds the implementation of the TilingInterface for LinalgOp using External Models. Basically we can implement the interface out of tree (without changing Linalg Op in any way). The down side is that for every op we want to try using the interface for, we need to register the interface separately. I think that is OK for now, its one line per op. When we move this to core, we can use some tablegen tricks to forward the implementation of the TilingInterface to methods in LinalgOp interface.

Note that the second commit needs an LLVM change : https://reviews.llvm.org/D115101 .

I just ran it to make sure it runs, but I havent looked deep into whether everything is correct. It seems to be from a first glance.

include/Dialects/LinalgExt/Passes.h
```diff
@@ -11,17 +11,24 @@
...
+class DialectRegistry;
+
 namespace linalg_ext {
 
 /// Creates a pass to drive tiling of LinalgExt operations.
 std::unique_ptr<OperationPass<FuncOp>>
 createLinalgExtTilingPass(ArrayRef<int64_t> tileSizes = {});
 
+std::unique_ptr<OperationPass<FuncOp>>
+createLinalgExtTilingToTileOpPass(int64_t tileSize = 0);
```
apaszke, line 23: How is that different from the function above?

include/Dialects/LinalgExt/Passes.td
```diff
@@ -27,6 +27,33 @@ def LinalgExtTiling : Pass<"linalg-ext-tiling", "FuncOp"> {
   ];
 }
 
+def LinalgExtTilingToTileOp : Pass<"linalg-ext-tiling-to-tile-op", "FuncOp"> {
+  let summary = "Pass to tiling TilingInterface operations to 1 linalg_ext.tile op.";
```
apaszke, line 31: typo? tiling instead of tile?

include/Dialects/LinalgExt/Passes.td
```diff
@@ -27,6 +27,33 @@ def LinalgExtTiling : Pass<"linalg-ext-tiling", "FuncOp"> {
...
+  let description = [{
+    Rewrite a TilingInterface operation to a linalg_ext.tile.
+
+    This transformation is meant to tile only a single dimension.
+    This is because TileOp is meant to remain a 1-D loop construct with parallel 
+    semantics. We specifically want to avoid TileOp becoming an n-D loop construct
+    in the absence of a very compelling motivation.
+    Unlike scf.for we do not need a one-shot multi-loop tiling when targeting TileOp.
+    Indeed the implicitly stacking terminator + outSlices operands bypass the 
+    traditional issues with inserting subset.insert/subset.extract op pairs.
```
apaszke, line 41: Can you specify what happens if you have multiple operations that satisfy `TilingInterface` in your program? The two interesting cases is when they are nested and when they are not nested I guess.

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -145,6 +145,27 @@ Operation *ReverseOp::getTiledImplementation(OpBuilder &builder,
...
+  bodyRegion->push_back(new Block);
+  Block &bodyBlock = bodyRegion->front();
+  bodyBlock.addArgument(builder.getIndexType());
+  bodyBlock.addArgument(builder.getIndexType());
+  bodyBlock.addArguments(outSlices.getType());
+
+  OpBuilder::InsertionGuard guard(builder);
+  builder.setInsertionPointToStart(&bodyBlock);
+  bodyBuilder(builder, result.location, bodyBlock.getArgument(0),
+              bodyBlock.getArgument(1), bodyBlock.getArguments().drop_front(2));
```
apaszke, line 165: should we `ensureTerminator`?

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
```diff
@@ -145,6 +145,27 @@ Operation *ReverseOp::getTiledImplementation(OpBuilder &builder,
...
+  result.addOperands(tileSize);
+  result.addOperands(outSlices);
+  result.addTypes(outSlices.getType());
+
+  Region *bodyRegion = result.addRegion();
+  bodyRegion->push_back(new Block);
+  Block &bodyBlock = bodyRegion->front();
+  bodyBlock.addArgument(builder.getIndexType());
+  bodyBlock.addArgument(builder.getIndexType());
+  bodyBlock.addArguments(outSlices.getType());
```
apaszke, line 160: The out arguments to the body don't have the same types as the outer arguments. E.g. if someone passes in `tensor<100xf32>` as an output of a `tile` op with 4 tiles, the out argument should have a type of either `tensor<25xf32>` or `tensor<?xf32>`.

lib/Dialects/LinalgExt/Transforms/TilingExternalModels.cpp
```diff
@@ -0,0 +1,113 @@
...
+template <typename LinalgOpTy>
+struct LinalgOpTilingInterface
+    : public TilingInterface::ExternalModel<LinalgOpTilingInterface<LinalgOpTy>,
+                                            LinalgOpTy> {
+  SmallVector<Value> getDestinationOperands(Operation *op, OpBuilder &b) const {
+    LinalgOp linalgOp = cast<LinalgOp>(op);
+    return linalgOp.getOutputOperands();
+  }
+
+  SmallVector<StringRef> getLoopIteratorTypes(Operation *op) const {
```
apaszke, line 48: There's no enum for iterator types? 😮

lib/Dialects/LinalgExt/Transforms/TilingExternalModels.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+    return iteratorTypes;
+  }
+
+  SmallVector<Range> getLoopBounds(Operation *op, OpBuilder &b) const {
+    LinalgOp linalgOp = cast<LinalgOp>(op);
+    return linalgOp.createLoopRanges(b, op->getLoc());
+  }
+
+  Operation *getTiledImplementation(Operation *op, OpBuilder &b,
```
apaszke, line 63: This function is beyond my understanding of linalg, so if you want a proper review, you'll probably need to ask somebody else 😕

lib/Dialects/LinalgExt/Transforms/TilingToTileOp.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    return success();
+  }
+
+private:
+  linalg::LinalgTilingOptions options;
+  linalg::LinalgTransformationFilter filter;
+  // Poor man's single pattern application to avoid infinite pattern
+  // application wihtout having to hack with attributes and filters.
+  mutable int64_t successfulApplicationCount = 1;
```
apaszke, line 100: Since you're decrementing perhaps rename to `remainingApplicationCount`? `successfulApplicationCount` sounds like it's counting up

test/LinalgExt/tiling-to-tile-op-e2e.mlir
```diff
@@ -0,0 +1,46 @@
...
+
+  //      CHECK: %[[C10:.*]] = arith.constant 10 : index
+  //      CHECK:   %[[I:.*]] = linalg.init_tensor {{.*}} : tensor<?x?xf32>
+  //      CHECK: linalg_ext.tile %[[C10]] outs(%[[I]]: tensor<?x?xf32>) -> (tensor<?x?xf32>) {
+  //      CHECK: ^bb0(%[[OFF:.*]]: index, %[[SZ:.*]]: index, %[[I_ITER:.*]]: tensor<?x?xf32>):
+  //      CHECK:   %[[in:.*]] = tensor.extract_slice %[[T]]{{.*}} : tensor<?x?xf32> to tensor<?x?xf32>
+  //      CHECK:   %[[out:.*]] = tensor.extract_slice %[[I_ITER]]{{.*}} : tensor<?x?xf32> to tensor<?x?xf32>
+  //      CHECK:   %[[RES:.*]] = linalg_ext.reverse dimensions(dense<0> : tensor<1xi64>) 
+  // CHECK-SAME:     ins(%[[in]] : tensor<?x?xf32>) outs(%[[out]] : tensor<?x?xf32>) : tensor<?x?xf32>
+  //      CHECK:   linalg_ext.tile_yield %[[RES]] : tensor<?x?xf32>
```
apaszke, line 20: Huh, I would expect the reverse op to _not_ be tileable in our understanding of it. It seems like you want the reverse of the last tile to update the place occupied by the first tile?

apaszke (commented):
Awesome! Left a few comments, but mostly LGTM.

include/Dialects/LinalgExt/Passes.h
nicolasvasilache: The function above tiles directly to scf.for while still using the interface.

nicolasvasilache: Added a commentto distinguish the two.

include/Dialects/LinalgExt/Passes.td
nicolasvasilache: The nested case is not yet a thing.

nicolasvasilache: Added some comment to explain this is experimental and not yet meant to be used with multiple ops.

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
nicolasvasilache: I don't think we want it here: when ambiguous it is better to require the builder do it.
In the scf::ForOp, there is a weird situation when sometimes it gets added for you and sometimes not, this is inconsistent and very surprising.
If anything I am leaning towards dropping it everywhere non-ambiguous.

lib/Dialects/LinalgExt/IR/LinalgExtOps.cpp
nicolasvasilache: I was a bit overzealous with the "semantically meaningful naming" .. the slices are only the BBArgs.
Adding a test to confirm.

test/LinalgExt/tiling-to-tile-op-e2e.mlir
MaheshRavishankar: if you are not doing in-place updates, the reverse op is tilable.

lib/Dialects/LinalgExt/Transforms/TilingExternalModels.cpp
```diff
@@ -0,0 +1,113 @@
...
+                                    ValueRange dest,
+                                    ArrayRef<OpFoldResult> offsets,
+                                    ArrayRef<OpFoldResult> sizes) const {
+    LinalgOp linalgOp = cast<LinalgOp>(op);
+    Location loc = op->getLoc();
+    AffineMap shapeSizesToLoopsMap = linalgOp.getShapesToLoopsMap();
+    auto allShapeSizes = linalgOp.createFlatListOfOperandDims(b, loc);
```
MaheshRavishankar, line 70: Sorry, I didnt add any comments when I initially wrote this. Would it be possible to comment this a little bit?

MaheshRavishankar (approved):

lib/Dialects/LinalgExt/Transforms/TilingExternalModels.cpp
nicolasvasilache: Originally, MLIR did not have the support and the person who added the support did not update Linag.
Add it? That's a good starter task for MLIR Core :)

test/LinalgExt/tiling-to-tile-op-e2e.mlir
nicolasvasilache: I would say it is tilable but not "inplace".
Bufferization could not happen inplace here, this is similar to what a transposition would be: not inplace (see https://github.com/google/iree-llvm-sandbox/tree/main/python/examples/transpose).

lib/Dialects/LinalgExt/Transforms/TilingExternalModels.cpp
nicolasvasilache: Rewrote pieces, I'll do post commit if someone has comments.

PR #83, opened 2021-12-03
Integrating alp with iree-llvm-sandbox

giuseros:
This is the first integration for our project in `alp/experimental`. I tried to make the minimal number of changes outside of this folder, but:
- I still wanted to integrate with `mlir-proto-opt`
- I still want to be able to switch the  `experimental/alp` folder from the root `CMakeLists.txt` (although everything is disabled by default)
All of this is super-early stage and it is very buggy and not tested. However, I wanted to show earlier than later how we would like to integrate with this, and what were the main ideas. In particular:
- I integrate at the c++ level (so mostly reusing `mlir-proto-opt`, but not reusing the Pyhton harness you have)
- The major components are there (minimal python harness+autotuner+few simple passes). Once I have it cleaned up, I can add a `README.md` to explain how to use it. 
- If you have any early comment on anything, please feel free to add them or to open an issue

Thank you so much,
Giuseppe

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,13 @@ set(MLIR_MAIN_SRC_DIR ${LLVM_MAIN_SRC_DIR}/../mlir)
 set(MLIR_INCLUDE_DIR ${LLVM_MAIN_SRC_DIR}/../mlir/include)
 set(MLIR_TABLEGEN_OUTPUT_DIR ${CMAKE_BINARY_DIR}/tools/mlir/include)
 
+if (SANDBOX_ENABLE_ALP)
+  # ALP includes
+  add_compile_definitions("SANDBOX_ENABLE_ALP")
+  include_directories(experimental/alp/include)
+  include_directories(${CMAKE_BINARY_DIR}/tools/sandbox/experimental/alp/include)
+endif()
+
 list(APPEND CMAKE_MODULE_PATH ${MLIR_MAIN_SRC_DIR}/cmake/modules)
 list(APPEND CMAKE_MODULE_PATH ${LLVM_MAIN_SRC_DIR}/cmake)
 set(MLIR_TABLEGEN_EXE mlir-tblgen)
@@ -80,3 +87,8 @@ add_subdirectory(lib)
 add_subdirectory(python)
 add_subdirectory(test)
 add_subdirectory(tools)
+if (SANDBOX_ENABLE_ALP)
+# ALP cmake files
+add_subdirectory(experimental)
+endif()
+
--- a/experimental/alp/CMakeLists.txt
+++ b/experimental/alp/CMakeLists.txt
@@ -1,0 +1,14 @@
+get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
+get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
+
+set(LLVM_LINK_COMPONENTS
+  Core
+  Support
+  nativecodegen
+  native
+  OrcJIT
+  )
+
+include_directories(include/)
+add_subdirectory(include)
+add_subdirectory(lib)
--- a/experimental/alp/alp/compile_op.py
+++ b/experimental/alp/alp/compile_op.py
@@ -1,0 +1,219 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See https://llvm.org/LICENSE.txt for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+import os
+import tempfile
+import shutil
+from pathlib import Path
+from .utils import print_command, run_and_save, run_command, add_extension
+from .library.blas import gemm
+
+def build_main_obj(benchmark_prog, m, n, k, op, reps, mktmp_fn):
+    benchmark_prog = benchmark_prog.replace("_M_", str(m))
+    benchmark_prog = benchmark_prog.replace("_K_", str(k))
+    benchmark_prog = benchmark_prog.replace("_N_", str(n))
+    benchmark_prog = benchmark_prog.replace("__OP__", op)
+    benchmark_prog = benchmark_prog.replace("_REPS_", str(reps))
+
+    main_mlir = mktmp_fn("test.mlir")
+    main_mlir_lowered = mktmp_fn("test.llvm.mlir")
+    main_llvm = mktmp_fn("test.ll")
+    main_obj = mktmp_fn("test.o")
+
+    f = open(main_mlir, "w")
+    f.write(benchmark_prog)
+    f.close()
+
+    # main program
+    cmd = ["mlir-opt"]
+    cmd.append(main_mlir)
+    cmd.append("--linalg-bufferize")
+    cmd.append("--std-bufferize")
+    cmd.append("--tensor-constant-bufferize")
+    cmd.append("--tensor-bufferize")
+    cmd.append("--func-bufferize")
+    cmd.append("-convert-linalg-to-affine-loops")
+    cmd.append("-lower-affine")
+    cmd.append("-convert-scf-to-std")
+    cmd.append("-convert-memref-to-llvm")
+    cmd.append("-convert-std-to-llvm")
+    cmd.append("-reconcile-unrealized-casts")
+    cmd.append(f"> {main_mlir_lowered}")
+    run_command(cmd)
+    print_command(cmd)
+
+    cmd = ["mlir-translate"]
+    cmd.append("--mlir-to-llvmir")
+    cmd.append(f"{main_mlir_lowered}")
+    cmd.append(f"> {main_llvm}")
+    run_command(cmd)
+
+    cmd = ["llc"]
+    cmd.append(f"{main_llvm}")
+    cmd.append("-O3")
+    cmd.append("-filetype=obj")
+    cmd.append(f"-o {main_obj}")
+    run_command(cmd)
+
+
+def apply(transform_list, op_mlir_file, verbosity_level):
+    cmd = ["$IREE_LLVM_SANDBOX_BUILD_DIR/bin/mlir-proto-opt "]
+    for t in transform_list:
+      if not t:
+        continue
+      if type(t) is tuple:
+        (l, ext) = t
+        if l >= verbosity_level:
+          run_and_save(cmd, op_mlir_file, add_extension(op_mlir_file, ext))
+      else:
+        cmd.append(t)
+
+    output = add_extension(op_mlir_file, "llvm")
+    run_and_save(cmd, op_mlir_file, output)
+    return output
+    
+def SaveIR(x, ext):
+  return (x, ext)
+
+def build_operator_obj(op_prog, m, n, k, op, option_list, mktmp_fn, verbosity_level=0):
+    op_prog = op_prog.replace("_M_", str(m))
+    op_prog = op_prog.replace("_K_", str(k))
+    op_prog = op_prog.replace("_N_", str(n))
+
+    op_mlir = mktmp_fn(f"{op}.mlir")
+    f = open(f"{op_mlir}", "w")
+    f.write(op_prog)
+    f.close()
+
+    # Transformation options
+    tile_sizes = option_list["tile_sizes"]
+    reorder_tile_sizes = option_list["reorder_tile_sizes"]
+    register_tile_sizes = option_list["register_tile_sizes"]
+    reorder_register_tile_sizes = option_list["reorder_register_tile_sizes"]
+    hoist_packing = option_list['hoist_packing']
+    split_vector_transfer = option_list['split_vector_transfers_to']
+    extract_micro_kernel = option_list['extract_micro_kernel']
+    modulo_scheduling = option_list['modulo_scheduling']
+
+    Canonicalize = " --canonicalize --cse"
+    CodegenDriver = "--linalg-tensor-codegen-driver=\"anchor-func=gemm anchor-op=linalg.generic" 
+
+    # Transformations
+    OuterTiling = CodegenDriver + f" tile-sizes={tile_sizes} tile-interchange={reorder_tile_sizes}\"" + Canonicalize
+
+    InnerTiling = CodegenDriver + f" tile-sizes={register_tile_sizes} tile-interchange={reorder_register_tile_sizes}" + \
+                                  f" pad pack-paddings=1,1,0 hoist-paddings={hoist_packing} \"" + Canonicalize
+
+    DecomposeToLowerDimensionalNamedOp = CodegenDriver + " decompose-to-lower-dim\"" + Canonicalize
+
+    Vectorize = CodegenDriver + " vectorize vectorize-padding\"" + Canonicalize
+
+    Bufferize = "--linalg-bufferization-driver" + Canonicalize
+
+    LowerVector = "--linalg-vector-lowering=\"max-transfer-rank=1 " +\
+                 f" split-transfers={split_vector_transfer}" +\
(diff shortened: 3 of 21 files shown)


CMakeLists.txt
```diff
@@ -32,6 +32,15 @@ set(MLIR_MAIN_SRC_DIR ${LLVM_MAIN_SRC_DIR}/../mlir)
 set(MLIR_INCLUDE_DIR ${LLVM_MAIN_SRC_DIR}/../mlir/include)
 set(MLIR_TABLEGEN_OUTPUT_DIR ${CMAKE_BINARY_DIR}/tools/mlir/include)
 
+# Disable experimental alp by default
+set(SANDBOX_ENABLE_ALP OFF)
```
nicolasvasilache, line 36: I think this will always override whatever you give when calling the cmake command?

CMakeLists.txt
```diff
@@ -32,6 +32,15 @@ set(MLIR_MAIN_SRC_DIR ${LLVM_MAIN_SRC_DIR}/../mlir)
...
+if (SANDBOX_ENABLE_ALP)
```
nicolasvasilache, line 37: I am fine with this level of integration, I agree for projects that want it it is good to be buildable and integrated from the get go.
The fact that it is optional and cannot break the CI SGTM!

experimental/alp/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
+get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
+
+set(LLVM_LINK_COMPONENTS
+  Core
+  Support
+  nativecodegen
```
nicolasvasilache, line 7: I think native for you is AArch64 ?
For us it is still X86.
I don't think it is a problem for now but we should also think about setting things up so we can cross-compile and look at assembly at some ponit.
I am interested in seeing arm assembly even if I cannot run it myself.

experimental/alp/alp/compile_op.py
```diff
@@ -0,0 +1,216 @@
+import os
```
nicolasvasilache, line 1: I get that you didn't want to use the harness in this first CL but this is really hacky.

Ideally you'd be porting to using the harness later?

One of the reasons I shot for some amount of load bearing logic in python (the minimum possible) is that this is where I expect learning, stats, plotting etc to happen.

I am not interested in reimplementing SciKitLearn, Pandas, Pytorch or other tools that work well out of the box in C++ to be able to experiment.

experimental/alp/alp/library/blas.py
```diff
@@ -0,0 +1,135 @@
+typest = """
```
nicolasvasilache, line 1: Same comment re using the harness and metaprogramming IR from python in the future: this is really hacky (fine for a first commit).

experimental/alp/alp/mlirc.py
```diff
@@ -0,0 +1,57 @@
+import sys
```
nicolasvasilache, line 1: Same comment re using the harness and metaprogramming IR from python in the future: this is really hacky (fine for a first commit).

experimental/alp/alp/tuner.py
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env python
```
nicolasvasilache, line 1: Same comment re using the harness and metaprogramming IR from python in the future: this is really hacky (fine for a first commit).

This one should really be integrated closely with both the experts and the problem definitions.

There have been a lot of recent refactorings and QoL improvements by @ftynse and it will be much more future-proof to integrate on top of what he did.

experimental/alp/include/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(alp)
```
nicolasvasilache, line 1: Trailing newline missing here and other places.

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
```diff
@@ -0,0 +1,186 @@
...
+  if (add_yield){
+    rewriter.create<scf::YieldOp>(loc);
+  }
+}
+
+
+// Walk the for loops and find the one that as operands. In GEMM is the micro-kernel.
+// TODO: we should have the linalg::split to signal the microkernel of the operation
+// and use it to run the function extractor if needed
+struct ExtractKernelPass : public ExtractKernelPassBase<ExtractKernelPass> {
```
nicolasvasilache, line 157: I am unclear why this is needed per-se.
If your objective is to improve the level of control and try to scale this up, be aware that we are exploring a few different things that seem to not intersect with this pass.

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+    void runOnFunction() override
+    {
+      // Get the current FuncOp operation being operated on.
+      auto f = getFunction();
+
+      scf::ForOp loop;
+
+      //  Unroll the kernel
+      f.walk([&](scf::ForOp forop)
```
nicolasvasilache, line 43: Nit: A run of clang-format on this would restructure things very differently.
I locally added this to my `.bashrc`:

```
function git-format-add-and-amend(){
  echo "git add $1 && git show --name-only | egrep \"*.(\.cpp|\.h)\" | xargs -i clang-format --style=file -i {}; git add $1; git commit --amend"
  git add $1 && git show --name-only | egrep "*.(\.cpp|\.h)" | xargs -i clang-format --style=file -i {}; git add $1; git commit --amend
}
```

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
```diff
@@ -0,0 +1,136 @@
...
+             {
+               if (forop.getNumIterOperands())
+               {
+                 loop = forop;
+               }
+             });
+
+      if (loop)
```
nicolasvasilache, line 51: Structure-wise, we would try to layer this as:

1.  a simple helper function that takes a builder and returns the elements of IR that it creates/modify so that the APIs would compose. An example/motivation/analogy is: think of tiling that returns the newly tiled loops for further composition rather than void (which would force a subsequent analysis to re-traverse the IR and infer what are the tiled loops it wants to do something on).
2. ensure that helper function can be called with a RewritePattern builder and wrap it into a pattern that we can then expose.
3. later, hook up to mechanisms that control pattern application.

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "mlir/Dialect/SCF/Utils.h"
+#include "mlir/Dialect/Vector/VectorOps.h"
+#include "mlir/Transforms/LoopUtils.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+
+using namespace mlir;
+
+namespace{
+  struct ModuloSchedulingPass : public ModuloSchedulingPassBase<ModuloSchedulingPass>
```
nicolasvasilache, line 26: Not reviewing the pass itself for now, this would need a bunch of .mlir tests to qualify as "in reviewable state".

tools/mlir-proto-opt/mlir-proto-opt.cpp
```diff
@@ -54,17 +54,27 @@ static void registerIreeDialects(DialectRegistry &registry) {}
 #endif
 
 void registerTestPasses() { registerTestVectorMaskingUtils(); }
+#ifdef SANDBOX_ENABLE_ALP
```
nicolasvasilache, line 57: Could you directly add an extra level of indirection please ?

I.e. add `registerExperimentalPasses` and keep the `ifdefs` in that function.

This way future projects only have a single place to modify and the `main` does not grow up over time with mostly optional functions.

nicolasvasilache (approved):

nicolasvasilache:
```
I still wanted to integrate with mlir-proto-opt
I still want to be able to switch the experimental/alp folder from the root CMakeLists.txt (although everything is disabled by default)
```

SGTM, since everything is optional I don't see an issue.
We may want to restructure the CMake a bit so that the number of places touched by a future such addition is minimized
but this is fine for now and you don't need to shoulder this burden (I expect the next project will have some refactoring requests).

```
All of this is super-early stage and it is very buggy and not tested. However, I wanted to show earlier than later how we would like to integrate with this, and what were the main ideas. In particular:
I integrate at the c++ level (so mostly reusing mlir-proto-opt, but not reusing the Pyhton harness you have)
```

Sounds good for a first commit and to get started. I still think in the future the proper efforts should be made to either integrate with the harness or to come up with something better we can all reuse; I added comments where appropriate.
Note in particular the point about just reusing ScikitLearn/pandas/pytorch rather than having to reimplement a lot of functionality in C++ to just be able to do anything: composition is our friend, both in IR and with libraries.

```
The major components are there (minimal python harness+autotuner+few simple passes). Once I have it cleaned up, I can add a README.md to explain how to use it.
If you have any early comment on anything, please feel free to add them or to open an issue
```

I gave you review comments, no need to address everything now but please open the proper issues (prefixed with `[ALP]`) for the stuff you punt on.

nicolasvasilache:
Note I create and landed the alp subdir in an effort to try and give you rights on it specifically.
However it is not a default github mode and I have to parse through this:
https://stackoverflow.com/questions/40567468/give-permissions-on-project-folder-in-github?noredirect=1&lq=1

It will take me a little time as I have a bunch of stuff in my stack.

CMakeLists.txt
giuseros: Good point!

experimental/alp/alp/compile_op.py
giuseros: Yes, I agree. But I had major problems with the harness as it is was:
* I wanted to be able to go down to assembly without using the `mlir-c-runner`.
* I wanted to produce the intermediate results in different files
* In each file I wanted to have the `mlir-proto-opt` command to reproduce the IR (starting from linalg)
* I want also to be able to start the transformation pipeline in the middle (this is yet to be implemented, but very easy to do with the hacky harness)
So, yes, this was intended as a quick and hacky way to get these features reusing the main transformations. We are very focused on performance for now, so I need all the IRs in the simplest and quickest way to analyse them and understand what is wrong between one transform and the next. I promise that once we are satisfied with performance we will revisit this.

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
giuseros: It's unclear to me as well :) The only thing I know is that if we don't extract the kernel in a separate function there is significant spilling. This does not happen always, but for specific cases where we particularly stress the scheduler (e.g., 12x8 inner  tiling). We are investigating the LLVM scheduler, but this for now seemed the only ( yet hacky) solution

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
giuseros: Thanks for this, I will copy as is!

giuseros:
Thanks Nicolas for this first round of comments. I started to reply to some. Few overall replies:
* Our main focus for now is performance. We need to show that, by using this framework we can compete with handwritten routines.  We have  proof of this for a very specific case, but now we are trying to get more general results.
* Once we are happy with performance we will refactor everything to generalize to other operations. There are some features in the harness that I feel are missing, but we can try to add them

One question I didn't ask yet. Do you have any performance result for x86? Do they look promising?

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
giuseros: This is  a genuine question to learn about compiler development: what do you mean exactly? My idea for this pass was:
a) Look for the kernel loop
b) Apply modulo scheduling (I am writing a version to generalize to more than double buffering)
Is your suggestion to return the modified loop? Or also its prologue/epilogue ?

experimental/alp/alp/compile_op.py
nicolasvasilache: Understood, note that I don't consider any of these to be more than a few minutes to implement in the harness.
You could create a few more starter tasks like #9 and depending on priority we can help address.
You should be able at each step to print in a single file the mlir-proto-opt command embedded with the IR at that level (wherever the pipeline is applied is where you have the command and the IR).

Now your focus should indeed be to accelerate yourself. 
What I am suggesting is you may be hitting diminishing returns of maintaining a custom string-stitchy path.

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
nicolasvasilache: Ah ok, good to know, thanks for sharing. I have seen similar behavior myself in the past but not on these examples here. Makes sense to me, I'll think about it more.

Could you maybe add this rationale to the doc of you pass / in the cpp code somewhere ?

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
nicolasvasilache: This is probably a longer discussion but the basic idea is we try to apply transformations as rewrite patterns in a bunch of places so that they compose better. We specifically try to avoid that thinking derived from traditional compiler passes design brings us into corners re composition. This is related to the phase ordering problem and Cliff Click's 95 paper "Combining Analyses, Combining Optimizations". No need to go deeper for now, we'll have plenty of time to revisit when appropriate.

nicolasvasilache:
Re. `Our main focus for now is performance.`
Re. `Once we are happy with performance we will refactor everything`
Yes understood, the objective here is to give you some broad feedback so you have a few directions for the future.
You should absolutely optimize for your velocity at this point.

Still, it looks like you have a string-stichy and inflexible bandaid that limits you in ways you may not realize (what you have is very close to what the sandbox started with, we invested into making the QoL better for these reasons). You have to decide
what is best for your current sprint, I can only share my experience 😄  

Re x86, the various benchmarks run at a high fraction of peak on my AVX512 machine. I have previously dumped some perf results here:

  * matmul: https://gist.github.com/nicolasvasilache/9a526e6af1aae841a4f97d49f9b37db1
  * 1-d conv: https://gist.github.com/nicolasvasilache/356c1be72370f8d82b0fa6cadd7676b2
  * 1-d depthwise conv: https://gist.github.com/nicolasvasilache/570868fddf59c141a8647ec28de99aee

The matmul cases has mixes of divisible and non-divisble sizes; I get similar perf with `?` everywhere.
The conv / depthwise conv cases are all divisible, the first objective was to ensure we can get the main kernel to high-perf.
I have not yet tried to plug padding for non-divisible cases with those.
I also plugged and ran the 2-d cases for conv and depthwise conv to similar high perf (note that depthwise 1-D is much less arithmetic intensive and it is prob limited by instruction issue / number of instructions).

Lastly, I did some moderate amount of manual searching to get to these perf points, scaling up search (esp. for the kernel) and rooting out the type of issue you mention re outlining will be important for us too.

Hope this helps!

experimental/alp/alp/compile_op.py
giuseros: I created:
https://github.com/google/iree-llvm-sandbox/issues/92
https://github.com/google/iree-llvm-sandbox/issues/93
https://github.com/google/iree-llvm-sandbox/issues/94

nicolasvasilache:
could you please rebase/fix CI/land?

giuseros:
I tried to rebase and fix few things. Let's see if CI is happy.

giuseros:
Alright, CI is happy, but I am not authorized to merge

PR #86, opened 2021-12-06
Extend TestVectorMaskingUtils to predicate function operations

dcaballe:


test/lib/Dialect/VectorExt/TestVectorMaskingUtils.cpp
```diff
@@ -35,11 +35,47 @@ struct TestVectorMaskingUtils
...
+          if (elemType.isInteger(1))
+            return mask;
+        }
+
+        return Value();
+      };
+
+      OpBuilder builder(funcOp);
+      if (!predicateOp(builder, funcOp, &funcOp.body(),
+                       createPredicateForFuncOp))
```
dcaballe, line 76: I think this proofs that the utility is kind of generic!

dcaballe:
Ping :)

test/lib/Dialect/VectorExt/TestVectorMaskingUtils.cpp
```diff
@@ -35,11 +35,47 @@ struct TestVectorMaskingUtils
...
 
   void runOnFunction() override {
-    getFunction().walk([](TiledLoopOp loopOp) {
-      OpBuilder builder(loopOp);
-      if (failed(predicateTiledLoop(builder, loopOp)))
-        loopOp.emitError("Predication of tiled loop failed");
-    });
+    // Try different testing approaches until one triggers the predication
+    // transformation for that particular function.
+    bool predicationSucceeded = false;
```
gysit, line 40: Did you consider pass flags to control tiled loop vs function predication?

gysit (approved):

test/lib/Dialect/VectorExt/TestVectorMaskingUtils.cpp
dcaballe: Yes, the problem is that I have to add support for more ops and things will get complicated if we keep the tests in the same file (multiple RUN rules + different CHECK prefixes, etc.). We can revisit this when we add more tests or hit any limitation. Thanks for the feedback!

gysit: Sounds good!

test/VectorExt/vector_predication.mlir
```diff
@@ -50,3 +50,27 @@ func @tiled_loop_1d(%arg0: tensor<?xf32>,
 // CHECK-NEXT:        }
```
nicolasvasilache, line 50: `-` instead of `_` in test names (we can also do as a followup on all file names)

nicolasvasilache: also should this test filename specifically spell out "util" ?
Since this is not semantically preserving I would just call it "vector-predication".

test/lib/Dialect/VectorExt/TestVectorMaskingUtils.cpp
nicolasvasilache: very nice indeed, maybe spell out in the comment at the top of block that you go through these manipulations specifically to test the utility can be applied to a FuncOp body and this shows good reuse properties?

nicolasvasilache (approved):

nicolasvasilache:
thanks and sorry for the delay, been pulled into other things ..

nicolasvasilache:
ah damnit .. clicked the wrong button I am sorry, well I guess post-commit changes would do it :)

PR #95, opened 2021-12-09
 Add a primitive staged pattern rewrite driver

nicolasvasilache:


lib/Transforms/StagedPatternRewriteDriver.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+} // namespace
+
+LogicalResult
+mlir::applyStagedPatterns(ArrayRef<Operation *> roots,
+                          const FrozenRewritePatternSet &stage1Patterns,
+                          const FrozenRewritePatternSet &stage2Patterns,
+                          function_ref<LogicalResult(FuncOp)> stage3Lambda) {
+  if (roots.empty())
+    return failure();
```
ftynse, line 270: I'd say this is success. There is just nothing to do with no roots, which never produces errors.

lib/Transforms/StagedPatternRewriteDriver.cpp
```diff
@@ -0,0 +1,289 @@
...
+      driver.logger.startLine() << "Start applying stage 2 patterns\n";
+      driver.logger.indent();
+    });
+    if (failed(mlir::applyPatternsAndFoldGreedily(funcOp, stage2Patterns)))
+      return failure();
+    if (stage3Lambda && failed(stage3Lambda(funcOp)))
+      return failure();
+  }
+  return success();
+}
```
ftynse, line 289: Newline plz.

lib/Transforms/StagedPatternRewriteDriver.cpp
```diff
@@ -0,0 +1,289 @@
...
+        return success();
+      };
+
+      if (succeeded(matcher.matchAndRewrite(op, *this, canApply, onFailure,
+                                            onSuccess)))
+        res = success();
+
+      if (appliedThisRound)
+        break;
+    } while (!worklist.empty());
```
ftynse, line 190: Just to make sure the application logic that I understand from code is consistent with intent:
- if no patterns apply, an op is just removed from the worklist;
- if no patterns apply to any op, the entire application is considered a failure;
- one pattern can only ever apply to one op.

The first two seems reasonable. The last part feels too constraining. I understand we don't want repeated application of a pattern to the ops produced by said pattern, but we likely do want application of a pattern to independent ops that match (otherwise, we would need to insert as many pattern clones as there are ops that match it, doing the matching upfront).

lib/Transforms/StagedPatternRewriteDriver.cpp
```diff
@@ -0,0 +1,289 @@
...
+    assert(worklist[it->second] == op && "malformed worklist data structure");
+    worklist[it->second] = nullptr;
+    worklistMap.erase(it);
+  }
+}
+
+const char *logLineComment =
+    "//++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++//\n";
+
+class SinglePatternListener : OpBuilder::Listener {
```
ftynse, line 70: This isn't set on the builder, and duplicates the worklists from the `PatternRewriter`. Can we just have the rewriter also derive OpBuilder::Listener instead? + set the listener.

ftynse (changes requested):

lib/Transforms/StagedPatternRewriteDriver.cpp
nicolasvasilache: I am biased by the test I wrote:

```
// CASE3-LABEL: func @test3() {
func @test3() {
  // CASE3: "start"() {__test_attr__ = 1 : i32} : () -> ()
  // CASE3: "step_1"() : () -> ()
  "start"() : () -> ()
  "start"() : () -> ()
}
```

Basically I wanted the pattern to only apply to the first op but maybe that is not a constraint that should be put on this level of the IR / application hierarchy.

nicolasvasilache: what do you think of evolving this after we connect your stuff ?

ftynse: Let's just have a TODO that clearly explains what happens so we know easily where to go when revisiting this. My thinking is already at a point where we target operations more specifically in transformations, "the first op that matches" being only one of the possibilities.

ftynse (approved):

lib/Transforms/StagedPatternRewriteDriver.cpp
nicolasvasilache: Added more doc to the method, thanks!

PR #96, opened 2021-12-10
Fix VectorExt/Transform build

ftynse:
Virtually none of the library dependencies were listed in
CMakeLists.txt.

### DIFF
--- a/lib/Dialects/VectorExt/Transform/CMakeLists.txt
+++ b/lib/Dialects/VectorExt/Transform/CMakeLists.txt
@@ -5,5 +5,9 @@ add_mlir_library(MLIRVectorExtTransform
   MLIRVectorExt
 
   LINK_LIBS PUBLIC
+  MLIRAffine
   MLIRIR
+  MLIRLinalg
+  MLIRVector
+  MLIRVectorExt
 )


nicolasvasilache (approved):

nicolasvasilache:
Thanks!

Note am starting a fix to bump LLVM (minor test breakage due to memref verif tightening).
If you have them lmk, otherwise I'll just do it land in the next 15-30 minutes.

ftynse:
> Note am starting a fix to bump LLVM (minor test breakage due to memref verif tightening).

I just needed the project to build, my own tests don't run yet :)

dcaballe:
Thanks! Are we doing any build check as part of CI? Could we add a build with dynamic linking to avoid this problem?

nicolasvasilache:
@dcaballe file a task please ?

dcaballe:
I created this issue: https://github.com/google/iree-llvm-sandbox/issues/97

> I think CI only builds static libs. What about just switching it to build dynamic libs instead? Are there any issues that wouldn't be caught in this case?

Yes, that could also work. We just need to add the flag here: https://github.com/google/iree-llvm-sandbox/blob/main/configure.py#L137

PR #98, opened 2021-12-10
Enabled shared libs in configure.py

dcaballe:
Build is constantly broken since we are building with static libs
by default, which hides missing lib dependencies. This patch enables
shared libs by default to make sure we catch any missing dependency
in CI before landing any PR.


ftynse (approved):

nicolasvasilache:
let's test this live :) !

PR #100, opened 2021-12-10
Run vectorization twice for convolution to obtain good performance.

gysit:
At the moment, we need to run vectorization twice and only in the second run lower the pad tensor ops. The canonicalization and enabling pass running in between unlocks a better pad tensor op vectorization.

### DIFF
--- a/python/examples/conv/conv_2d_bench.py
+++ b/python/examples/conv/conv_2d_bench.py
@@ -24,7 +24,16 @@ all_experts = [
             op_name=op_name,
             #           N  H  W  C  KH  KW  F
             tile_sizes=[1, 1, 8, 32, 1, 1, 8],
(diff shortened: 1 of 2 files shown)


nicolasvasilache:
It would be nice to dump some patterns somewhere so we can improve vectorization.
I'd go for a github issue in this repo and a gist dump of IR.
Once LLVM core is done migrating from bugzilla to github we can move it there.

nicolasvasilache (approved):

gysit:
> It would be nice to dump some patterns somewhere so we can improve vectorization. I'd go for a github issue in this repo and a gist dump of IR. Once LLVM core is done migrating from bugzilla to github we can move it there.

Good point! I will generate an issue.

PR #105, opened 2021-12-10
[VectorExt] Introduce vector masking utility

dcaballe:
This patch introduces a utility to lower the vector.predicate op by
materializing the masks derived from its predicate. For now, we only
mask vector xfer ops but we should add support for masking more operations
and different lowering strategies.


lib/Dialects/VectorExt/Transform/VisitorUtils.h
```diff
@@ -0,0 +1,168 @@
...
+#include <utility>
+
+// This file defines generic (pre/in/post)-order MLIR IR visitors/walkers. The
+// walk() utility that MLIR core provides traverses operations in a block/
+// blocks in a region in the program order, and these walkers do the same. When
+// operations have regions attached to them, the core MLIR walkers visit the
+// regions attached to an Op first, and then visit the op. So within the context
+// of a single Op, the traversal is post-order (considering the Op as the parent
+// node and regions as the children). For certain use cases, it may be more
+// efficient/desirable to visit the parent Op before visiting the attached
```
nicolasvasilache, line 28: The need for such infra resurfaces time and again.
Can we split this in a proper independent PR with its own tests + test lib that will essentially print the name of ops traversed and cover all the cases of interest?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -154,3 +154,107 @@ LogicalResult mlir::vector_ext::predicateTiledLoop(OpBuilder &builder,
...
+      auto newXferOp = builder.create<TransferReadOp>(
+          op->getLoc(), xferReadOp.getType(), xferReadOp.source(),
+          xferReadOp.indices(), xferReadOp.permutation_map(),
+          xferReadOp.padding(), activeMask, xferReadOp.in_boundsAttr());
+      xferReadOp.replaceAllUsesWith(newXferOp.getResult());
+      xferReadOp.erase();
+      return;
+    }
+
+    // Mask vector.transfer_read.
```
nicolasvasilache, line 249: typo: write

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -154,3 +154,107 @@ LogicalResult mlir::vector_ext::predicateTiledLoop(OpBuilder &builder,
...
+LogicalResult mlir::vector_ext::maskVectorPredicateOps(
+    OpBuilder &builder, Operation *op, GenericOpMaskingStrategy maskGenericOp) {
+  OpBuilder::InsertionGuard guard(builder);
+
+  // Initialize stack of active masks. An empty stack means that the active mask
+  // is all-ones (masking is not required).
+  SmallVector<Value, 8> activeMasks;
+  activeMasks.push_back(Value());
+
+  genericWalk(op, [&](Operation *op, const WalkStage &stage) {
```
nicolasvasilache, line 217: I was pondering whether to invert control to avoid passing WalkStage around (i.e. passins multiple lambdas, e.g. the "doBeforeAllRegions" lambda).

But the way you have it now seems better as we add new ops: we don't want the interleaving of partial control at a distance to become hard to follow; the price is we manipulate WalkStage explicitly.

So more a Q than a comment: have you thought about this (or other tradeoffs) and do you see a reasonable way to avoid leaking the WalkStage ?

One issue in particular is that this is not compatible with pattern-based and will likely be more difficult to integrate in the upcoming "more flexible control" mechanisms that we are iterating on with @ftynse .

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -154,3 +154,107 @@ LogicalResult mlir::vector_ext::predicateTiledLoop(OpBuilder &builder,
...
+    const WalkStage &stage) {
+  // Nothing to do. All-ones mask to apply.
+  if (!activeMask)
+    return;
+
+  if (stage.isBeforeAllRegions()) {
+    // Mask vector.transfer_read.
+    if (auto xferReadOp = dyn_cast<TransferReadOp>(op)) {
+      builder.setInsertionPoint(op);
+      auto newXferOp = builder.create<TransferReadOp>(
```
nicolasvasilache, line 240: setIP without a Builder::InsertionGuard is a big nono for me :) 
Could you please add a guard?

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -154,3 +154,107 @@ LogicalResult mlir::vector_ext::predicateTiledLoop(OpBuilder &builder,
...
+          op->getLoc(), xferReadOp.getType(), xferReadOp.source(),
+          xferReadOp.indices(), xferReadOp.permutation_map(),
+          xferReadOp.padding(), activeMask, xferReadOp.in_boundsAttr());
+      xferReadOp.replaceAllUsesWith(newXferOp.getResult());
```
nicolasvasilache, line 244: Copy/pasting here: One issue in particular is that this is not compatible with pattern-based and will likely be more difficult to integrate in the upcoming "more flexible control" mechanisms that we are iterating on with @ftynse .

nicolasvasilache (changes requested):
Let's please split the fancier walkers in a separate commit with proper tests.

The comments about patterns can be ignored short-term.

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
dcaballe: Note that the `genericWalk` code is a copy-paste from the TF repo (see comment at the top VisitorUtils.cpp/h). I think Rahul Joshi (its owner) had plans to upstream this to core.

IMO, to give this a more pattern-rewriter feeling maybe we could describe the visits for each operation with a class implementing methods like: `doBeforeRegionX`, `doAfterRegionY`, etc. We could then populate a list of these classes with all the visits we want to apply. The main problem I see with this approach is that the visitor algorithm is generic and it should be able to deal with an arbitrary number of regions per op. Mapping each class method to its specific region would require some thought.

I'll ping Rahul about this

lib/Dialects/VectorExt/Transform/VisitorUtils.h
dcaballe: Same as before: Note that the genericWalk code is a copy-paste from the TF repo (see comment at the top VisitorUtils.cpp/h). I think Rahul Joshi (its owner) had plans to upstream this to core.

nicolasvasilache: ah I missed that sorry, LGTM to land then and we'll iterate towards a good solution with core and reuse.

nicolasvasilache (approved):

PR #117, opened 2021-12-15
Add an initial version of the Linalg Transform pass

ftynse:
This pass is intended to control Linalg transformations at a finer
granularity and to be able to serialize transforamtion specifications as
IR. Currently, it matches the control approach adopted by
CodegenStrategy, i.e., targeting ops of a specific kind in a named
function when applicable, but this is subject to future evolution.


nicolasvasilache (approved):
Nice! I'll let your rebase and merge.

ftynse:
So it appears that CI checks out LLVM HEAD, not the pinner version...

PR #118, opened 2021-12-15
Revisit cmake and registration.

nicolasvasilache:


nicolasvasilache:
Landing as is, feel free to iterate and land to your liking, the CLI roundtrip for this is counterproductive: we'll test it live ! :)

PR #119, opened 2021-12-15
Update Linalg header from "LinalgOps.h" to "Linalg.h".

gysit:
Update the Linalg headers after https://reviews.llvm.org/D115728 and bumping the LLVM version.


ftynse (approved):
#117 may have introduced another use of the old header

gysit:
I already rebased on #117. Will land and post fix if I missed something.

Issue #109, opened 2021-12-13
Connect experimental/alp autotuning with harness' tuning hooks


shabalind:
Earlier, we experimented with simple randomised sampling of harness tuning variables (see [variables.py](https://github.com/google/iree-llvm-sandbox/blob/main/python/examples/core/variables.py) and their use in [transforms.py](https://github.com/google/iree-llvm-sandbox/blob/main/python/examples/core/transforms.py)), but eventually we decided against rolling our own tuning framework in the sandbox and it was removed in #76. 

The longer term idea was to use externally provided search such as OpenTuner that is now used in experimental/alp. Conceptually both rely on the same concept of tuning variables with fixed domains, so it seems like there is an opportunity for a common codebase. 

This issue is meant to be a discussion ground on whether it's a good idea, and what are the incremental steps to move towards that goal.

shabalind:
@giuseros Do you think you'd be interested in merging the efforts here to have a common OpenTuner-tunable infrastructure? Are there any other blockers apart from #93, #94, #95 that would prevent things from connecting?

shabalind:
@giuseros If you look at [examples/matmul/bench.py](https://github.com/google/iree-llvm-sandbox/blob/main/python/examples/matmul/bench.py) you'll see a few key things that you need to build an equivalent of your snippet. 

`test_harness` is roughly equivalent of `call_program` in your example. It evaluates a concrete configuration (in the same python process) and prints the performance results to stdout. Currently the function doesn't actually return anything yet, but it's easy to modify it to return performance numbers as well (PR coming soon from my side).

The configuraton is defined by an expert, which is a composition of a number of transformations (see [experts.py](https://github.com/google/iree-llvm-sandbox/blob/main/python/examples/core/experts.py)). Each transformation can be parametrized by variables (see [transforms.py](https://github.com/google/iree-llvm-sandbox/blob/main/python/examples/core/transforms.py)). Variables are roughly equivalent to OpenTuner parameters: they are typed and have a clear precise domain of all possible valid values.

Harness accepts an instantiation of expert to some concrete values and runs it on a given problem definition (i.e. matmul vs conv vs reduction vs ...) + sizes + information of wheather sizes should be passed statically or dynamically.

Once harness can be invoked programmatically to obtain performance numbers it looks like it should be easy to make an equivalent of the code snippet you provided above.

giuseros:
Hi @shabalind , 
[...]

PR #120, opened 2021-12-15
Make harness return performance results

shabalind:
As discussed in #109 there is an opportunity to connect OpenTuner
to tune parameters for existing experts.

This change makes test_harness return performance result as a nested
dictionary that contains the same information as what's currently
printed to stdout. As the result, it can be invoked programmatically.

### DIFF
--- a/python/examples/core/harness.py
+++ b/python/examples/core/harness.py
@@ -26,8 +26,11 @@ def log(*args):
   sys.stderr.flush()
 
 
+TimingResults = Mapping[str, Sequence[float]]
+
+
 def timed_invoke(run_n_iters: Callable, gflop_count: float, gbyte_count: float,
-                 n_iters: int) -> Mapping[str, Any]:
+                 n_iters: int) -> TimingResults:
   elapsed_ns = run_n_iters(n_iters)
   elapsed_s = np.flip(np.sort(elapsed_ns / 1.e9))
   elapsed_s_per_iter = [                  \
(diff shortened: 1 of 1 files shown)


python/examples/core/harness.py
```diff
@@ -26,7 +27,7 @@ def log(*args):
 
 
 def timed_invoke(run_n_iters: Callable, gflop_count: float, gbyte_count: float,
-                 n_iters: int):
+                 n_iters: int) -> Mapping[str, Any]:
```
ftynse, line 30: Nit: isn't the mapped type always a `List[float]` ?

python/examples/core/harness.py
```diff
@@ -286,7 +297,7 @@ def test_harness(
             transform=expert,
             dump_ir_to_file=kwargs.get('dump_ir_to_file', ''))
 
-        problem.run(
+        results['expert'][expert] = problem.run(
```
ftynse, line 300: What about flattening the dictionary and using something like `expert.__name__` ? They names aren't very human-friendly, but that is one of the feature requests on them.

ftynse:
Please rebase on #121, this should solve the CI brekage.

python/examples/core/harness.py
shabalind: Improved signatures as requested.

python/examples/core/harness.py
shabalind: Good idea, I've used `str(expert)` as the key here which is more unique (includes a heap address hash) than just a type name.

ftynse (approved):

PR #121, opened 2021-12-15
[github] Read the LLVM commit from environment

ftynse:
The workflow was set up to read the LLVM commit from action outputs, but the action was writing it to the environment instead. The workflow has been always checking out LLVM HEAD instead of the pinned version.


ftynse:
Please do not merge until CI passes.

gysit (approved):

PR #124, opened 2021-12-16
Introduce transform-aware canonicalization

ftynse:
Introduce a new variant of the greedy pattern rewrite driver, adapted
from upstream MLIR since the class itself is an implementation detail
and cannot be derived. This variant is aware of the mapping between
values used in the Transform dialect and the operations and updates it
if the mapped operations are being updated by canonicalization patterns
(which does happen, e.g., after tiling to propagate static shape
information into types). This driver allows the transformation
interpreter to run a "canonicalization" pass without resorting to actual
passes and without risking to lose track of the operations that are
being transformed.

### DIFF
--- a/lib/Dialects/LinalgTransform/Transforms/TrackingRewriteDriver.cpp
+++ b/lib/Dialects/LinalgTransform/Transforms/TrackingRewriteDriver.cpp
@@ -1,0 +1,439 @@
+//===-- TrackingRewriteDriver.cpp - Pattern rewriter keeping track of ops -===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "Dialects/LinalgTransform/TrackingRewriteDriver.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Rewrite/PatternApplicator.h"
+#include "mlir/Transforms/FoldUtils.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "mlir/Transforms/RegionUtils.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/ScopedPrinter.h"
+
+#define DEBUG_TYPE "tracking-rewrite-driver"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE << "]: ")
+
+using namespace mlir;
+using namespace mlir::linalg;
+
+namespace {
+
+/// Find the linalg op that defines all values in range, potentially
+/// transitively through tensor casts.
+static LinalgOp findSingleLinalgOpDefiningAll(ValueRange range) {
+  LinalgOp sourceOp = nullptr;
+  for (Value value : range) {
+    // See through tensor casts.
+    //
+    // TODO: we may need some generalization (interfaces?) of this for other
+    // operations, especially multi-operand ones to understand which of their
+    // operands may be coming from a Linalg op. Or a completely different
+    // mechanism of tracking op replacement at creation, or even different
+    // patterns that identify the "main" result of a transformation.
+    while (auto castOp = value.getDefiningOp<tensor::CastOp>())
+      value = castOp.source();
+
+    if (auto currentSourceOp = value.getDefiningOp<LinalgOp>()) {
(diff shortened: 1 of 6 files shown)


lib/Dialects/LinalgTransform/Transforms/TrackingRewriteDriver.cpp
```diff
@@ -0,0 +1,439 @@
...
+
+    for (auto &pair : trackedOperations)
+      trackedOperationKeys.try_emplace(pair.second, pair.first);
+  }
+
+  void replaceOp(Operation *op, ValueRange newValues) override {
+    if (isa<LinalgOp>(op)) {
+      LinalgOp replacement = findSingleSourceOp(newValues);
+      assert(replacement &&
+             "repacling a linalg op, but could not find the replacement op");
```
nicolasvasilache, line 78: typo

lib/Dialects/LinalgTransform/Transforms/TrackingRewriteDriver.cpp
```diff
@@ -0,0 +1,439 @@
...
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE << "]: ")
+
+using namespace mlir;
+using namespace mlir::linalg;
+
+namespace {
+
+/// Find the linalg op that defines all values in range, potentially
+/// transitively through tensor casts.
+static LinalgOp findSingleSourceOp(ValueRange range) {
```
nicolasvasilache, line 28: more explicit naming ? `findSingleLinalgOpDefiningAll` ?

lib/Dialects/LinalgTransform/Transforms/TrackingRewriteDriver.cpp
```diff
@@ -0,0 +1,439 @@
...
+                  llvm::unique_function<bool(OpOperand &) const>) override {
+    llvm_unreachable("TrackingRewriteDriver::replaceOpWithIf not implemented");
+  }
+
+private:
+  DenseMap<Value, Operation *> &trackedOperations;
+  DenseMap<Operation *, Value> trackedOperationKeys;
+
+  //--------------------------------------------------------------------------//
+  // The code below is taken from GreedyPatternRewriteDriver.
```
nicolasvasilache, line 115: Ok, so I am skipping reviewing until the matching `end` without risk of errors.
Please shout if incorrect.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -576,13 +577,45 @@ static LogicalResult executeSequence(linalg::transform::SequenceOp sequence,
...
   FuncOp func = target->getParentOfType<FuncOp>();
+  ModuleOp module = func->getParentOfType<ModuleOp>();
+
+  MLIRContext *ctx = target->getContext();
+  RewritePatternSet patternList(ctx);
+  for (Dialect *dialect : ctx->getLoadedDialects())
+    dialect->getCanonicalizationPatterns(patternList);
+  for (RegisteredOperationName op : ctx->getRegisteredOperations())
+    op.getCanonicalizationPatterns(patternList, ctx);
+  FrozenRewritePatternSet patterns(std::move(patternList));
```
nicolasvasilache, line 588: There are some tiling enabling patterns that I am not 100% sure are canonicalizations.
@Mogball's PatternSets could help organize things.
No change needed here.

nicolasvasilache (approved):
Nice!

lib/Dialects/LinalgTransform/Transforms/TrackingRewriteDriver.cpp
```diff
@@ -0,0 +1,439 @@
...
+    //
+    // TODO: we may need some generalization (interfaces?) of this for other
+    // operations, especially multi-operand ones to understand which of their
+    // operands may be coming from a Linalg op. Or a completely different
+    // mechanism of tracking op replacement at creation, or even different
+    // patterns that identify the "main" result of a transformation.
+    while (auto castOp = value.getDefiningOp<tensor::CastOp>())
+      value = castOp.source();
+
+    if (auto currentSourceOp = value.getDefiningOp<LinalgOp>()) {
```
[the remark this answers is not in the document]
ftynse: Nice catch!

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: Then they should not run here, which happens after every transformation, but within the tiling transformation itself IMO.

PR #125, opened 2021-12-16
Introduce transform-aware CSE

ftynse:
Introduce a new variant of common subexpression elimination, adapted
from upstream MLIR since the pass itself is an implementation detail and
cannot be derived. This variant can be run as a utility function and
does not plug into the pass infrastructure directly. It is also aware of
the operations being mapped to values used in the Linalg Transform
dialect and avoids deleting them so as not to break further
transformations.

### DIFF
--- a/lib/Dialects/LinalgTransform/Transforms/TrackingCSE.cpp
+++ b/lib/Dialects/LinalgTransform/Transforms/TrackingCSE.cpp
@@ -1,0 +1,292 @@
+#include "Dialects/LinalgTransform/TrackingCSE.h"
+#include "mlir/IR/Dominance.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
+#include "llvm/ADT/ScopeExit.h"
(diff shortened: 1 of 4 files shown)


lib/Dialects/LinalgTransform/Transforms/TrackingCSE.cpp
```diff
@@ -0,0 +1,284 @@
...
+/// Attempt to eliminate a redundant operation.
+LogicalResult TrackingCSE::simplifyOperation(ScopedMapTy &knownValues,
+                                             Operation *op,
+                                             bool hasSSADominance) {
+  // Don't simplify terminator operations.
+  if (op->hasTrait<OpTrait::IsTerminator>())
+    return failure();
+
+  // Never remove tracked ops.
+  if (trackedOps.count(op)) {
```
nicolasvasilache, line 110: could you please make it very visible with (comments) that this is the only place in the TrackingCSE that differs from the original impl?

Better yet, would it be possible to paste the original CSE and just change it slightly so that it can be derievd and the extra bahavior just be injected with a few lines?

nicolasvasilache (approved):

PR #126, opened 2021-12-16
Run PDL matching patterns once per operatio

ftynse:
Introduce a new helper function that wraps PatternApplicator so that
each pattern (in the case of Transform dialect, only the matching part
is actually relevant) is checked at most once against every operation.
No greedy recursive pattern application or worklist of newly created
operations is needed for this purpose. This allows us to stop relying
on the absence of the "matched" attribute in the IR to stop recursive
matching. The attribute itself is still added as a debugging aid.


nicolasvasilache (approved):
Do you see this + previous stuff connecting to #95 or it just being obsolete?

ftynse:
> Do you see this + previous stuff connecting to #95 or it just being obsolete?

Either obsolete (I pulled some ideas from there) or not-ready-yet (most of stage-1 transformations are not actually a single pattern, but a sequence of patterns+enablers). I'd still keep it for a while.

PR #127, opened 2021-12-16
Add vector legalization pass and fix some issues

giuseros:
- Adding a legalization pass that works across for-loops
- Change the scheduling pass to work with the legalized version
- Add the possibility to interleave MLIR inside the micro-kernel
- Fixed the Alp runtime to include functions we are using to print time and the PID of the program
- Fixed cmake infra

### DIFF
--- a/experimental/alp/lib/Transforms/extract_kernel_pass.cpp
+++ b/experimental/alp/lib/Transforms/extract_kernel_pass.cpp
@@ -32,9 +32,16 @@ void extract_function(StringRef func_name, Block *block, ModuleOp parentModule,
 
   // Create the function (callee site) with an empty block
   rewriter.setInsertionPointToStart(parentModule.getBody());
-  auto func_op = rewriter.create<FuncOp>(
-      parentModule.getLoc(), func_name,
-      FunctionType::get(parentModule.getContext(), Input, Output));
+  auto func_op = rewriter.create<FuncOp>(parentModule.getLoc(), func_name, FunctionType::get(parentModule.getContext(), Input, Output));
+  mlir::MLIRContext *ctx = parentModule.getContext();
+  llvm::SmallVector<mlir::Attribute, 4> attrs;
+  attrs.push_back(mlir::ArrayAttr::get(ctx,
+                                  {mlir::StringAttr::get(ctx, "prefer-vector-width"),
+                                  mlir::StringAttr::get(ctx, "128")}));
+  attrs.push_back(mlir::ArrayAttr::get( ctx, {mlir::StringAttr::get(ctx, "target-cpu"),
+                                          mlir::StringAttr::get(ctx, "thunderx2t99")}));
+  func_op->setAttr("passthrough", mlir::ArrayAttr::get(ctx, attrs));
+
   auto entry_block = func_op.addEntryBlock();
 
   // Build the dominance tree of the parent op of the block
@@ -53,8 +60,12 @@ void extract_function(StringRef func_name, Block *block, ModuleOp parentModule,
   // entry block are constants. For all other variables, we will add them as
   // inputs to the function
   block->walk([&](Operation *inst) {
-    for (Value val : inst->getOperands()) {
-      if (dom_info.properlyDominates(val, parent_op)) {
+    for (Value val: inst->getOperands()){
+      if (dom_info.properlyDominates(val, parent_op)){
+        if(func_name=="kernel_tail"){
+        llvm::errs()<<"dominates?\n";
+
+        }
         arith::ConstantOp const_op = val.getDefiningOp<arith::ConstantOp>();
         if (const_op) {
           // It's useless to add many times the same index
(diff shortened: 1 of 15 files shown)


experimental/alp/lib/Transforms/extract_kernel_pass.cpp
```diff
@@ -53,8 +60,12 @@ void extract_function(StringRef func_name, Block *block, ModuleOp parentModule,
   // entry block are constants. For all other variables, we will add them as
   // inputs to the function
   block->walk([&](Operation *inst) {
-    for (Value val : inst->getOperands()) {
-      if (dom_info.properlyDominates(val, parent_op)) {
+    for (Value val: inst->getOperands()){
+      if (dom_info.properlyDominates(val, parent_op)){
+        if(func_name=="kernel_tail"){
+        llvm::errs()<<"dominates?\n";
```
nicolasvasilache, line 66: debug spew ?

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
```diff
@@ -183,4 +194,149 @@ struct ExtractKernelPass : public ExtractKernelPassBase<ExtractKernelPass> {
...
+  //std::set<int64_t> consts;
+  llvm::SmallVector<Value, 1> vals;
+  llvm::SmallVector<arith::ConstantOp, 1> consts;
+
+  // Walk the block and find out all the variables that were defined outside 
+  // this block and are used inside the block (i.e., all the variables x that
+  // properly dominate the block). The only things we will redefine inside the
+  // entry block are constants. For all other variables, we will add them as 
+  // inputs to the function
+  block->walk([&](Operation *inst) {
```
nicolasvasilache, line 222: I've seen this implemented in various places as:
```
bool isDefinedOutsideOfX(Value value) {
  return !region().isAncestor(value.getParentRegion());
}
```

experimental/alp/lib/Transforms/legalize_vector.cpp
```diff
@@ -0,0 +1,211 @@
...
+              extractOp.getResult().replaceAllUsesWith(newForOp.getResults().back());
+              rewriter.eraseOp(extractOp);
+          }
+      }
+    }
+    return success();
+  }
+};
+
+struct LegalizePass : public LegalizePassBase<LegalizePass>{
```
nicolasvasilache, line 123: I am not too sure what this pass does as there are not tests but I'd say it looks a bit fishy and at the intersection of other things that should exist in core.
Maybe if you describe the objective here I can provide some links?

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
```diff
@@ -22,99 +22,146 @@
...
-        op->setAttr("unrolled_iteration", b.getUI32IntegerAttr(i));
-      };
+namespace{
+  enum StageType{
+    Compute,
+    Load
+  };
 
-      (void)loopUnrollByFactor(loop, 2, annotateFn);
+  struct ModuloSchedulingPass : public ModuloSchedulingPassBase<ModuloSchedulingPass>
```
nicolasvasilache, line 31: note that there is a loop pipelining pass that does similar things, could it be adapted to your uses?

nicolasvasilache (approved):

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
giuseros: If you refer to `ForLoopPipelining`, yes, it is exactly what I am using !

experimental/alp/lib/Transforms/legalize_vector.cpp
giuseros: We discussed this here: https://llvm.discourse.group/t/outer-product-spilling/4405/7. I am basically chunking writes, reads and contracts in chunks of 4. As you correctly pointed out [here](https://llvm.discourse.group/t/outer-product-spilling/4405/3),  `scf.for + iter_args throw a wrench into the unrolling patterns`. I added the code to remove the `wrench`

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
giuseros: Thanks Nicolas! Can I ask more info? What does this replace? I still need to walk every operation of the block, right?

giuseros:
@nicolasvasilache a friendly ping! I have another small set of transformations after this PR which got our performance to 94% of the library (finally).

nicolasvasilache:
@giuseros the PR is accepted, I won't be able to give you much deeper insights on the passes themselves at this time so please just merge at your convenience.
Once you have everything in and we can repro and dump some assembly we can iterate on generalizing some of the transforms; it is possible/likely that some of the stuff you have is also needed to bridge a perf gap I currently see on CPU.

nicolasvasilache:
> I have another small set of transformations after this PR which got our performance to 94% of the library (finally).

This sounds awesome!

giuseros:
Ah, so the problem is that I cannot merge it, as apparently I am not authorized to merge things in :) Sorry for the misunderstanding, I thought you were waiting to merge because wanted to add more replies

nicolasvasilache:
my apologies, it is still the subdir permission thing .. didn't get a cycle to look at that

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
```diff
@@ -183,4 +190,149 @@ struct ExtractKernelPass : public ExtractKernelPassBase<ExtractKernelPass> {
...
+  // Build the dominance tree of the parent op of the block
+  Region *region = block->getParent();
+  Operation *parent_op = region->getParentOp();
+  auto dom_info = mlir::DominanceInfo(parent_op);
+
+  //std::set<int64_t> consts;
+  llvm::SmallVector<Value, 1> vals;
+  llvm::SmallVector<arith::ConstantOp, 1> consts;
+
+  // Walk the block and find out all the variables that were defined outside
```
nicolasvasilache, line 213: @giuseros side note, did you see this older commit: https://reviews.llvm.org/D85449 ?

In particular the "defined outside" thing is a common core function.

PR #130, opened 2021-12-21
Finer-grain targeting in transform dialect

ftynse:
The transform dialect progressively relaxes the current approach in
codegen strategy that applies identical transformations to all
operations of the same kind in a given function. After introducing
PDL-based matching mechanism, now make this mechanism applicable for
each targetable operation directly instead of only the first operation
in a sequence. This makes it possible to apply transformations with
different options to different operations transparently.

### DIFF
--- a/include/Dialects/LinalgTransform/LinalgTransformOps.h
+++ b/include/Dialects/LinalgTransform/LinalgTransformOps.h
@@ -14,6 +14,28 @@
 #include "mlir/IR/OpDefinition.h"
 
 #include "Dialects/LinalgTransform/LinalgTransformOpsDialect.h.inc"
+#include <mlir/IR/BuiltinAttributes.h>
+
+namespace mlir {
+/// A trait for transform ops that can be targeted at either the result of a
+/// matcher (identified by its symbol name) or at the result of another
+/// transformation (identified by the value it produced).
+template <typename ConcreteOp>
+class TargetableTransformOpTrait
+    : public OpTrait::TraitBase<ConcreteOp, TargetableTransformOpTrait> {
+public:
+  static LogicalResult verifyTrait(Operation *op) {
+    Optional<SymbolRefAttr> matcher = cast<ConcreteOp>(op).matcher();
+    Value input = cast<ConcreteOp>(op).op();
+    if (!((matcher.hasValue() && matcher.getValue() != nullptr) ^
+          (input != nullptr)))
+      return op->emitOpError()
+             << "expects either an `op` operand or a `matcher` attribute";
+
+    return success();
+  }
+};
+} // namespace mlir
 
 #define GET_OP_CLASSES
 #include "Dialects/LinalgTransform/LinalgTransformOps.h.inc"
--- a/include/Dialects/LinalgTransform/LinalgTransformOps.td
+++ b/include/Dialects/LinalgTransform/LinalgTransformOps.td
@@ -45,32 +45,18 @@ def SequenceOp : Linalg_Transform_Operation<"sequence",
   }];
 }
 
-def ApplyOp : Linalg_Transform_Operation<"apply", [NoTerminator]> {
-  let description = [{A container operation that connects a sequence of
-  transformations to apply to the condition at which the operation should
-  be applied. The condition should be a PDLMatchOp producing the root
-  operation, which is forwarded as the unique basic block argument to the
-  `transforms` region of the operation.}];
-
-  let regions = (region SizedRegion<1>:$transforms, SizedRegion<1>:$condition);
-  let assemblyFormat = "$transforms `when` $condition attr-dict-with-keyword";
-}
-
-def PDLMatchOp : Linalg_Transform_Operation<"pdl_match"> {
-  let description = [{Specifies the name of the PDL pattern used to match the
-  root operation on which the transformations are applied.}];
-
-  let arguments = (ins SymbolRefAttr:$pattern);
-  let assemblyFormat = "$pattern attr-dict";
-}
-
 //===----------------------------------------------------------------------===//
 
-def TileOp : Linalg_Transform_Operation<"tile"> {
+def TargetableOpTrait : NativeOpTrait<"TargetableTransformOpTrait"> {
+  let cppNamespace = "::mlir";
+}
+
+def TileOp : Linalg_Transform_Operation<"tile", [TargetableOpTrait]> {
   let description = [{Indicates that ops of a specific kind in the given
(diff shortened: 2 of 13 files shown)


lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -671,85 +570,173 @@ makePDLRewriter(SmallVectorImpl<Operation *> &targets) {
...
 
-namespace {
-/// The only purpose of this class is to enable creation of PatternRewriter
-/// instances as the base class doesn't have a public constructor.
-class SimplePatternRewriter : public PatternRewriter {
-public:
-  explicit SimplePatternRewriter(MLIRContext *ctx) : PatternRewriter(ctx) {}
-};
-} // namespace
+/// Appends to `targets` the operations that `transformOp` should be run op.
```
nicolasvasilache, line 573: typo: `on`

include/Dialects/LinalgTransform/LinalgTransformOps.td
```diff
@@ -45,32 +45,18 @@ def SequenceOp : Linalg_Transform_Operation<"sequence",
...
 
-//===----------------------------------------------------------------------===//
-
-def TileOp : Linalg_Transform_Operation<"tile"> {
+def TileOp : Linalg_Transform_Operation<"tile", [TargetableOpTrait]> {
   let description = [{Indicates that ops of a specific kind in the given
   function should be tiled with the options provided as attributes.}];
 
-  let arguments = (ins PDL_Operation:$op,
+  let arguments = (ins Optional<PDL_Operation>:$op,
```
nicolasvasilache, line 58: Something more descriptive than "op" plz.

include/Dialects/LinalgTransform/LinalgTransformOps.h
```diff
@@ -14,6 +14,28 @@
 #include "mlir/IR/OpDefinition.h"
 
 #include "Dialects/LinalgTransform/LinalgTransformOpsDialect.h.inc"
+#include <mlir/IR/BuiltinAttributes.h>
+
+namespace mlir {
+/// A trait for transform ops that can be targeted at either the result of a
+/// matcher (identified by its symbol name) or at the result of another
+/// transformation (identified by the value it produced).
```
nicolasvasilache, line 22: Why a trait and not an interface ?
Plz add a comment to explain.

include/Dialects/LinalgTransform/LinalgTransformOps.td
```diff
@@ -45,32 +45,18 @@ def SequenceOp : Linalg_Transform_Operation<"sequence",
...
+                   OptionalAttr<SymbolRefAttr>:$matcher,
```
nicolasvasilache, line 59: also, something more descriptive than "matcher" plz.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -494,139 +495,37 @@ executeLowerToLLVMOp(FuncOp target,
...
-      // TODO: this may go away if we have a better scoping mechanism for large
-      // transformations such as bufferization and lowerings and track functions
-      // across them similarly to how we track linalg operations.
-      assert(&transform == &sequence.body().front().back() &&
-             "expected lowering to llvm to be the last transformation");
-      canonicalizationRoot = module;
-    }
+/// Apply at most one pattern from the given list to each operation nested in
+/// `parent`.
+static void applyPatternsOnce(Operation *parent,
```
nicolasvasilache, line 512: Found this quite confusing as it is very specific to PDL's way of doing things, this would be better as a lambda hidden within `findTransformTarget`.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -671,85 +570,173 @@ makePDLRewriter(SmallVectorImpl<Operation *> &targets) {
...
-  // later, at which point the IR transformations will have to notify this
-  // rewriter somehow. Alternatively, we could have only the matching part, but
-  // we would need directl access to PDLBytecode for that.
-  SimplePatternRewriter rewriter(parent->getContext());
-  parent->walk([&](Operation *op) {
-    rewriter.setInsertionPoint(op);
-    if (failed(applicator.matchAndRewrite(op, rewriter))) {
-      LLVM_DEBUG(DBGS() << "failed to match any pattern to " << *op << "\n");
+  pdlModule.registerRewriteFunction("linalg_transform.apply",
+                                    makePDLRewriter(targets));
```
nicolasvasilache, line 608: Same thing for this makePDLRewriter, would be better as a local lambda that  makes it clear it only appends.
Cross-function reading makes it much more confusing than it is worth.

nicolasvasilache: call the lambda `gatherOps` or something similar.

nicolasvasilache (approved):
Bunch of comments to address as followups to improve readability and reduce surprise.

include/Dialects/LinalgTransform/LinalgTransformOps.td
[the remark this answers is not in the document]
ftynse: What's the point of having a specific method that only forwards to an existing method? It just adds boilerplate.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: This is not specific to PDL but to the way PDL is used in the transform dialect. I find that a reasonable thing to assume in a static function of a file within transform dialect. Done.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: I will implement the change, but I quite strongly disagree with the rationale. Beyond the cases where (mostly short) lambdas decrease boilerplate, they negatively affect readability. There's trivially more horizontal whitespace because of them. It is no longer possible to just scan the file for all function definitions. It's easy to capture-world-by-reference without thinking and have definitions of the variables screens away from their uses, plus run the risk of use-after-free and UB that is virtually impossible with first-class functions. The functions become excessively long: instead of just assuming the called function does what its name says it does, one now has to read the entire body and keep the rest of the context in cache.

PR #137, opened 2022-01-04
Add a pass to target a loop enclosing a given op in a function and ou…

nicolasvasilache:
…tlining it via encapsulation in an scf::ExecuteRegion


nicolasvasilache:
My .gitconfig was wrong and the CLA is not properly recognized, closing this diff.

PR #139, opened 2022-01-05
Implement Listeners in CSE and Greedy Pattern Rewriter

Mogball:
In this PR, I've made a demo functional pattern implementation in C++ used it in parts of the LinalgTransform interpreter. Notably, the use of attribute filters and `findMarkedOps` is no longer needed. Although, these specific cases were uninteresting: since the `execute*` functions were passed the Operation handle directly... why not just call the pattern on the handle?

Ideally, mapping `Value` to operations would not be needed at all and the operation handles can be passed around directly. However, since not all of the interpreter transformations are targetable (yet) and I am unsure as to what certain transformations should be returning, I left them as is. If these issues are resolved, the interpreter will disappear entirely and `linalg_transform` operations will just be a description on how to construct and chain the functional patterns.

A few other things to note: I added a small scoping mechanism to restrict the application of patterns, and I reimplemented the tracking greedy rewriting using a listener. I also implemented a rollback mechanism for applying chains of patterns, but it's not used anywhere. It's very inefficient (maintains copies of the IR and associates them) because the dialect conversion system is unwieldy.

### DIFF
--- a/include/Transforms/ListenerCSE.h
+++ b/include/Transforms/ListenerCSE.h
@@ -1,0 +1,23 @@
+//===-- ListenerCSE.h - Common subexpr elimination with a listener --------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_IREE_SANDBOX_TRANSFORMS_LISTENERCSE_H
+#define LLVM_IREE_SANDBOX_TRANSFORMS_LISTENERCSE_H
+
+#include "Transforms/Listener.h"
+
+namespace mlir {
+class DominanceInfo;
+class Operation;
+
+LogicalResult eliminateCommonSubexpressions(Operation *op,
+                                            DominanceInfo *domInfo,
+                                            RewriteListener *listener);
+} // namespace mlir
+
+#endif // LLVM_IREE_SANDBOX_TRANSFORMS_LISTENERCSE_H
--- a/lib/Transforms/ListenerCSE.cpp
+++ b/lib/Transforms/ListenerCSE.cpp
@@ -1,0 +1,310 @@
+//===-- ListenerCSE.cpp - Common subexpr elimination with a listener ------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "Transforms/ListenerCSE.h"
+#include "mlir/IR/Dominance.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
+#include "llvm/ADT/ScopedHashTable.h"
+#include "llvm/Support/RecyclingAllocator.h"
+#include <deque>
+
+using namespace mlir;
+
+//===----------------------------------------------------------------------===//
+// BEGIN copied from mlir/lib/Transforms/CSE.cpp
+//===----------------------------------------------------------------------===//
+namespace {
+struct SimpleOperationInfo : public llvm::DenseMapInfo<Operation *> {
+  static unsigned getHashValue(const Operation *opC) {
+    return OperationEquivalence::computeHash(
+        const_cast<Operation *>(opC),
+        /*hashOperands=*/OperationEquivalence::directHashValue,
+        /*hashResults=*/OperationEquivalence::ignoreHashValue,
+        OperationEquivalence::IgnoreLocations);
+  }
+  static bool isEqual(const Operation *lhsC, const Operation *rhsC) {
+    auto *lhs = const_cast<Operation *>(lhsC);
+    auto *rhs = const_cast<Operation *>(rhsC);
+    if (lhs == rhs)
+      return true;
+    if (lhs == getTombstoneKey() || lhs == getEmptyKey() ||
+        rhs == getTombstoneKey() || rhs == getEmptyKey())
+      return false;
+    return OperationEquivalence::isEquivalentTo(
+        const_cast<Operation *>(lhsC), const_cast<Operation *>(rhsC),
+        /*mapOperands=*/OperationEquivalence::exactValueMatch,
+        /*mapResults=*/OperationEquivalence::ignoreValueEquivalence,
+        OperationEquivalence::IgnoreLocations);
+  }
+};
+} // namespace
+
+namespace {
+/// Simple common sub-expression elimination.
+struct CSE {
+  /// Shared implementation of operation elimination and scoped map definitions.
+  using AllocatorTy = llvm::RecyclingAllocator<
+      llvm::BumpPtrAllocator,
+      llvm::ScopedHashTableVal<Operation *, Operation *>>;
+  using ScopedMapTy = llvm::ScopedHashTable<Operation *, Operation *,
+                                            SimpleOperationInfo, AllocatorTy>;
+
+//===----------------------------------------------------------------------===//
+// END copied from mlir/lib/Transforms/CSE.cpp
+//===----------------------------------------------------------------------===//
+  CSE(DominanceInfo *domInfo, RewriteListener *listener)
+      : domInfo(domInfo), listener(listener) {}
+//===----------------------------------------------------------------------===//
+// BEGIN copied from mlir/lib/Transforms/CSE.cpp
+//===----------------------------------------------------------------------===//
+
+  /// Represents a single entry in the depth first traversal of a CFG.
+  struct CFGStackNode {
+    CFGStackNode(ScopedMapTy &knownValues, DominanceInfoNode *node)
+        : scope(knownValues), node(node), childIterator(node->begin()),
+          processed(false) {}
+
+    /// Scope for the known values.
+    ScopedMapTy::ScopeTy scope;
+
+    DominanceInfoNode *node;
+    DominanceInfoNode::const_iterator childIterator;
+
+    /// If this node has been fully processed yet or not.
+    bool processed;
+  };
+
+  /// Attempt to eliminate a redundant operation. Returns success if the
+  /// operation was marked for removal, failure otherwise.
+  LogicalResult simplifyOperation(ScopedMapTy &knownValues, Operation *op,
+                                  bool hasSSADominance);
+  void simplifyBlock(ScopedMapTy &knownValues, Block *bb, bool hasSSADominance);
+  void simplifyRegion(ScopedMapTy &knownValues, Region &region);
+  /// Return the number of erased operations.
+  unsigned simplify(Operation *rootOp);
+
+private:
+  /// Operations marked as dead and to be erased.
+  std::vector<Operation *> opsToErase;
+
+  /// The dominance info to use.
+  DominanceInfo *domInfo;
+//===----------------------------------------------------------------------===//
+// END copied from mlir/lib/Transforms/CSE.cpp
+//===----------------------------------------------------------------------===//
+  /// An optional listener to notify of replaced or erased operations.
+  RewriteListener *listener;
+//===----------------------------------------------------------------------===//
+// BEGIN copied from mlir/lib/Transforms/CSE.cpp
+//===----------------------------------------------------------------------===//
+};
+
+} // namespace
+
+/// Attempt to eliminate a redundant operation.
+LogicalResult CSE::simplifyOperation(ScopedMapTy &knownValues, Operation *op,
+                                     bool hasSSADominance) {
+  // Don't simplify terminator operations.
+  if (op->hasTrait<OpTrait::IsTerminator>())
+    return failure();
+
+  // If the operation is already trivially dead just add it to the erase list.
+  if (isOpTriviallyDead(op)) {
+    opsToErase.push_back(op);
+    return success();
+  }
+
+  // Don't simplify operations with nested blocks. We don't currently model
+  // equality comparisons correctly among other things. It is also unclear
+  // whether we would want to CSE such operations.
+  if (op->getNumRegions() != 0)
+    return failure();
+
+  // TODO: We currently only eliminate non side-effecting
+  // operations.
+  if (!MemoryEffectOpInterface::hasNoEffect(op))
+    return failure();
+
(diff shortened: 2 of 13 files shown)


include/Dialects/LinalgTransform/LinalgTransformOps.td
```diff
@@ -33,6 +34,34 @@ class Linalg_Transform_Operation<string name, list<OpTrait> props = []>
 
 //===----------------------------------------------------------------------===//
 
+def ScopeOp : Linalg_Transform_Operation<"scope",
```
nicolasvasilache, line 37: We were discussing doing exactly this but on FuncOps, with the extra benefit of being able to "name" things.
So I started slicing out functionality (e.g. https://reviews.llvm.org/D116591).

Looking forward, do you see value in:
1. keeping this op without a Symbol interface ?
2. having a separate op than just FuncOp ?

Mogball: This operation is meant to be short-lived. I.e.

```
ScopeOp scope = wrapInScope(myOp);
/// do a bunch of transformations in the scope
unwrapScope(scope);
```

In which case it would not need a symbol to look up later. If it needs to exist longer than a few transformations, then you could pass the handle around instead of passing the name around and looking it up.

lib/Dialects/LinalgTransform/Transforms/ScopedTransform.cpp
```diff
@@ -0,0 +1,128 @@
...
+  rewriter.setInsertionPoint(op);
+
+  auto scope = rewriter.create<linalg::transform::ScopeOp>(
+      op->getLoc(), op->getResultTypes(), op->getOperands());
+  Region &body = scope.body();
+  rewriter.setInsertionPointToStart(&body.emplaceBlock());
+  BlockAndValueMapping bv;
+  for (auto arg_operand :
+       llvm::zip(body.addArguments(op->getOperandTypes()), op->getOperands()))
+    bv.map(std::get<1>(arg_operand), std::get<0>(arg_operand));
```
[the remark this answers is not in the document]
Mogball: Right, thanks!

include/Dialects/LinalgTransform/FunctionalTransforms.h
```diff
@@ -0,0 +1,77 @@
...
+#define IREE_LLVM_SANDBOX_DIALECTS_LINALGTRANSFORM_FUNCTIONALTRANSFORMS_H
+
+#include "mlir/Dialect/Linalg/Transforms/Transforms.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/PatternMatch.h"
+
+namespace mlir {
+namespace linalg {
+
+class FunctionalPaddingPattern {
```
nicolasvasilache, line 19: I did some refactoring upstream and you should soon be able to just:
```
FailureOr<LinalgOp> applyPaddingPattern(
    LinalgOp op, PatternRewriter&rewriter, const LinalgPaddingOptions &options) {
  LinalgPaddingPattern pattern(op->getContext(), options);
  return pattern.returningMatchAndRewrite(op, rewriter);
}
```
everywhere here and below.

nicolasvasilache: The default empty filter will be passed and we wouldn't need to duplicate all those lines.
Do you have further extensions planned that wouldn't work with this?

include/Dialects/LinalgTransform/FunctionalTransforms.h
```diff
@@ -0,0 +1,77 @@
...
+      : options(options) {}
+
+  FailureOr<TiledLinalgOp> operator()(Operation *op,
+                                      PatternRewriter &rewriter) const;
+
+private:
+  LinalgTilingOptions options;
+};
+
+class FunctionalGenericTilingPattern : FunctionalBaseTilingPattern {
```
nicolasvasilache, line 45: There will soon be a single LinalgTilingPattern after https://reviews.llvm.org/D116733 lands.

lib/Dialects/LinalgTransform/Transforms/FunctionalTransforms.cpp
```diff
@@ -0,0 +1,222 @@
...
+         "requested peeling of non-existing loop");
+  TiledLoopOp result;
+  if (succeeded(peelAndCanonicalizeTiledLoop(rewriter, tiledLoop, idx, result)))
+    return result->getResults();
+  assert(!result && "expected that loop was not peeled");
+  return tiledLoop->getResults();
+}
+
+/// Peel loops after tiling.
+static void peelLoops(RewriterBase &rewriter, TiledLinalgOp &res,
```
nicolasvasilache, line 49: Exposing this function in core.
OTOH if you just wrap the unique LinalgTilingPattern we should be in a better place.

include/Transforms/Functional.h
```diff
@@ -0,0 +1,364 @@
...
+auto applyAt(OpT op, PatternT &&pattern, Args &&...args) {
+  detail::SimpleRewriter rewriter(op->getContext());
+  rewriter.setInsertionPoint(op);
+  return pattern(op, rewriter, std::forward<Args>(args)...);
+}
+
+/// Given a scope, apply a pattern until the first successful match and return
+/// the result.
+template <typename PatternT, typename... Args>
+auto applyOnce(Operation *scope, PatternT &&pattern, Args &&...args) {
```
nicolasvasilache, line 37: you could have the scopeop directly and simplify a bit

include/Transforms/Functional.h
```diff
@@ -0,0 +1,364 @@
...
+  assert(scope->hasTrait<OpTrait::IsIsolatedFromAbove>() &&
+         "scope is not isolated from above");
+  using Traits = llvm::function_traits<std::decay_t<PatternT>>;
+  using OpT = typename Traits::template arg_t<0>;
+
+  detail::SimpleRewriter rewriter(scope->getContext());
+  typename Traits::result_t result = failure();
+  scope->walk([pattern = std::forward<PatternT>(pattern), &result, &rewriter,
+               &args...](OpT op) {
+    rewriter.setInsertionPoint(op);
```
nicolasvasilache, line 47: changing insertion points without a previous OpBuilder::InsertionGuard is a red flag to me, can you make sure we RAII it properly here and everywhere?

include/Transforms/Functional.h
```diff
@@ -0,0 +1,364 @@
...
+}
+
+namespace detail {
+template <typename ReturnType>
+struct UnpackFailureOr {
+  using type = ReturnType;
+  static type unpack(type &&value) { return value; }
+};
+template <typename NestedType>
+struct UnpackFailureOr<FailureOr<NestedType>> {
```
nicolasvasilache, line 61: Do we need this rvalue dark-magic?
I.e. is it purely for optimization?
If so is it maybe premature?

include/Transforms/Functional.h
```diff
@@ -0,0 +1,364 @@
...
+    auto result = applyAt(op, std::forward<PatternT>(pattern),
+                          std::forward<Args>(args)...);
+    if (succeeded(result))
+      results.push_back(Unpack::unpack(std::move(result)));
+  }
+  return results;
+}
+
+/// Find the nearest proper ancestor that is isolated from above.
+Operation *findNearestScope(Operation *op);
```
nicolasvasilache, line 106: just getparentoftype since you went throught the trouble of creating an op?

include/Transforms/Functional.h
```diff
@@ -0,0 +1,364 @@
...
+    llvm::unique_function<LogicalResult(Operation *, PatternRewriter &)>;
+
+namespace impl {
+void applyPatternsRecursively(Operation *scope,
+                              MutableArrayRef<SequencePattern> patterns);
+} // end namespace impl
+
+namespace detail {
+template <typename OpT>
+struct IsaOr {
```
nicolasvasilache, line 195: I do not understand what this does :)

nicolasvasilache (commented):
Could you start splitting a bit?

The Listener + rewriter + CSE could go in first with some test taken from core that would call this directly?

Then once some of the functional stuff is rebased we should have a much more manageable diff?

Thanks for pushing on this!

Mogball:
Yeah I can split things up. Thanks for the review!

include/Transforms/Functional.h
Mogball: Let me document this in a little bit more detail...

include/Transforms/Functional.h
```diff
@@ -0,0 +1,364 @@
...
+      return WalkResult::advance();
+    result = op;
+    return WalkResult::interrupt();
+  });
+  return result;
+}
+} // end namespace detail
+
+template <typename SequenceT, typename... Args>
+auto applySequenceOnce(Operation *scope, SequenceT &&sequence, Args &&...args) {
```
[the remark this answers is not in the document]
Mogball: Yeah. This is not efficient.

include/Transforms/Functional.h
Mogball: These functions are meant to work in general, and not specifically with `ScopeOp` (although, yes, the naming is confusing...). I intend to keep the functionality separate as long as `ScopeOp` is tied to `linalg_transform` (it doesn't have to; I might make it more first-class).

include/Transforms/Functional.h
Mogball: It is customary to set the insertion point of the pattern rewriter to the root operation before calling into a pattern. An insertion guard is not needed because it would just reset the insertion point to null (and the rewriter is not used afterwards).

See:
https://github.com/llvm/llvm-project/blob/ab7618914dec6384f24b30ad2d45b82a51647a33/mlir/lib/Rewrite/PatternApplicator.cpp#L188

include/Transforms/Functional.h
Mogball: It's more for correctness. Many types aren't copy-constructable but are move-constructable. I.e., what happens if a pattern decides to return a `FailureOr<unique_ptr<SomeType>>`?

include/Transforms/Functional.h
Mogball: `ScopeOp` is specific to `linalg_transform`. These utilities work in general.

Mogball:
I've split out the changes so that this PR is just the listeners. I've added two tests from MLIR core for the canonicalizer and CSE passes with listeners.

include/Transforms/ListenerCSE.h
```diff
@@ -0,0 +1,23 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_IREE_SANDBOX_TRANSFORMS_LISTENERCSE_H
+#define LLVM_IREE_SANDBOX_TRANSFORMS_LISTENERCSE_H
+
+#include "Transforms/Listener.h"
+#include "mlir/IR/Operation.h"
```
ftynse, line 13: Nit: could we forward-declare Operation as well? We are only using it as a pointer in signature.

lib/Transforms/ListenerCSE.cpp
```diff
@@ -0,0 +1,279 @@
...
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/IR/Operation.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
+#include "llvm/ADT/ScopedHashTable.h"
+#include "llvm/Support/RecyclingAllocator.h"
+#include <deque>
+
+using namespace mlir;
+
+namespace {
```
ftynse, line 20: Could you please annotate which parts of this are taken from the "main" CSE verbatim and which parts are new? This will help keeping code in sync and eventually merging it.

lib/Transforms/ListenerGreedyPatternRewriteDriver.cpp
```diff
@@ -0,0 +1,391 @@
...
+#define DEBUG_TYPE "listener-greedy-rewriter"
+
+//===----------------------------------------------------------------------===//
+// GreedyPatternRewriteDriver
+//===----------------------------------------------------------------------===//
+
+namespace {
+/// This is a worklist-driven driver for the PatternMatcher, which repeatedly
+/// applies the locally optimal patterns in a roughly "bottom up" way.
+class GreedyPatternRewriteDriver : public RewriteListener {
```
ftynse, line 33: Ditto.

test/lib/Transforms/TestListenerPasses.cpp
```diff
@@ -0,0 +1,94 @@
...
+struct TestListener : public RewriteListener {
+  void notifyOperationReplaced(Operation *op, ValueRange newValues) override {
+    llvm::outs() << "REPLACED " << op->getName() << "\n";
+  }
+  void notifyOperationRemoved(Operation *op) override {
+    llvm::outs() << "REMOVED " << op->getName() << "\n";
+  }
+};
+
+struct TestCanonicalizePass : public PassWrapper<TestCanonicalizePass, Pass> {
```
ftynse, line 30: Nit: I wouldn't call it "test canonicalize" to (a) avoid confusion with the "main" canonicalize and (b) make it clear that it is testing the rewriter infra rather than canonicalization patterns.

lib/Transforms/ListenerCSE.cpp
```diff
@@ -0,0 +1,310 @@
...
+#include "mlir/IR/Operation.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
+#include "llvm/ADT/ScopedHashTable.h"
+#include "llvm/Support/RecyclingAllocator.h"
+#include <deque>
+
+using namespace mlir;
+
+//===----------------------------------------------------------------------===//
+// BEGIN copied from mlir/lib/Transforms/CSE.cpp
```
nicolasvasilache, line 21: Much better to help review, thanks!

test/Transforms/test-listener-canonicalize.mlir
```diff
@@ -0,0 +1,98 @@
+// RUN: mlir-proto-opt %s -allow-unregistered-dialect -test-listener-canonicalize --split-input-file | FileCheck %s
+
```
nicolasvasilache, line 2: Same thing here re copied from ?
Maybe just a top of the file comment that this is fully copied from core and to not modify?

test/Transforms/test-listener-cse.mlir
```diff
@@ -0,0 +1,246 @@
+// RUN: mlir-proto-opt -allow-unregistered-dialect %s -pass-pipeline='builtin.func(test-listener-cse)' | FileCheck %s
+
```
nicolasvasilache, line 2: Same thing here re copied from ?
Maybe just a top of the file comment that this is fully copied from core and to not modify?

nicolasvasilache (approved):
Very nice!

Please merge at your convenience.

PR #143, opened 2022-01-05
add argument parser

gysit:
- add an argument parsing function to the test harness
- update the matmul benchmark to test it

TODO: 
- update all remaining benchmarks in a follow up commit


python/examples/core/harness.py
```diff
@@ -228,6 +229,44 @@ def _run_benchmark_n_iters(callback: Callable[[int], None], n_iters: int,
   return np.asarray([_pytimed(callback, *args) for _ in range(n_iters)])
 
 
+def test_argparser(benchmark_name: str,
+                   problem_sizes_list: Sequence[Sequence[int]],
```
nicolasvasilache, line 233: can we rename all default args with default_ ?

nicolasvasilache (approved):
LGTM modulo some minor renamings.

python/examples/matmul/bench.py
```diff
@@ -144,13 +153,13 @@ def pytorch_kernel(args, sizes, types):
 
 def benchmark():
```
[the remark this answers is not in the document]
gysit: Yes we can remove it. Technically it has a slightly different purpose since the run_benchmarks script basically automatically collects all benchmarks that implement a benchmark method. However, I think this has not been used to much so I am fine removing it!

python/examples/core/harness.py
gysit: Yes, I thought the names get lengthy but since these are the default values it makes sense to prefix

PR #145, opened 2022-01-05
fix convolution argument parsing

gysit:
- add support to parse convolution problem sizes with nested stride and dilation arrays
- use the argument parsing in additional benchmarks

### DIFF
--- a/python/examples/core/harness.py
+++ b/python/examples/core/harness.py
@@ -1,4 +1,5 @@
 import argparse
+import re
 import sys
 import os
 import time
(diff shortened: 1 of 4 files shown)


python/examples/core/harness.py
```diff
@@ -229,6 +229,56 @@ def _run_benchmark_n_iters(callback: Callable[[int], None], n_iters: int,
   return np.asarray([_pytimed(callback, *args) for _ in range(n_iters)])
 
 
+def _parse_problem_sizes(argument: str) -> Sequence[Union[int, Sequence[int]]]:
```
nicolasvasilache, line 232: Seems like this could be implemented much more simply with comprehensions?

Some cleaned/compressed version of:
```
cur, next = [elem for elem in c.split(',')], []

# At 
changed = True
while changed:
  changed = False

  for c in cur:
    if type(c) == str and c.startswith('['):
      next.append([elem for elem in c[1:-1].split(',')])
      changed = True
    else:
      next.append(int(c))

  cur = next

return cur
```

python/examples/core/harness.py
```diff
@@ -229,6 +229,56 @@ def _run_benchmark_n_iters(callback: Callable[[int], None], n_iters: int,
...
+  """Parse a sequence of read only dimensions.
+
+  Examples:
+  k,m -> ['k', 'm']
+  [] -> []
+  """
+  if argument == '[]':
+    return []
+  read_only = []
+  for token in argument.split(','):
```
nicolasvasilache, line 275: same here: 
```
read_only = argument.split(')
for x in read_only:
  assert x.isalpha()
```

Also, is it "read_only" or "compile_time_symbols" ?
Can we use a better name?

python/examples/core/harness.py
gysit: yep will try to simplify this as discussed offline!

nicolasvasilache (approved):

PR #147, opened 2022-01-07
Add warp_single_lane op to VectorExt dialect

ThomasRaoux:
Add a new op that will be useful to do vector programming targeting GPU
warps.

### DIFF
--- a/include/Dialects/VectorExt/VectorExtOps.td
+++ b/include/Dialects/VectorExt/VectorExtOps.td
@@ -59,4 +59,72 @@ def VectorExt_YieldOp : VectorExt_Op<"yield", [
   let assemblyFormat = "attr-dict ($operands^ `:` type($operands))?";
 }
 
+def VectorExt_WarpSingleLaneOp : VectorExt_Op<"warp_single_lane",
+      [DeclareOpInterfaceMethods<RegionBranchOpInterface>,
+       SingleBlockImplicitTerminator<"vector_ext::YieldOp">,
+       RecursiveSideEffects]> {
+  let summary = "vector.warp_single_lane operation";
+  let description = [{
+    Operation that encapsulate code that should be ran on a single lane within a
+    warp. Any code present in the region would only be executed on first lane
+    based on the laneid paramter. The op assumes `laneid` operand is an id to
+    identify the warp and it goes from 0 to warp_size.
+
+    // TODO: Add warp size as an attribute and support multiple dimensions lane
+    ids.
+
+    Operands are vector values distributed on all threads that may be used by
+    the single lane execution. The matching region argument is a vector of all
+    the values of those threads available to the single active lane. The
+    distributed dimension is implicit based on the shape of the operand and
+    argument. In the future this may be described by an affine map.
+
+    Return values are distributed on all threads using laneId as index. The
+    vector is distributed based on the shape ratio between the vector type of
+    the yield and the result type.
+    If the shapes are the same this means the value is broadcasted to all lanes.
+    In the future the distribution can be made more explicit using affine_maps
+    and will support having multiple Ids.
+
+    During lowering values passed as operands and return value need to be
+    visible to different lanes within the warp. This would usually be done by
+    going through memory.
+
+    Example:
+    ```
+    vector_ext.warp_single_lane (%laneid) {
+      ...
+    }
+
+    %0 = vector_ext.warp_single_lane(%laneid)
+    args(%v0 : vector<4xi32>) -> (vector<1xf32>) {
+    ^bb0(%arg0 : vector<128xi32>) :
+      ...
+      vector_ext.yield %1 : vector<32xf32>
+    }
+    ```
+
+    This may be lowered to an scf.if region as below:
+    ```
+      %cnd = arith.cmpi eq, %laneid, %c0 : index
+      scf.if %cnd {
+         ...
+      }
+    ```
+  }];
+
+  // TODO: Add verifier.
+
+  let arguments = (ins Index:$laneid, Variadic<AnyType>:$args);
+  let results = (outs Variadic<AnyType>:$results);
+  let regions = (region SizedRegion<1>:$warpRegion);
+
+  let skipDefaultBuilders = 1;
+  let builders = [
+    OpBuilder<(ins "Value":$laneid)>,
+    // TODO: Add arguments support in the builder.
+    OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid)>
+  ];
+}
+
 #endif // DIALECTS_VECTOREXT_VECTOREXTOPS
--- a/include/Dialects/VectorExt/VectorExtOps.td
+++ b/include/Dialects/VectorExt/VectorExtOps.td
@@ -139,7 +139,7 @@ def VectorExt_WarpSingleLaneOp : VectorExt_Op<"warp_execute_on_lane_0",
   }];
 
   // TODO: Add verifier to check type consistentcy at the region boundary based
-  on warp size and future permutation map.
+  // on warp size and future permutation map.
 
   let arguments = (ins Index:$laneid, Variadic<AnyType>:$args);
   let results = (outs Variadic<AnyType>:$results);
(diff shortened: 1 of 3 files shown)


include/Dialects/VectorExt/VectorExtOps.td
```diff
@@ -59,4 +59,72 @@ def VectorExt_YieldOp : VectorExt_Op<"yield", [
...
+       SingleBlockImplicitTerminator<"vector_ext::YieldOp">,
+       RecursiveSideEffects]> {
+  let summary = "vector.warp_single_lane operation";
+  let description = [{
+    Operation that encapsulate code that should be ran on a single lane within a
+    warp. Any code present in the region would only be executed on first lane
+    based on the laneid paramter. The op assumes `laneid` operand is an id to
+    identify the warp and it goes from 0 to warp_size.
+
+    // TODO: Add warp size as an attribute and support multiple dimensions lane
```
nicolasvasilache, line 73: Also to implement the verifier?
I.e. attribute + warps sizes (+ in the future permutation) + vector type -> type of distributed vector ?

include/Dialects/VectorExt/VectorExtOps.td
```diff
@@ -59,4 +59,72 @@ def VectorExt_YieldOp : VectorExt_Op<"yield", [
...
+    going through memory.
+
+    Example:
+    ```
+    vector_ext.warp_single_lane (%laneid) {
+      ...
+    }
+
+    %0 = vector_ext.warp_single_lane(%laneid)
+    args(%v0 : vector<4xi32>) -> (vector<1xf32>) {
```
nicolasvasilache, line 100: I am a little confused now as the stuff we had discussed offline seemed to:
1. capture from above (i.e. not be isolated)
2. take all other data from buffers

What you propose here makes sense too but seems to be a different design (which makes sense as things are evolving).
I would formalize the thinking a little more in the doc and explain how the distributed vectors and the sequential vectors relate. A slightly larger example connecting to function boundaries would help too.

I suggest something like:

```
// future: attribute or extra operand to describe #lanes
func @distributed_running_on_all_lanes(%laneid: index, %v: vector<4xi32>) -> vector<1xi32>{
  // Execute in parallel on all lanes.
  ...
  some_synchronization_primitive(...)
  // Alternatively, allow: vector_ext.warp_single_lane [%lane_id = 0] ... 
  %0 = vector_ext.warp_single_lane when %lane_id is %c0 args(%v0 : vector<4xi32>) -> (vector<1xf32>) {

      // 32 * vector<4xi32> are packed into a vector<128xi32> following the natural order 
      // This is the simplest form that makes sense.
      // This may have to go through memory (i.e. explicit buffers) when lowering to executable IR.
      // TODO: more advanced mechanisms than simple natural order packing. 
      ^bb0(%arg0 : vector<128xi32>) :
        // Execute sequentially only on lane 0.
        ...

        // vector<32xi32> unpack into 32 * vector<1xi32> following the natural order 
        // This is the simplest form that makes sense.
        // This may have to go through memory (i.e. explicit buffers) when lowering to executable IR.
        // TODO: more advanced mechanisms than simple natural order unpacking. 
        vector_ext.yield %1 : vector<32xf32>
    }

    some_synchronization_primitive(...)

    // Execute in parallel on all lanes.
}
```

The more advanced mechanisms for stacking / decomposing start to feel a bit like the indexing_maps of a linalg.generic on vectors but with some variations that we need to discover.
The packing/unpacking ops seem close to the vector.insert/extract_map which at this point could be renamed into distributed_packing / distributed_unpacking ?

Does the intuition of the above description make sense ?

include/Dialects/VectorExt/VectorExtOps.td
```diff
@@ -59,4 +59,72 @@ def VectorExt_YieldOp : VectorExt_Op<"yield", [
...
+    ^bb0(%arg0 : vector<128xi32>) :
+      ...
+      vector_ext.yield %1 : vector<32xf32>
+    }
+    ```
+
+    This may be lowered to an scf.if region as below:
+    ```
+      %cnd = arith.cmpi eq, %laneid, %c0 : index
+      scf.if %cnd {
```
nicolasvasilache, line 110: Now that I wrote the more fleshed out example above, this seems to need reconciliation: atm you omit the body but 
I would again spell it out to ensure all makes sense e2e, and clearly specifiy what surrounding code runs in parallel and what is sequential.

```
// future: attribute or extra operand to describe #lanes
func @distributed_running_on_all_lanes(%laneid: index, %v: vector<4xi32>) -> vector<1xi32>{
  // Execute in parallel on all lanes.
  ...
  some_synchronization_primitive(...)

  // pack 32 * vector<4xi32> into a single vector<128xi32> along the natural dimension
  // this will likely go through memory and involve extra synchronizations
  %sv0 = vector.distributed_packing / vector.insert_map (%v0): vector<4xi32> into  vector<128xi32>
  %cnd = arith.cmpi eq, %laneid, %c0 : index
  %sv1 = scf.if %cnd {
      // Execute sequentially only on lane 0.
      // %arg0 : vector<128xi32> is RAUW by %sv0
      ...
      scf.yield %1 : vector<32xf32>
    }
    // unpack vector<32xi32> into 32 * vector<1xi32> along the natural dimension
    // this will likely go through memory and involve extra synchronizations
    %v1 = vector.distributed_unpacking / vector.extract_map (%v0, %laneid): vector<32xi32> into  vector<1xi32>
    some_synchronization_primitive(...)

    // Execute in parallel on all lanes.
}
```

include/Dialects/VectorExt/VectorExtOps.td
```diff
@@ -59,4 +59,72 @@ def VectorExt_YieldOp : VectorExt_Op<"yield", [
...
+    visible to different lanes within the warp. This would usually be done by
+    going through memory.
+
+    Example:
+    ```
+    vector_ext.warp_single_lane (%laneid) {
+      ...
+    }
+
+    %0 = vector_ext.warp_single_lane(%laneid)
```
nicolasvasilache, line 99: Ok now that I wrote the e2e mapping and understand things better, I would claim you either want this op to be called `vector_ext.warp_execute_on_lane_0` or `vector_ext.warp_execute_on_single_lane when %laneid is %some_constant`.

nicolasvasilache (commented):
Ok, made a pass on the design and I think this is promising.
I made suggestions on improving the doc to convey a little more fleshed out e2e path.


Digging a bit deeper, I think this is still quite close to what we used to talk about in the past with the following differences / design improvements:
1. vector.transfer would not be in tensor land but rather in buffer land. Phase ordering means it would happen after vectorization and after bufferization.
2. your new op seems to remove all the weirdness of a "fake distributed loop" by introducing sequential execution, which makes a lot of sense.
3. we will need some explicit packing/unpacking ops that actually seem very close to the vector.insert/extract_map
4. these packing/unpacking ops must not fold away anymore but lower through memory.


Does the above mental model make sense to you?

ThomasRaoux:
> Ok, made a pass on the design and I think this is promising. I made suggestions on improving the doc to convey a little more fleshed out e2e path.
> 
> Digging a bit deeper, I think this is still quite close to what we used to talk about in the past with the following differences / design improvements:
> 
> 1. vector.transfer would not be in tensor land but rather in buffer land. Phase ordering means it would happen after vectorization and after bufferization.

Yes

> 2. your new op seems to remove all the weirdness of a "fake distributed loop" by introducing sequential execution, which makes a lot of sense.

Yes

> 3. we will need some explicit packing/unpacking ops that actually seem very close to the vector.insert/extract_map
> 4. these packing/unpacking ops must not fold away anymore but lower through memory.
> 
I don't think we need the explicit packing/unpacking. My idea was to have it being done at the region boundary. The behavior would still be exactly like vector.insert/extract_map. I added some example of things would be lowered to memory in the op description. Let me know if i makes sense to you. I don't a use for extra packing/unpacking ops at the time.


> Does the above mental model make sense to you?

Yes, thanks for summarizing! Overall what you described matches what I had in mind. I haven't introduced the affine map to try to keep things simple at the beginning but as you pointed out it will help making the distribution more explicit.

nicolasvasilache (approved):

PR #148, opened 2022-01-07
Add plot method to the test result class.

gysit:
- Add a plot method to plot configurable performance numbers across problem sizes.

By default no plotting happens. If a 'plot_path' kwarg is passed to the test_harness, plots will be generated in the given path.

### DIFF
--- a/python/examples/core/harness.py
+++ b/python/examples/core/harness.py
@@ -9,6 +9,7 @@ from typing import AbstractSet, Any, Callable, List, Mapping, Optional, Sequence
 
 import numpy
 import pandas
+import seaborn
 
 from mlir.execution_engine import *
 from mlir.ir import *
@@ -38,7 +39,7 @@ class Measurements:
   config_keys = [
     "expert",
     "np_types",
-    "compile_time_problem_sizes_dict",
+    "runtime_only",
     "runtime_problem_sizes_dict"
   ]
   data_keys = [
@@ -52,15 +53,16 @@ class Measurements:
         dict([(col, []) for col in self.config_keys + self.data_keys]))
 
   def append(self, expert: str, np_types: Sequence[np.dtype],
-             compile_time_problem_sizes_dict: ProblemSizes,
+             runtime_only_sizes: AbstractSet[str],
              runtime_problem_sizes_dict: ProblemSizes,
              timing_results_dict: TimingResults):
     """Append measurement results."""
     config = pandas.DataFrame(dict(
       zip(self.config_keys,
-          [[expert], [np_types],
-            [compile_time_problem_sizes_dict],
-            [runtime_problem_sizes_dict]])))
+          [[expert],
+            [self._stringify_types(np_types)],
+              [self._stringify_set(runtime_only_sizes)],
+              [self._stringify_dict(runtime_problem_sizes_dict)]])))
     results = pandas.DataFrame(dict(
       [(k, timing_results_dict[k]) for k in self.data_keys]))
     product = config.merge(results, how='cross')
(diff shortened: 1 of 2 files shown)


python/examples/core/harness.py
```diff
@@ -38,7 +39,7 @@ class Measurements:
   config_keys = [
     "expert",
     "np_types",
-    "compile_time_problem_sizes_dict",
+    "runtime_only",
```
nicolasvasilache, line 42: I'd love to get rid of `runtime_only` and replace by `dynamic_at_compile_time` or something similarly descriptive.

python/examples/core/harness.py
```diff
@@ -465,8 +541,12 @@ def test_harness(problem_factory: Callable[
                 'pytorch_benchmark'], n, args, problem_sizes_dict, np_types),
             gflops, gbytes, n_iters)
 
-        measurements.append('pytorch', np_types,
-                            compile_time_problem_sizes_dict,
+        measurements.append('pytorch', np_types, runtime_only_sizes,
```
nicolasvasilache, line 544: that nesting of numpy and pytorch within the execution was really to get started and likely needs to be torn out and refactored (not in this PR).

python/examples/core/harness.py
gysit: I would suggest to do that in a followup. We also wanted to name the experts. All these changes should probably be separate commits since they touch all the test_harness calls?

python/examples/core/harness.py
```diff
@@ -74,6 +76,81 @@ def to_data_frame(self) -> pandas.DataFrame:
     """Return a data frame containing the aggregated data."""
     return self.data
 
+  def plot(self, path: os.path, data_key: str, data_label: str):
+    """Plot the measurements."""
+    config_key = "runtime_problem_sizes_dict"
+    keys_to_iter = [k for k in self.config_keys if k is not config_key]
```
nicolasvasilache, line 82: significantly better variable naming would greatly increase the understanding of what is happening here

python/examples/core/harness.py
```diff
@@ -74,6 +76,81 @@ def to_data_frame(self) -> pandas.DataFrame:
...
+    config_to_iter = self.data[keys_to_iter].drop_duplicates()
+    for _, row in config_to_iter.iterrows():
+      # Compute the plot file name by filtering non alpha numberic characters.
```
nicolasvasilache, line 85: seems like this should be extracted out in a separate helper func

python/examples/core/harness.py
```diff
@@ -74,6 +76,81 @@ def to_data_frame(self) -> pandas.DataFrame:
...
+      file_name = "plot"
+      for k, v in row.to_dict().items():
+        alnum_name = ''.join([c for c in v if c.isalnum()])
+        file_name += str.format(f"_{k}_{alnum_name}")
+      file_name += ".pdf"
+      file_name = os.path.join(path, file_name)
+      # Obtain the data elements to plot.
```
nicolasvasilache, line 92: same thing: plz factor out and give better names

python/examples/core/harness.py
```diff
@@ -74,6 +76,81 @@ def to_data_frame(self) -> pandas.DataFrame:
...
+      filtered = self.data
+      for k, v in row.to_dict().items():
+        filtered = filtered[filtered[k] == v]
+      # Plot the selected data.
```
nicolasvasilache, line 96: same

python/examples/core/harness.py
```diff
@@ -74,6 +76,81 @@ def to_data_frame(self) -> pandas.DataFrame:
...
+      fig = plt.get_figure()
+      fig.tight_layout()
+      fig.savefig(file_name)
+
+  def _compress_problem_sizes_label(self,
+              labels: Sequence[str]) -> (Sequence[str], Sequence[str]):
+    """Compress the problem size labels.
+
+    Identify the dimensions that take different values and compress the
+    labels to contain only the values for these varying dimensions.
```
nicolasvasilache, line 113: would be nice to add a line to explain why this is useful

nicolasvasilache (commented):
Would be nice to also put some resulting graph in a gist somewhere and paste the link in the commit message somewhere.

python/examples/core/harness.py
gysit: done

python/examples/core/harness.py
gysit: done

python/examples/core/harness.py
gysit: done

python/examples/core/harness.py
gysit: done

python/examples/core/harness.py
nicolasvasilache: SG

nicolasvasilache (approved):
Nice, thanks much!

PR #154, opened 2022-01-07
Add possibility to name experts and update benchmarks and tests.

gysit:
Add a test_sizes and a test_expert method to pass in problem sizes and experts. The test_sizes method replaces the make_problem_sizes methods found in all tests and benchmarks. The test_expert method names the experts and returns an expert dictionary. The names can be set manually which is done for all the benchmarks that can be controlled using arguments.

Additionally, the revision updates all tests and benchmarks to use test_sizes and test_experts and extends the padded convolution experiment with argument parsing.


ftynse (approved):

gysit:
> Have you considered adding a name to TransformationList rather than always tying it dynamically in the harness?

I thought it makes more sense for the limited use cases - benchmarks - to name them separately. Additionally, we may want to name the TransformationList (eg the SingleTilingExpert) differently depending on benchmark and parameters. That means we probably would need to add the names similarly to the variables?

PR #155, opened 2022-01-08
Re-implement Tracking CSE and Greedy Rewriter using listeners

Mogball:
Depends on https://github.com/google/iree-llvm-sandbox/pull/139

The tracking CSE and greedy rewriter are reimplemented with a listener.


lib/Dialects/LinalgTransform/Transforms/TrackingCSE.cpp
```diff
@@ -7,296 +7,16 @@
...
-void TrackingCSE::simplify(Operation *root) {
-  for (Region &region : root->getRegions()) {
-    ScopedMapTy knownValues;
-    simplifyRegion(knownValues, region);
-  }
-
-  for (Operation *op : opsToErase)
-    op->erase();
-}
+using namespace mlir::linalg;
```
ftynse, line 15: Why is this needed?

include/Dialects/LinalgTransform/TrackingCSE.h
```diff
@@ -9,13 +9,11 @@
 #ifndef MLIR_DIALECT_LINALG_TRANSFORMS_TRACKINGCSE_H
 #define MLIR_DIALECT_LINALG_TRANSFORMS_TRACKINGCSE_H
 
+#include "mlir/Dialect/Linalg/IR/LinalgInterfaces.h"
```
ftynse, line 12: I wouldn't want this dependency unless strictly necessary.

ftynse (commented):
LGTM

include/Dialects/LinalgTransform/TrackingCSE.h
Mogball: `SmallVector<LinalgOp>` requires the type to be complete, so unfortunately a forward declaration won't work.

PR #156, opened 2022-01-08
Introduce a ScopeOp

Mogball:
Depends on https://github.com/google/iree-llvm-sandbox/pull/155

`ScopeOp` is an isolated from above operation that can be used to perform finer-grain application of transformations. Adds two utility functions to wrap a single operation into a scope and then unwrap a scope of many operations.

### DIFF
--- a/include/Dialects/LinalgTransform/LinalgTransformOps.td
+++ b/include/Dialects/LinalgTransform/LinalgTransformOps.td
@@ -17,6 +17,7 @@
 include "mlir/IR/OpBase.td"
 include "mlir/IR/OpAsmInterface.td"
 include "mlir/Dialect/PDL/IR/PDLTypes.td"
+include "mlir/Interfaces/ControlFlowInterfaces.td"
 
 def Linalg_Transform_Dialect : Dialect {
   let name = "linalg_transform";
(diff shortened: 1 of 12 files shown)


include/Dialects/LinalgTransform/LinalgTransformOps.td
```diff
@@ -33,6 +34,34 @@ class Linalg_Transform_Operation<string name, list<OpTrait> props = []>
 
 //===----------------------------------------------------------------------===//
 
+def ScopeOp : Linalg_Transform_Operation<"scope",
```
ftynse, line 37: Bikeshed: given that this operation is a utility and we don't expect to see it in the transform description, I would reflect that in the name. Something like `util.scope` would be better.

include/Dialects/LinalgTransform/LinalgTransformOps.td
```diff
@@ -33,6 +34,34 @@ class Linalg_Transform_Operation<string name, list<OpTrait> props = []>
...
+  let description = [{An operation to restrict transformation scopes.}];
+
+  let regions = (region AnyRegion:$body);
+  let arguments = (ins Variadic<AnyType>:$ins);
+  let results = (outs Variadic<AnyType>:$outs);
+  let assemblyFormat = "`(` operands `)` attr-dict-with-keyword $body `:` functional-type(operands, results)";
+
+  let extraClassDeclaration = [{
+    static StringRef getDefaultDialect() { return "linalg_transform"; }
+  }];
```
ftynse, line 49: I'm not sure this is relevant since the operation will be wrapping the IR being transformed rather than the IR that describes the transformation.

include/Dialects/LinalgTransform/ScopedTransform.h
```diff
@@ -0,0 +1,33 @@
+//===-- ScopedTransform.h -------------------------------------------------===//
```
ftynse, line 1: Nit: there's usually a tiny description in this line.

lib/Dialects/LinalgTransform/Transforms/ScopedTransform.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  auto scope = rewriter.create<linalg::transform::ScopeOp>(
+      op->getLoc(), op->getResultTypes(), op->getOperands());
+  Region &body = scope.body();
+  rewriter.setInsertionPointToStart(&body.emplaceBlock());
+  BlockAndValueMapping bv;
+  bv.map(op->getOperands(), body.addArguments(op->getOperandTypes()));
+
+  Operation *cloneInScope = op->clone(bv);
+  rewriter.insert(cloneInScope);
```
ftynse, line 32: `rewriter.clone`?

lib/Dialects/LinalgTransform/Transforms/ScopedTransform.cpp
```diff
@@ -0,0 +1,84 @@
...
+  rewriter.setInsertionPointToEnd(&body.front());
```
ftynse, line 33: Strictly speaking, this is unnecessary. The insertion point will be pointing there anyway.

include/Dialects/LinalgTransform/LinalgTransformOps.td
Mogball: The sole purpose of this is so that the `linalg_transform.forward` required terminator is just `forward`.

include/Dialects/LinalgTransform/LinalgTransformOps.td
Mogball: Sure. Once this feature gets more use, I'm planning to move it into its own dialect (with only `scope` and `forward` as the operations).

nicolasvasilache (approved):
LG!

PR #157, opened 2022-01-08
A light functional rewrite API in C++

Mogball:
Depends on https://github.com/google/iree-llvm-sandbox/pull/156

Introduce a lightweight functional rewrite API in C++ and some utility functions that can be used to build a fully-featured functional rewrite system moving forward (e.g. https://github.com/google/iree-llvm-sandbox/issues/149).

Simplifies the linalg transform interpreter, as operation handles can be passed around directly instead of relying on attributes.

### DIFF
--- a/include/Transforms/Functional.h
+++ b/include/Transforms/Functional.h
@@ -1,0 +1,348 @@
+//===- Functional.h ---------------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef MLIR_REWRITE_FUNCTIONAL_H
+#define MLIR_REWRITE_FUNCTIONAL_H
+
+#include "mlir/IR/PatternMatch.h"
+#include "llvm/ADT/STLExtras.h"
+
+namespace mlir {
+namespace functional {
+
+/// A "Functional Pattern" is a callable concept that accepts as its first
+/// argument an operation or operation interface and as its second argument a
+/// `RewriterBase` or `PatternRewriter`. Beyond these, it can accept additional
+/// parameters of any type.
+///
+/// A functional pattern is expected to return a type convertible to
+/// `LogicalResult`. If the result is a `FailureOr<T>`, then `T` is forwarded to
+/// subsequent patterns in sequences. Additionally, if `T` is a pair or tuple,
+/// its elements are unpacked and passed as separate arguments to subsequent
+/// patterns.
+template <typename PatternT>
+struct PatternConcept {
+  static constexpr bool verify() {
+    using Traits = llvm::function_traits<std::decay_t<PatternT>>;
+    static_assert(Traits::num_args >= 2,
+                  "Patterns must have at least two arguments.");
+
+    using OpT = typename Traits::template arg_t<0>;
+    static_assert(std::is_convertible<OpT, Operation *>::value,
+                  "The first argument of a pattern must be Operation * or "
+                  "convertible to Operation *");
+
+    using RewriterT = typename Traits::template arg_t<1>;
+    static_assert(std::is_convertible<PatternRewriter &, RewriterT>::value,
+                  "The second argument of a pattern must be convertible from "
+                  "PatternRewriter & (e.g. RewriterBase &)");
+
+    using ResultT = typename Traits::result_t;
+    static_assert(
+        std::is_convertible<ResultT, LogicalResult>::value,
+        "The result of a pattern must be convertible to LogicalResult");
+
+    return true;
+  }
+};
+
+namespace detail {
+/// A simple pattern rewriter that implements no special logic.
+class SimpleRewriter : public PatternRewriter {
+public:
+  SimpleRewriter(MLIRContext *context) : PatternRewriter(context) {}
+};
+} // end namespace detail
+
+/// Apply a pattern directly on an operation. This function instantiates a
+/// simple pattern rewriter and calls the pattern directly on the operation with
+/// the given arguments.
+template <typename OpT, typename PatternT, typename... Args,
+          bool = PatternConcept<PatternT>::verify()>
+auto applyAt(OpT op, PatternT &&pattern, Args &&...args) {
+  detail::SimpleRewriter rewriter(op->getContext());
+  rewriter.setInsertionPoint(op);
+  return pattern(op, rewriter, std::forward<Args>(args)...);
+}
+
+/// Given a scope, apply a pattern with the given arguments until the first
+/// successful match and return the result. This function instantiates a simple
+/// pattern rewriter.
+template <typename PatternT, typename... Args,
+          bool = PatternConcept<PatternT>::verify()>
+auto applyOnceIn(Operation *scope, PatternT &&pattern, Args &&...args) {
+  assert(scope->hasTrait<OpTrait::IsIsolatedFromAbove>() &&
+         "scope is not isolated from above");
+  using Traits = llvm::function_traits<std::decay_t<PatternT>>;
+  using OpT = typename Traits::template arg_t<0>;
+
+  detail::SimpleRewriter rewriter(scope->getContext());
+  typename Traits::result_t result = failure();
+  scope->walk([pattern = std::forward<PatternT>(pattern), &result, &rewriter,
+               &args...](OpT op) {
+    rewriter.setInsertionPoint(op);
+    result = pattern(op, rewriter, std::forward<Args>(args)...);
+    return failed(result) ? WalkResult::advance() : WalkResult::interrupt();
+  });
+  return result;
+}
+
+namespace detail {
+/// If a pattern returns `FailureOr<Type>`, unpack the nested value of `Type`.
+/// Otherwise, just return the whole value.
+template <typename ReturnType>
+struct UnpackFailureOr {
+  using type = ReturnType;
+  /// Base case. If a pattern does not return `FailureOr`, just forward the
+  /// whole result. Usually, this is a `LogicalResult`.
+  static type unpack(type &&value) { return value; }
+};
+template <typename NestedType>
+struct UnpackFailureOr<FailureOr<NestedType>> {
+  using type = NestedType;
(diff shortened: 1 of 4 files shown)


include/Transforms/Functional.h
```diff
@@ -0,0 +1,442 @@
...
+using SequencePattern =
+    llvm::unique_function<LogicalResult(Operation *, PatternRewriter &)>;
+/// Greedily and recursively apply a set of functional patterns.
+void applyPatternsRecursively(Operation *scope,
+                              MutableArrayRef<SequencePattern> patterns);
+} // end namespace impl
+
+namespace detail {
+/// Utility struct for handling functional patterns that may operate on generic
+/// `Operation *` or a more specific interface or op type. In the base base,
```
ftynse, line 216: Typo: "in the base base"

include/Transforms/Functional.h
```diff
@@ -0,0 +1,487 @@
...
+
+    using RewriterT = typename Traits::template arg_t<1>;
+    static_assert(std::is_convertible<PatternRewriter &, RewriterT>::value,
+                  "The second argument of a pattern must be convertible from "
+                  "PatternRewriter & (e.g. RewriterBase &)");
+
+    using ResultT = typename Traits::result_t;
+    static_assert(
+        std::is_convertible<ResultT, LogicalResult>::value,
+        "The result of a pattern must be convertible ot LogicalResult");
```
nicolasvasilache, line 48: nit `to`

include/Transforms/Functional.h
```diff
@@ -0,0 +1,487 @@
...
+  static type unpack(FailureOr<type> &&value) { return std::move(*value); }
+};
+} // end namespace detail
+
+/// Given a scope, apply a pattern once per operation in the scope, saving the
+/// result of each match. The result list is empty if the pattern failed to
+/// match at all.
+template <typename PatternT, typename... Args,
+          bool = PatternConcept<PatternT>::verify()>
+auto applyEach(Operation *scope, PatternT &&pattern, Args &&...args) {
```
nicolasvasilache, line 119: `auto applyToEachInScope(ScopeOp scope, ... )` ?

include/Transforms/Functional.h
```diff
@@ -0,0 +1,487 @@
...
+    if (succeeded(result))
+      results.push_back(Unpack::unpack(std::move(result)));
+  });
+  return results;
+}
+
+/// Apply a pattern directly on an operation, for each operation in a list.
+template <typename ListT, typename PatternT, typename... Args,
+          bool = PatternConcept<PatternT>::verify()>
+auto applyForEach(ListT &&list, PatternT &&pattern, Args &&...args) {
```
nicolasvasilache, line 143: I get the "for_each" analogy but I think I'd prefer terminology consistency with the other version:
aaplyToEachInScoped / applyToEachInList ?

include/Transforms/Functional.h
```diff
@@ -0,0 +1,487 @@
...
+  }
+  return results;
+}
+
+/// Find the nearest proper ancestor that is isolated from above.
+Operation *findNearestScope(Operation *op);
+
+/// Clone the regions of an operation.
+SmallVector<std::unique_ptr<Region>> cloneRegionsOf(Operation *op);
+/// Replace the regions of an operation with the given regions.
```
nicolasvasilache, line 164: nit newline

include/Transforms/Functional.h
```diff
@@ -0,0 +1,487 @@
...
+
+/// Apply a pattern sequence directly on an operation. If any pattern in the
+/// sequence fails, then the state of the IR is restored. Patterns in the
+/// sequence should not query or modify IR outside the nearest scope.
+/// TODO: The rollback is not efficient. All of the IR underneath the nearest
+/// scope is copied.
+template <typename OpT, typename SequenceT, typename... Args,
+          bool = PatternConcept<SequenceT>::verify()>
+auto applySequenceAt(OpT op, SequenceT &&sequence, Args &&...args) {
+  Operation *scope = findNearestScope(op);
```
nicolasvasilache, line 193: Same comment re ScopeOp vs Operation* and future with interfaces.
But I won't insist anymore if you fell more strongly here :)

include/Transforms/Functional.h
```diff
@@ -0,0 +1,487 @@
...
+namespace detail {
+template <typename OpT, typename SequenceT, typename... Args,
+          bool = PatternConcept<SequenceT>::verify()>
+auto applySequence(PatternRewriter &rewriter, Operation *scope, OpT op,
+                   SequenceT &&sequence, Args &&...args) {
+  // Save a copy of the scope.
+  SmallVector<std::unique_ptr<Region>> backup = cloneRegionsOf(scope);
+
+  rewriter.setInsertionPoint(op);
+  auto result = std::forward<SequenceT>(sequence)(op, rewriter,
```
nicolasvasilache, line 176: Could you add a comment here to say what this is forwarded to? I am not clear which of the other functions can get called as the result of that.

I am asking because this relates to your generic findNearestScope but I am wondering whether some paths may bottom out on the wrap/unwrap whose impl. is very specific to ScopeOp.
Outlining generic isolated form above ops can get quite painful (i.e. finding and reconciling terminators)

lib/Transforms/Functional.cpp
```diff
@@ -0,0 +1,153 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#include "Transforms/Functional.h"
+#include "mlir/IR/BlockAndValueMapping.h"
+
+using namespace mlir;
+using functional::impl::SequencePattern;
+
+Operation *mlir::functional::findNearestScope(Operation *op) {
```
nicolasvasilache, line 15: Ah sorry, I get it now that I see the impl.
Do you already see your mechanisms apply to generic isolated form above?
I am also wondering how this relates to the ScopeOp if at all.

lib/Transforms/Functional.cpp
```diff
@@ -0,0 +1,153 @@
...
+  } while (op && !op->hasTrait<OpTrait::IsIsolatedFromAbove>());
+  return op;
+}
+
+SmallVector<std::unique_ptr<Region>>
+mlir::functional::cloneRegionsOf(Operation *op) {
+  SmallVector<std::unique_ptr<Region>> regions;
+  BlockAndValueMapping mapping;
+  for (Region &region : op->getRegions()) {
+    region.cloneInto(&*regions.emplace_back(std::make_unique<Region>()),
```
nicolasvasilache, line 27: nit: maybe if you s/region/r and s/mapping/bvm you can save 1 line + braces (i.e. 2 lines total)?

lib/Transforms/Functional.cpp
```diff
@@ -0,0 +1,153 @@
...
+  assert(op->getNumRegions() == regions.size() && "unequal number of regions");
+  for (auto it : llvm::zip(op->getRegions(), regions)) {
+    Region &region = std::get<0>(it);
+    region.getBlocks().clear();
+    region.takeBody(*std::get<1>(it));
+  }
+}
+
+namespace {
+class RecursiveMatchDriver : public PatternRewriter {
```
nicolasvasilache, line 44: Unless I missing something, it seems that everything below here is not exercised in the revision (or if it is is is hard for me to find)?

Could we split the recursive concepts it in a separate revision where we more easily see where they get applied ?

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -231,58 +197,30 @@ executeInterchangeFromTileOp(linalg::LinalgOp target,
...
-  nextTargets.clear();
+    return result->op;
+  };
 
-  for (Operation *nextTarget : targets) {
-    if (failed(executeInterchangeFromTileOp(dyn_cast<LinalgOp>(nextTarget),
-                                            tileOp, nextTargets)))
-      return failure();
-  }
+  auto tileSeq = functional::SequenceBuilder()
```
nicolasvasilache, line 217: Beautiful!

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -149,80 +144,51 @@ executePadFromTileOp(linalg::LinalgOp target, linalg::transform::TileOp tileOp,
...
-  return success();
+static FunctionalLinalgTransform
+buildGeneralizeFromTileOpPattern(linalg::transform::TileOp tileOp) {
+  if (!tileOp.generalize())
+    return forwardOp;
+  auto pattern =
+      callFunctional<LinalgGeneralizationPattern>(tileOp.getContext());
+  return [pattern = std::move(pattern)](
+             LinalgOp op, PatternRewriter &rewriter) -> FailureOr<LinalgOp> {
+    FailureOr<GenericOp> result = pattern(op, rewriter);
```
nicolasvasilache, line 163: Feel free to make LinalgGeneralizationPattern::returningMatchAndRewrite return FailureOr<GenericOp> upstream if  you think it is better.

nicolasvasilache (approved):
Awesome!

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
Mogball: `callFunctional`  does invoke `returningMatchAndRewrite` but because some of these return `GenericOp` instead of `LinalgOp`, I have to wrap them again to cast to `LinalgOp`

lib/Transforms/Functional.cpp
Mogball: Let me just remove this from the revision. I was experimenting with a lot of stuff on a separate branch and I just imported all the changes over.

include/Transforms/Functional.h
Mogball: A `ScopeOpInterface` would be any op that is isolated from above.... which might be useful to have actually. I'll look at introducing one in a later patch.

include/Transforms/Functional.h
Mogball: I'll introduce an interface

lib/Transforms/Functional.cpp
Mogball: This predates ScopeOp (which is a special case/utility op) but I removed some of these changes since they aren't needed (for now). I.e. Scope op might be something specific to linalg, whereas the rewrite system in general will work on any op isolated from above (the name is confusing...)

PR #158, opened 2022-01-10
Fix configure script retriggering compilation at every build

nicolasvasilache:


configure.py
```diff
@@ -95,14 +96,20 @@ def main(args):
...
+  # clangpp_path = shutil.which("clang++")
+  # has_clang = False
+  # if clang_path and clangpp_path:
+  #   # llvm_configure_args.append(f"-DCMAKE_C_COMPILER={clang_path}")
+  #   # llvm_configure_args.append(f"-DCMAKE_CXX_COMPILER={clangpp_path}")
+  #   has_clang = True
+  # else:
+  #   print(
+  #       "WARNING: Could not find clang. Building with default system compiler")
+  if shutil.which("clang") is None or shutil.which("clang++") is None:
```
ftynse, line 112: Even if clang _is_ found, this will now build with the default system compiler, so the warning is useless. How about adding an argument to the script, `--use-system-cc`, and wrapping the code above in `if not use_system_cc` instead of commenting it out?

PR #162, opened 2022-01-11
Fix error while tiling Linalg operations using scf.for.

MaheshRavishankar:
Tiling using switch of op -> extract_slice can only give you the tile
sizes of the loops used to access the outputs. The tile sizes of the
other dimension have to be set to untiled values for tiling to work.

### DIFF
--- a/lib/Dialects/LinalgExt/Transforms/Tiling.cpp
+++ b/lib/Dialects/LinalgExt/Transforms/Tiling.cpp
@@ -32,31 +32,39 @@ SmallVector<Value> tileToSCF(PatternRewriter &rewriter, TilingInterface op,
                              TilingInterface clonedOp, ValueRange tileSizes) {
   // Compute lower and upper bounds of the loop nest.
   SmallVector<Range> ranges = clonedOp.getIterationDomain(rewriter);
-  assert(tileSizes.size() == ranges.size() &&
+  assert(tileSizes.size() <= ranges.size() &&
          "expected tile sizes to match the number of loops");
+
+  // Fill the tile sizes with zeros for the untiled dimensions.
+  Location loc = op->getLoc();
+  SmallVector<Value> tileSizesVec(tileSizes.begin(), tileSizes.end());
+  if (ranges.size() != tileSizes.size()) {
+    Value zero = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+    tileSizesVec.resize(ranges.size(), zero);
+  }
+
   SmallVector<Value> lbs, dims, allDims, steps;
   for (auto it : llvm::enumerate(ranges)) {
     allDims.push_back(it.value().size);
-    if (!isZero(tileSizes[it.index()])) {
+    if (!isZero(tileSizesVec[it.index()])) {
       lbs.push_back(it.value().offset);
       dims.push_back(it.value().size);
-      steps.push_back(tileSizes[it.index()]);
+      steps.push_back(tileSizesVec[it.index()]);
     }
   }
 
   // Generate loop nest: One loop per dimension.
   llvm::SmallPtrSet<Operation *, 1> preservedUses;
   SmallVector<Value> destOperand = clonedOp.getDestinationOperands(rewriter);
-  Location loc = op->getLoc();
   auto loopNest = mlir::scf::buildLoopNest(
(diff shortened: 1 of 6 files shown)


lib/Dialects/LinalgExt/Transforms/Tiling.cpp
```diff
@@ -32,39 +32,47 @@ SmallVector<Value> tileToSCF(PatternRewriter &rewriter, TilingInterface op,
                              TilingInterface clonedOp, ValueRange tileSizes) {
   // Compute lower and upper bounds of the loop nest.
   SmallVector<Range> ranges = clonedOp.getIterationDomain(rewriter);
-  assert(tileSizes.size() == ranges.size() &&
+  assert(tileSizes.size() <= ranges.size() &&
```
nicolasvasilache, line 35: Yeah I have something similar to make tiling work upstream with padtensor, feel free to update that one too if you need it first.

lib/Dialects/LinalgExt/Transforms/Tiling.cpp
```diff
@@ -32,39 +32,47 @@ SmallVector<Value> tileToSCF(PatternRewriter &rewriter, TilingInterface op,
...
          "expected tile sizes to match the number of loops");
+
+  // Fill the tile sizes with zeros for the untiled dimensions.
```
nicolasvasilache, line 38: Should the "completion mechanism" be part of the interface itself?

lib/Dialects/LinalgExt/Transforms/TilingExternalModels.cpp
```diff
@@ -103,16 +103,40 @@ struct LinalgOpTilingInterface
                          ArrayRef<OpFoldResult> sizes,
                          bool tileDestOperands) const {
     LinalgOp linalgOp = cast<LinalgOp>(op);
+    if (op->getNumResults() != 1) {
+      // TODO: Need a failure message here, but `notifyMatchFailure` is only a
+      // method on `PatternRewriter`.
+      return {};
```
nicolasvasilache, line 109: Looks like this method should return `FailureOr<SmallVector<Operation*>>`

test/LinalgExt/tiling.mlir
```diff
@@ -44,3 +43,51 @@ func @reverse_1d_tensor(%arg0 : tensor<?x?xf32>) -> tensor<?x?xf32> {
 //      CHECK:       scf.yield %[[INSERT_SLICE]]
 //      CHECK:     scf.yield %[[YIELD]]
 //      CHECK:   return %[[RESULT]]
+
+// -----
+
+func @gemm(%lhs : tensor<?x?xf32>, %rhs : tensor<?x?xf32>,
```
nicolasvasilache, line 49: nit: matmul is not gemm

nicolasvasilache (approved):

lib/Dialects/LinalgExt/Transforms/Tiling.cpp
MaheshRavishankar: ok, I might need that depending on how we handle pad-tensors.

lib/Dialects/LinalgExt/Transforms/Tiling.cpp
MaheshRavishankar: I am back and forth on this. Its a bit opinionated to do that, and actually the completion mechanism is different if it is a tensor semantics op or a memref semantics op. I am waiting to get enough signal to go one way or the other. What I more thinking of is instead of the completion mechanism being part of the interface, we can break it up to have an op publish tile size and offsets of the result the tiled computation updates (using the offsets and sizes specified through the interface). That might be enough to move the completion mechanism to the tiling algorithm. I add actually had that as part of this PR since I thought I needed it, but turns out I didnt.

lib/Dialects/LinalgExt/Transforms/TilingExternalModels.cpp
MaheshRavishankar: Agreed. I think the interface needs to be updated. Ill do that in a follow up.

test/LinalgExt/tiling.mlir
MaheshRavishankar: `linalg.matmul` is almost gemm (missing the scalar values :P )

PR #165, opened 2022-01-11
Introduce an "expert" operation and its expansion with PDL

ftynse:
Introduce a new "expert" operation into the LinalgTransform dialect.
This operation provides a mechanism to associate expert names with lists
of transformations, all described using the PDL dialect. In future, the
PDL IR can be generated from some higher-level frontend.

### DIFF
--- a/lib/Dialects/LinalgTransform/Transforms/ExpertExpansion.cpp
+++ b/lib/Dialects/LinalgTransform/Transforms/ExpertExpansion.cpp
@@ -1,0 +1,108 @@
+//===-- ExpertExpansion.cpp - Expand "expert"s to transforms with PDL -----===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "Dialects/LinalgTransform/LinalgTransformOps.h"
+#include "Dialects/LinalgTransform/Passes.h"
+#include "Dialects/LinalgTransform/SimplePatternRewriter.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/Diagnostics.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Pass/PassRegistry.h"
+#include "mlir/Rewrite/FrozenRewritePatternSet.h"
+#include "mlir/Rewrite/PatternApplicator.h"
+#include "llvm/Support/Debug.h"
+#include <memory>
+
+#define DEBUG_TYPE "expert-expansion"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]")
+
+using namespace mlir;
+using namespace mlir::linalg;
+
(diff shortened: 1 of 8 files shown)


test/LinalgTransform/expert.mlir
```diff
@@ -0,0 +1,89 @@
...
+// CHECK-NOT: @strategies
+// EXPAND-NOT: @strategies
+module @strategies {
+  pdl.pattern @single_tiling : benefit(1) {
+    %tile_sizes = pdl.attribute
+    %vectorize_padding = pdl.attribute
+    %multireduction_lowering = pdl.attribute
+    %name = pdl.attribute : "single_tiling"
+    %target = pdl.attribute
+    %transformed = pdl.type : !pdl.operation
```
Mogball, line 51: Specifying `!pdl.operation` is redundant because the op always returns an operation.

test/LinalgTransform/expert.mlir
```diff
@@ -0,0 +1,89 @@
...
+
+linalg_transform.sequence {
+  // This should match the strategy below.
+  // EXPAND-NOT: expert when
+  // EXPAND: %[[HANDLE:.*]] = tile when @pdl_target {sizes = [4, 4, 4]}
+  // EXPAND: %[[HANDLE2:.*]] = vectorize %[[HANDLE]] {vectorize_padding = true}
+  // EXPAND: bufferize
+  // EXPAND: lower_vectors {multireduction_lowering = "innerreduce"}
+  // EXPAND: lower_to_llvm
+  expert when @pdl_target "single_tiling"
```
nicolasvasilache, line 34: can we have a test for the `$(target^)?` part of the op?

lib/Dialects/LinalgTransform/Transforms/ExpertExpansion.cpp
```diff
@@ -0,0 +1,92 @@
...
+  });
+}
+
+namespace {
+struct ExpertExpansion : public PassWrapper<ExpertExpansion, Pass> {
+  StringRef getArgument() const final {
+    return "linalg-transform-expert-expansion";
+  }
+  StringRef getDescription() const final {
+    return "Expands transformation experts into individual transformations";
```
nicolasvasilache, line 53: Can we beef up the doc a bit?

lib/Dialects/LinalgTransform/Transforms/ExpertExpansion.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (!module)
+      return signalPassFailure();
+
+    ModuleOp strategyModule = nullptr;
+    for (auto nestedModule : module.getOps<ModuleOp>()) {
+      Optional<StringRef> name = nestedModule.sym_name();
+      if (!name)
+        continue;
+
+      if (*name == "strategies") {
```
nicolasvasilache, line 67: Can we make this module name an option to the pass?
Not sure how long you want the "strategy" module name to be load bearing but it is magical atm.

nicolasvasilache (approved):

lib/Dialects/LinalgTransform/Transforms/ExpertExpansion.cpp
ftynse: This is the text that appears in the long list of passes in `mlir-opt --help`, it really should not be longer than this.

PR #168, opened 2022-01-12
Add argument parsing to reduction and add them to benchmarks.

gysit:
- Add argument parsing to reduction
- Add reduction to experiments
- Make sure data is appended if test_harness is called multiple times
- Clear data before running an experiment


nicolasvasilache (commented):
Thanks!

Separately, it would be useful to have the same type of einsum-like spec for reductions so we can more easily create problems. Atm this is still the old hardcoded way.

PR #179, opened 2022-01-13
Make run_tests preserve PYTHONPATH

ftynse:
run_tests.py sets the PYTHONPATH environment variable in a way that
ignores its current value, making it impossible for the caller to affect
it without modifying the script. Append the current value to the new
value instead.

Also dump error messages from the script so that the caller knows what
exactly went wrong.


gysit (approved):
Thanks for the fixes!

nicolasvasilache (approved):

nicolasvasilache:
seems to break CI ?

ftynse:
Quite strange. I can't reproduce this locally.

gysit:
Could it by that removing the iree dependency broke the build and for us stuff is still working since the files are cached? I removed some requirements in the morning https://github.com/gysit/iree-llvm-sandbox/commit/08684a7b35884bcc6557ff8b17682f1261810435 since the downloads failed...

ftynse:
Just too much magic hidden in scripts.

PR #181, opened 2022-01-14
Bufferize InParallelOp

matthias-springer:
The InsertSliceOp is bufferized together with its terminator. By
default, a copy is inserted for every ParallelInsertSliceOp. If the
insertion op bufferizes in-place, the copy will fold away.

The implementation strategy for this op closely follows the one of
InsertSliceOp and we may be able to share some of the code with that op
in the future.


nicolasvasilache (approved):
Thanks @matthias-springer this works e2e for me.
I'll merge for now and may review more deeply in the future.

PR #183, opened 2022-01-15
Add patterns to propagate ops out of warp_execute_on_lane_0 op

ThomasRaoux:
Add patterns to remove dead output as well as patterns to sink out
element and transfer_read operation feeding into a vector_ext.yield

### DIFF
--- a/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
+++ b/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
@@ -31,9 +31,9 @@ static Operation *cloneOpWithOperandsAndTypes(OpBuilder &builder, Location loc,
 
 /// Helper to create a new WarpSingleLaneOp regions with different signature.
 static WarpSingleLaneOp
-moveRegionToNewWarpOpAndOverrideReturns(OpBuilder &b, WarpSingleLaneOp warpOp,
-                                        ValueRange newYieldedValues,
-                                        TypeRange newReturnTypes) {
+moveRegionToNewWarpOpAndReplaceReturns(OpBuilder &b, WarpSingleLaneOp warpOp,
+                                       ValueRange newYieldedValues,
+                                       TypeRange newReturnTypes) {
   // Create a new op before the existing one, with the extra operands.
   OpBuilder::InsertionGuard g(b);
   b.setInsertionPoint(warpOp);
@@ -64,7 +64,7 @@ moveRegionToNewWarpOpAndAppendReturns(OpBuilder &b, WarpSingleLaneOp warpOp,
                                  yield.getOperands().end());
   yieldValues.append(newYieldedValues.begin(), newYieldedValues.end());
   WarpSingleLaneOp newWarpOp =
-      moveRegionToNewWarpOpAndOverrideReturns(b, warpOp, yieldValues, types);
+      moveRegionToNewWarpOpAndReplaceReturns(b, warpOp, yieldValues, types);
   for (auto it :
        llvm::zip(warpOp.getResults(),
                  newWarpOp.getResults().take_front(warpOp.getNumResults())))
@@ -245,7 +245,7 @@ struct WarpOpDeadResult : public OpRewritePattern<WarpSingleLaneOp> {
     }
     if (yield.getNumOperands() == yieldValues.size())
       return failure();
-    WarpSingleLaneOp newWarpOp = moveRegionToNewWarpOpAndOverrideReturns(
+    WarpSingleLaneOp newWarpOp = moveRegionToNewWarpOpAndReplaceReturns(
         rewriter, warpOp, yieldValues, resultTypes);
     unsigned resultIndex = 0;
     for (OpResult result : warpOp.getResults()) {
--- a/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
+++ b/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
@@ -1,0 +1,264 @@
+//===- VectorWarpUtils.cpp - Utilities vector warp ops --------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "Dialects/VectorExt/VectorExtOps.h"
+#include "Dialects/VectorExt/VectorExtWarpUtils.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/Vector/VectorOps.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/BlockAndValueMapping.h"
+#include "mlir/Support/LogicalResult.h"
+
+using namespace mlir;
+using namespace mlir::vector;
+using namespace mlir::vector_ext;
+
+// Clones `op` into a new operations that takes `operands` and returns
+// `resultTypes`.
+static Operation *cloneOpWithOperandsAndTypes(OpBuilder &builder, Location loc,
+                                              Operation *op,
+                                              ArrayRef<Value> operands,
+                                              ArrayRef<Type> resultTypes) {
+  OperationState res(loc, op->getName().getStringRef(), operands, resultTypes,
+                     op->getAttrs());
+  return builder.createOperation(res);
+}
+
+/// Helper to create a new WarpSingleLaneOp regions with different signature.
+static WarpSingleLaneOp
+moveRegionToNewWarpOpAndOverrideReturns(OpBuilder &b, WarpSingleLaneOp warpOp,
+                                        ValueRange newYieldedValues,
+                                        TypeRange newReturnTypes) {
+  // Create a new op before the existing one, with the extra operands.
+  OpBuilder::InsertionGuard g(b);
+  b.setInsertionPoint(warpOp);
+  auto newWarpOp = b.create<WarpSingleLaneOp>(warpOp.getLoc(), newReturnTypes,
+                                              warpOp.laneid());
+
+  Region &opBody = warpOp.getBodyRegion();
+  Region &newOpBody = newWarpOp.getBodyRegion();
+  newOpBody.takeBody(opBody);
+  auto yield =
+      cast<vector_ext::YieldOp>(newOpBody.getBlocks().begin()->getTerminator());
+  yield.operandsMutable().assign(newYieldedValues);
+  return newWarpOp;
+}
+
+/// Helper to create a new WarpSingleLaneOp region with extra outputs.
+static WarpSingleLaneOp
+moveRegionToNewWarpOpAndAppendReturns(OpBuilder &b, WarpSingleLaneOp warpOp,
+                                      ValueRange newYieldedValues,
+                                      TypeRange newReturnTypes) {
+
+  SmallVector<Type> types(warpOp.getResultTypes().begin(),
+                          warpOp.getResultTypes().end());
+  types.append(newReturnTypes.begin(), newReturnTypes.end());
+  auto yield = cast<vector_ext::YieldOp>(
+      warpOp.getBodyRegion().getBlocks().begin()->getTerminator());
+  SmallVector<Value> yieldValues(yield.getOperands().begin(),
+                                 yield.getOperands().end());
+  yieldValues.append(newYieldedValues.begin(), newYieldedValues.end());
+  WarpSingleLaneOp newWarpOp =
+      moveRegionToNewWarpOpAndOverrideReturns(b, warpOp, yieldValues, types);
+  for (auto it :
+       llvm::zip(warpOp.getResults(),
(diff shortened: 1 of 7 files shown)


lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+};
+
+/// Remove any result that has no use along with the matching yieldOp operand.
+// TODO: Move this in WarpSingleLaneOp canonicalization.
+struct WarpOpDeadResult : public OpRewritePattern<WarpSingleLaneOp> {
+  using OpRewritePattern<WarpSingleLaneOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(WarpSingleLaneOp warpOp,
+                                PatternRewriter &rewriter) const override {
+    SmallVector<Type> resultTypes;
+    SmallVector<Value> yieldValue;
```
nicolasvasilache, line 230: bit: `yieldValues` (plural)

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+        warpOp.getBodyRegion().getBlocks().begin()->getTerminator());
+    for (OpResult result : warpOp.getResults()) {
+      if (result.use_empty()) {
+        hasDeadResults = true;
+        continue;
+      }
+      resultTypes.push_back(result.getType());
+      yieldValue.push_back(yield.getOperand(result.getResultNumber()));
+    }
+    if (!hasDeadResults)
```
nicolasvasilache, line 242: -> `if (yield.getNumOperands() == yieldValues.size())` and drop `hasDeadResults` everywhere

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+      moveRegionToNewWarpSignature(b, warpOp, yieldValue, types);
+  for (auto it :
+       llvm::zip(warpOp.getResults(),
+                 newWarpOp.getResults().take_front(warpOp.getNumResults())))
+    std::get<0>(it).replaceAllUsesWith(std::get<1>(it));
+  return newWarpOp;
+}
+
+llvm::Optional<std::pair<Operation *, unsigned>>
+getWarpResult(WarpSingleLaneOp warpOp, std::function<bool(Operation *)> fn) {
```
nicolasvasilache, line 76: Seems like this should be `OpOperand *getWarpYieldOperand` ?

You can then easily get to the defining op, the result number, operand number etc from just that?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+    Operation *definedOp = yieldValue.getDefiningOp();
+    if (definedOp && fn(definedOp)) {
+      if (!warpOp.getResult(yieldOperand.getOperandNumber()).use_empty())
+        return std::make_pair(definedOp, yieldOperand.getOperandNumber());
+    }
+  }
+  return {};
+}
+
+// Currently the distribution map is implicit based on the vector shape. In the
```
nicolasvasilache, line 90: Please give an example of the current implicit behavior.
Also, seems to be more naturally expressible as an API on vector shapes?
```
static AffineMap calculateImplicitMap(ArrayRef<int64_t> shape, ArrayRef<int64_t> subShape) {
  int64_t i =0;
  for (auto it = llvm::zip(shape, subShape)) {
    assert(std::get<0>(it) % std::get<1>(it) == 0 ...);
    if (std::get<0>(it) != std::get<1>(it)) {
      AffineExpr d = getAffineDimExpr(i++, yield.getContext());
      perm.push_back(d * (std::get<0>(it) / std::get<1>(it))); // I think I'd put the scale in here too.
    }
  }
  return AffineMap::get(srcType.getRank(), 0, perm, yield.getContext());
}

File: lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp [hunk 5]
```diff
@@ -0,0 +1,261 @@
...
+    rewriter.setInsertionPoint(warpOp);
+    for (auto it : llvm::zip(indexMap.getResults(), map.getResults())) {
+      AffineExpr d0, d1;
+      bindDims(read.getContext(), d0, d1);
+      auto indexExpr = std::get<0>(it).dyn_cast<AffineDimExpr>();
+      if (!indexExpr)
+        continue;
+      unsigned indexPos = indexExpr.getPosition();
+      unsigned vectorPos = std::get<1>(it).cast<AffineDimExpr>().getPosition();
+      auto scale = getAffineConstantExpr(
```
nicolasvasilache, line 207: `AffineExpr::operator*` works with an `int64_t`.

But I think if you make it part of the implicit map you'd get that for free from composition?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+      return failure();
+    auto read = cast<vector::TransferReadOp>(operand->first);
+    Value distributedVal = warpOp.getResult(operand->second);
+
+    SmallVector<Value, 4> indices(read.indices().begin(), read.indices().end());
+    AffineMap map = calculateImplicitMap(read.getResult(), distributedVal);
+    AffineMap indexMap = map.compose(read.permutation_map());
+    OpBuilder::InsertionGuard g(rewriter);
+    rewriter.setInsertionPoint(warpOp);
+    for (auto it : llvm::zip(indexMap.getResults(), map.getResults())) {
```
nicolasvasilache, line 199: I think most of this loop goes away if you embed the multiplicity in the affine map ?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+                             TypeRange newReturnTypes) {
+  // Create a new op before the existing one, with the extra operands.
+  OpBuilder::InsertionGuard g(b);
+  b.setInsertionPoint(warpOp);
+  auto newWarpOp = b.create<WarpSingleLaneOp>(warpOp.getLoc(), newReturnTypes,
+                                              warpOp.laneid());
+
+  Region &opBody = warpOp.getBodyRegion();
+  Region &newOpBody = newWarpOp.getBodyRegion();
+  newOpBody.takeBody(opBody);
```
nicolasvasilache, line 45: not completely clear whether this is safe re. listeners; double-check plz?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+  auto yield =
+      cast<vector_ext::YieldOp>(newOpBody.getBlocks().begin()->getTerminator());
+  yield.operandsMutable().assign(newYieldedValues);
+  return newWarpOp;
+}
+
+/// Helper to create a new WarpSingleLaneOp region with extra outputs.
+static WarpSingleLaneOp
+moveRegionToNewWarpAddReturn(OpBuilder &b, WarpSingleLaneOp warpOp,
```
nicolasvasilache, line 54: naming makes it hard to understand what the "add" part does here
`moveRegionToNewWarpOpAndAppendReturns`

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+                                              ArrayRef<Value> operands,
+                                              ArrayRef<Type> resultTypes) {
+  OperationState res(loc, op->getName().getStringRef(), operands, resultTypes,
+                     op->getAttrs());
+  return builder.createOperation(res);
+}
+
+/// Helper to create a new WarpSingleLaneOp regions with different signature.
+static WarpSingleLaneOp
+moveRegionToNewWarpSignature(OpBuilder &b, WarpSingleLaneOp warpOp,
```
nicolasvasilache, line 34: better naming would be great.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+    auto yieldOperand = getWarpResult(warpOp, [](Operation *op) {
+      return OpTrait::hasElementwiseMappableTraits(op);
+    });
+    if (!yieldOperand)
+      return failure();
+    Operation *elementWise = yieldOperand->first;
+    Value distributedVal = warpOp.getResult(yieldOperand->second);
+    SmallVector<Value> yieldValues;
+    SmallVector<Type> retTypes;
+    SmallVector<unsigned> operandForwarded;
```
nicolasvasilache, line 137: drop this?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,261 @@
...
+          operand.get().getType().cast<VectorType>().getElementType());
+      operandForwarded.push_back(operand.getOperandNumber());
+      retTypes.push_back(targetType);
+      yieldValues.push_back(operand.get());
+    }
+    WarpSingleLaneOp newWarpOp =
+        moveRegionToNewWarpAddReturn(rewriter, warpOp, yieldValues, retTypes);
+    SmallVector<Value> newOperands(elementWise->getOperands().begin(),
+                                   elementWise->getOperands().end());
+    for (unsigned i : operandForwarded) {
```
nicolasvasilache, line 150: seems like this is always `llvm::seq(0, elementWise->getNumOperands())` ?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: I had checked on internal mlir chat and people said it was okay. This seems to do pretty much the same as `rewriter.mergeBlock`

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: Yes I was considering doing that but OpOperand get invalidated once I change the yield operand so this avoid potential mistake of using OpOperand after it being killed.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: I added an example. I'm considering allowing non vector values in the future so keeping the value seems more flexible for now.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: changed it to int64_t. I'll think more about how to make the map contain the multiplicity.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: oops, yes dropped.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -0,0 +1,264 @@
...
+                                              ArrayRef<Value> operands,
+                                              ArrayRef<Type> resultTypes) {
+  OperationState res(loc, op->getName().getStringRef(), operands, resultTypes,
+                     op->getAttrs());
+  return builder.createOperation(res);
+}
+
+/// Helper to create a new WarpSingleLaneOp regions with different signature.
+static WarpSingleLaneOp
+moveRegionToNewWarpOpAndOverrideReturns(OpBuilder &b, WarpSingleLaneOp warpOp,
```
nicolasvasilache, line 34: nit: override -> replace.
Override has specific meanings that don't seem to match here.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
nicolasvasilache: well that sounds like a red herring; if you need to memoize, you'd just:
```
Operation *op = yieldOperand->get().getDefiningOp();
int64_t operandIndex = yieldOperand->getOperandNumber();
```
and then not use `yieldOperand`.

It is bad form to be creative with types and APIs for things that are well-supported.

All the `yieldOperand->first` and `yieldOperand->second` should disappear.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
nicolasvasilache: SG, but please don't let it hanging, it will be significantly better if you can make it work.

nicolasvasilache (approved):
Approved conditioned on dropping API creativity.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: That makes sense. I changed the signature as suggested.

PR #185, opened 2022-01-17
Extend einsum spec to support one-operand operators.

gysit:
- Add support for a single operand.
- Add support for scalar outputs.

### DIFF
--- a/python/examples/contraction/definitions.py
+++ b/python/examples/contraction/definitions.py
@@ -21,7 +21,7 @@ class EinsumProblem(ProblemDefinition):
   specification of the operation is similar to that of np.einsum.
   """
 
-  def __init__(self, specification: str, flop_count: float):
+  def __init__(self, specification: str, flop_count_per_iter: int):
     """Creates a new EinsumProblem with the given specification.
 
     The specification is a string of the format:
@@ -36,10 +36,10 @@ class EinsumProblem(ProblemDefinition):
 
     Arguments:
     specification: textual specification of the einsum.
-    flop_count: floating-point operations executed per iteration.
+    flop_count_per_iter: floating-point operations executed per iteration.
     """
     self.specification = EinsumSpecification(specification)
-    self.flop_count = flop_count
+    self.flop_count_per_iter = flop_count_per_iter
 
   @property
   def keys(self) -> List[str]:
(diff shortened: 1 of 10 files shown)


python/examples/contraction/definitions.py
```diff
@@ -17,25 +17,29 @@
...
   """Benchmarking problem definition for einsum.
 
-  Currently only supports two-operand contractions. The textual specification of
-  the contraction to perform is similar to that of np.einsum.
+  Supports one-operand and two-operand einsum specifications. The textual
+  specification of the operation is similar to that of np.einsum.
   """
 
-  def __init__(self, specification: str):
+  def __init__(self, specification: str, flop_count: float):
```
nicolasvasilache, line 24: flop_count_per_iteration?
Also, this is an int, not a float.

python/examples/contraction/definitions.py
```diff
@@ -45,33 +49,37 @@ def keys(self) -> List[str]:
...
+    if self.specification.reduction_dims:
+      gbyte_count += 1.e-9 * \
+          (np.prod(shapes[-1]) * np.dtype(types[-1]).itemsize)
+    return gbyte_count
 
   def tensors_np_builder(self, sizes: Mapping[str, Any],
                          types: Sequence[np.dtype]) -> List[np.dtype]:
     """Returns random NumPy suitable for calling the kernel."""
-    shapes = self.shapes_builder(sizes)
+    shapes = [s if s else [1] for s in self.shapes_builder(sizes)]
```
nicolasvasilache, line 82: I do not follow this change, why is `[1]` the right answer? Is it because it is just ignored ? (maybe add some extra doc for the future?)

nicolasvasilache (approved):

python/examples/contraction/definitions.py
gysit: Ah this is to handle the scalar case. If we do not have a size we want to get a numpy scalar tensor which has size one. It is not new code we use it in other parts of the code base.

python/examples/contraction/definitions.py
gysit: sure

python/examples/contraction/definitions.py
nicolasvasilache: ok

PR #191, opened 2022-01-18
Fix build/test failures around WalkStage

matthias-springer:
This functionality has recently moved to MLIR. Another cleanup commit is
needed to remove old code from the Sandbox.


nicolasvasilache (approved):
@dcaballe please update to make a proper usage of the recent walker changes.

lib/Dialects/VectorExt/Transform/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
+set(LLVM_OPTIONAL_SOURCES
```
ftynse, line 1: You can add `PARTIAL_SOURCES_INTENDED` to `add_mlir_library` and avoid listing all files here (or, better, have one library per directory).

lib/Dialects/VectorExt/Transform/VectorMaskingUtils.cpp
```diff
@@ -216,21 +217,28 @@ LogicalResult mlir::vector_ext::maskVectorPredicateOps(
...
+
   return success();
 }
 
 /// Masking strategy that only masks vector transfer operations and operations
 /// with side effects. Non-side-effecting ops are left unmasked.
 void mlir::vector_ext::maskGenericOpWithSideEffects(
     OpBuilder &builder, Operation *op, Value activeMask,
-    const WalkStage &stage) {
+    const WalkStage &stage, SmallVector<Operation *> &erasedOps) {
```
ftynse, line 241: SmallVectorImpl, no need to tie a specific size

lib/Dialects/VectorExt/Transform/CMakeLists.txt
matthias-springer: I'll leave this as is because the file will be removed entirely very soon. But this is very useful for some of my other changes...

PR #193, opened 2022-01-19
[VectorExt] Remove generic visitor dead code

dcaballe:
The generic visitor was upstreamed to MLIR core.


nicolasvasilache (approved):
Thanks!

PR #200, opened 2022-01-19
Workaround memref.copy perf problems

nicolasvasilache:


nicolasvasilache:
@gysit note I am not yet able to make the 2d case work with the args parser.
The issue is that the expert tile sizes is dependent on problem sizes.

We can land this which has the problem sizes to measure BW in the 2d case (the 1-d case is not adaptive and does not reach the same level of perf).

gysit (approved):

gysit:
Transpose 2D is an example that computes many experts with different tile sizes. Selecting the expert is not possible in this case.

PR #204, opened 2022-01-20
[DoNotCommit] IR Repro

nicolasvasilache:
```
python -- -m python.examples.conv.conv_1d_bench --expert_list SingleTiling3DPad --dynamic_at_compile_time_list [] --problem_sizes_list 1,988,32,3,64,[1],[1]
```

Alternatively, set some names in `--dynamic_at_compile_time_list []` to add some `?` to the IR.


nicolasvasilache:
This has served its purpose.

PR #206, opened 2022-01-21
Add integration test for Vector GPU

ThomasRaoux:


ThomasRaoux:
Only the second commit is the test first commit is from https://github.com/google/iree-llvm-sandbox/pull/202

ThomasRaoux:
Is there an easy way to disable the gpu integration test when building without gpu support?

configure.py
```diff
@@ -67,6 +67,13 @@ def parse_arguments():
       default=False,
       action=argparse.BooleanOptionalAction,
   )
+  parser.add_argument(
+      "--cuda-runner",
```
matthias-springer, line 71: What happens when I run `configure.py` without this argument and then run all tests? (run_tests.py)

matthias-springer (approved):

configure.py
ThomasRaoux: Good point, I talked to Nicolas about that and I do need to change the script to only enable cuda tests on demand

ThomasRaoux:
@nicolasvasilache @matthias-springer I rebased and added a command line option to enable gpu integration tests since they require a gpu and cuda setup. Let me know if this looks okay to you.

PR #213, opened 2022-01-25
[VectorExt] Add indices and incoming mask to vector.predicate

dcaballe:
This patch extends the vector.predicate op to:
  1) take in a list of indices to map each dimension of the
     vector mask to a vector dimension of the iteration space.
  2) take in an incoming mask that is combined with the predicate mask
     to generate the "true" mask (need a better name) influencing the region.
  3) generate the "true" mask as a region argument.

### DIFF
--- a/include/Dialects/VectorExt/VectorExtOps.td
+++ b/include/Dialects/VectorExt/VectorExtOps.td
@@ -21,24 +21,25 @@ class VectorExt_Op<string mnemonic, list<OpTrait> traits = []> :
 
 def VectorExt_PredicateOp : VectorExt_Op<"predicate",
       [DeclareOpInterfaceMethods<RegionBranchOpInterface>,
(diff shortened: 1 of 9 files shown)


dcaballe:
Kind ping to the reviewers :)

gysit (approved):
LGTM! 

If the goal of the incoming mask is to support nested predication, "parent mask" or "outer mask" could be possible naming options?

dcaballe:
Thanks, Tobias! 

Good point! The problem is that for some cases the "incoming" mask could come from "the inside". For example, an `if` operation may produce a mask that is used as "incoming" mask outside the `if` operation. IIRC, this is called `BlockInMask` in LLVM, where "In" stands for incoming or input (?). We can definitely revisit the name when/if we move this to core. More suggestions are welcome, though!

gysit:
Thanks for the explanation. If there is a LLVM counterpart it definitely makes sense to stick to it.

PR #224, opened 2022-01-30
Integration fixes for recent changes.

matthias-springer:
Note: Do not submit yet. This still has failures related to non-bufferization changes.


matthias-springer:
Not up to date to HEAD yet, but this already fixes some issues.

tools/mlir-proto-opt/CMakeLists.txt
```diff
@@ -4,14 +4,35 @@ if (SANDBOX_ENABLE_ALP)
   )
 endif()
 
+get_property(dialect_libs GLOBAL PROPERTY MLIR_DIALECT_LIBS)
```
ftynse, line 7: I think this will override `dialect_libs` from line 2 above.

tools/mlir-proto-opt/CMakeLists.txt
```diff
@@ -4,14 +4,35 @@ if (SANDBOX_ENABLE_ALP)
...
+get_property(conversion_libs GLOBAL PROPERTY MLIR_CONVERSION_LIBS)
+
 add_llvm_executable(mlir-proto-opt
   mlir-proto-opt.cpp
 )
 
+set(LIBS
+  ${dialect_libs}
+  ${conversion_libs}
+  ${test_libs}
```
ftynse, line 17: Where is `test_libs` defined?

tools/mlir-proto-opt/mlir-proto-opt.cpp
```diff
@@ -23,8 +24,11 @@ using namespace mlir;
 
 int main(int argc, char **argv) {
   llvm::InitLLVM y(argc, argv);
+  registerAllPasses();
   registerOutsideOfDialectRegistry();
 
+  // TODO: Find a way to register test passes here.
```
ftynse, line 30: Add forward declarations above and call them here?

matthias-springer: `mlir-opt.cpp` has a long list of test passes to be registered. It seems impractible to copy the entire list. It would go out-of-sync pretty fast. Maybe there's a way to put these into a separate function in MLIR. Something like `registerAllTestPasses`. It seems like we don't depend on any MLIR test passes in the Sandbox yet, so we could keep the TODO for the moment.

tools/mlir-proto-opt/CMakeLists.txt
matthias-springer: Not defined and not needed until we start registering test passes.

tools/mlir-proto-opt/mlir-proto-opt.cpp
ftynse: We _shouldn't_ depend on MLIR test passes, that's why they are hidden. So I wouldn't even add a TODO here :)

nicolasvasilache (approved):

PR #229, opened 2022-02-02
Avoid keeping pointers to deleted ops in transform interpreter

ftynse:
The transform interpreter maintains a mapping between the values
produced by the transform dialect operations and (sets of) operations in
the main IR to transform. Each transformation is likely to change the
tracked operation so the transform dialect diallows more than one use.
However, the pointer to the operation may still be kept in the mapping,
leading to use-after-free (or, in the case where the pointer address is
reused for a different operation, to the transform interpreter accessing
unexpected operations). Leverage the single-use restriction and clean up
the mapping immediately after the corresponding value is "consumed" by
the transform dialect operation.

### DIFF
--- a/test/LinalgTransform/failure.mlir
+++ b/test/LinalgTransform/failure.mlir
@@ -26,3 +26,65 @@ linalg_transform.sequence {
   vectorize when @target_pattern
 }
 
+// -----
+
+#map0 = affine_map<(d0, d1, d2) -> (d0, d2)>
+#map1 = affine_map<(d0, d1, d2) -> (d2, d1)>
+#map2 = affine_map<(d0, d1, d2) -> (d0, d1)>
+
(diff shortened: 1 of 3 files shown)


test/LinalgTransform/failure.mlir
```diff
@@ -26,3 +26,65 @@ linalg_transform.sequence {
...
+  apply_native_constraint "nestedInFunc" [@benchmark](%2 : !pdl.operation)
+  rewrite %2 with "linalg_transform.apply"
+}
+
+linalg_transform.sequence {
+  %0 = tile when @target_pattern {interchange = [0, 2, 1],  sizes = [3, 5, 14]}
+  %1 = tile %0 { sizes = [3, 5, 2]}
+  // expected-error@below {{failed to apply}}
+  %2 = vectorize %1 {vectorize_padding = true}
+}
```
gysit, line 58: Do this tests fail due to dynamic dimensions that prevent vectorization?

gysit (approved):

test/LinalgTransform/failure.mlir
ftynse: Yes.

PR #231, opened 2022-02-02
Transform dialect: don't interchange twice

ftynse:
The linalg tiling pattern subsumes interchange, along with a bunch of
other conflated things. Don't apply the interchange pattern separately
during tiling.

One day this will hopefully be deconflated so that tiling does what its
name says it does...


test/LinalgTransform/tile_interchange.mlir
```diff
@@ -0,0 +1,72 @@
...
+  %1 = types
+  %2 = operation "linalg.generic"(%0 : !pdl.range<value>)  -> (%1 : !pdl.range<type>)
+  apply_native_constraint "nestedInFunc" [@matmul_210](%2 : !pdl.operation)
+  rewrite %2 with "linalg_transform.apply"
+}
+
+linalg_transform.sequence {
+  %0 = tile when @target_pattern {interchange = [2, 1, 0],  sizes = [3, 5, 14]}
+  %1 = tile %0 { sizes = [3, 5, 2]}
+  // expected-error@below {{failed to apply}}
```
gysit, line 70: shouldn't these be gone now?

gysit (approved):
LGTM!

True now that we are much more flexible in terms of combining transformations we should split them up into simpler pieces and combine them!

PR #232, opened 2022-02-02
Bump LLVM to b82a3a8ef3843a8b98807219db87bc12f38ed704

ftynse:
Leverage transfer read/write fix from https://reviews.llvm.org/D118758.
Account for FunctionPass deprecation.


gysit (approved):
Thanks!

PR #238, opened 2022-02-04
[DoNotCommit] Repro lack of pipelining

nicolasvasilache:
The following command is expected to dump some pipelined IR but it does not:
```
python -m python.examples.matmul.test
```

### DIFF
--- a/lib/LinalgTensorCodegenDriver.cpp
+++ b/lib/LinalgTensorCodegenDriver.cpp
@@ -551,7 +551,7 @@ loopScheduling(scf::ForOp forOp,
     unsigned earlyCycle = 0;
     for (Value operand : op.getOperands()) {
       Operation *def = operand.getDefiningOp();
-      if (!def)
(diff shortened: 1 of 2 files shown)


nicolasvasilache:
FYI @ThomasRaoux , this repro does not pipeline.
There are other cases that also don't pipeline (e.g. if you remove the interchange vector in the strategy and the vector is hoisted into an iter_args; then I'd try to pipeline the 2nd enclosing loop for instance).

ThomasRaoux:
Summarizing what I wrote on discord:
-There is a recurrence in the loop containing read_ + contract + write so we cannot pipeline
-Technically there is no "real" dependency across iterations however it would require some special handling and this would make the tensor liverange overlap across iteration.

Do we want to handle this kind of case?

ThomasRaoux:
Pushed the new example with bigger II and no interchange.
(the small II was causing the number of stages to be greater than number of iterations)

nicolasvasilache:
works for me, thanks!

PR #241, opened 2022-02-04
Run enabler transformations in the transform dialect interpreter

ftynse:
These were initially omitted from the interpreter for simplicity. Add
them back to ensure parity with the code generation strategy. In the
latter, enablers seem optional, but in practice are run always except
for a single test verifying that they can be omitted.

This exposes the redundancy and the likely overhead in the enabler
transformations running between every other pairs of transformations,
while only some of the enablers are necessary for some of the
transformations (e.g. no need to simplify single-iteration loops if no
new loops are introduced, no need to hoist loop-invariant transfers if
no transfers are introduced, etc.) and that CSE effectively runs twice
in a row.

### DIFF
--- a/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
+++ b/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
@@ -29,6 +29,7 @@
 #include "mlir/Conversion/SCFToStandard/SCFToStandard.h"
 #include "mlir/Conversion/StandardToLLVM/ConvertStandardToLLVMPass.h"
 #include "mlir/Conversion/VectorToLLVM/ConvertVectorToLLVM.h"
+#include "mlir/Dialect/Affine/LoopUtils.h"
 #include "mlir/Dialect/Arithmetic/Transforms/BufferizableOpInterfaceImpl.h"
 #include "mlir/Dialect/Bufferization/IR/Bufferization.h"
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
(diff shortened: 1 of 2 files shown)


lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -393,6 +397,56 @@ executeLowerToLLVMOp(ModuleOp module,
...
+          return moveLoopInvariantCode(loopLike);
+        });
+    if (result.wasInterrupted())
+      return failure();
+  }
+
+  func.walk([](Operation *op) {
+    (void)llvm::TypeSwitch<Operation *, LogicalResult>(op)
+        .Case<AffineForOp, scf::ForOp>(
+            [](auto loop) { return promoteIfSingleIteration(loop); })
```
ftynse, line 425: Anybody knows why this is not a canonicalization pattern?

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -393,6 +397,56 @@ executeLowerToLLVMOp(ModuleOp module,
...
+        .Default([](Operation *) { return success(); });
+  });
+
+  if (options.hoistRedundantVectorTransfers)
+    hoistRedundantVectorTransfers(func);
+  if (options.hoistRedundantVectorTransfersOnTensor)
+    hoistRedundantVectorTransfersOnTensor(func);
```
ftynse, line 432: Any reason why these are not handled by normal LICM?

nicolasvasilache: This requires an analysis of read/write pairs to check if "canBeHoisted" is true.
This is beyond what LICM supports.
Extending behavior with a more general interface is "on-a-per-need-basis" (i.e. not done).

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
nicolasvasilache: History I think, this likely originated a long time ago in affine and whoever ported it to scf did not push further.

nicolasvasilache (approved):

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
gysit: At least single iteration scf::ForOp should be canonicalized by SimplifyTrivialLoops.

PR #242, opened 2022-02-04
Refactor harness and add a minimal example with a PDL + LinalgTransfo…

nicolasvasilache:
…rm schedule builder

### DIFF
--- a/python/examples/core/harness.py
+++ b/python/examples/core/harness.py
@@ -21,6 +21,7 @@ from ..core.compilation import compile_to_execution_engine, \
     emit_benchmarking_function
 from ..core.experts import TransformationList
 from ..core.problem_definition import *
+from ..core.transforms import ApplySchedule
 from ..core.utils import *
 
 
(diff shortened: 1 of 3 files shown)


nicolasvasilache:
I also have this rebased on top of @Mogball 's MatchOp so I can experiment a bit with both and see whether surprises occur.

python/examples/core/harness.py
```diff
@@ -249,6 +250,60 @@ def __assert_matching_mapping_keys(self, mapping: Mapping[str, Any]):
...
+      func = self.problem_definition.build_problem_under_context_manager(
+          fun_to_benchmark_name, types, zero_at_each_iteration)
+      wrapper = emit_benchmarking_function(entry_point_name, func)
+
+  # Must be called under ContextManager with Context() and Location()
+  def _compile(
+      self,
+      module,
+      # TODO: Better type than Callable.
+      transform: Callable,
```
ftynse, line 274: You can at least have `Callable[[ModuleOp], None]`

python/examples/core/harness.py
```diff
@@ -249,6 +250,60 @@ def __assert_matching_mapping_keys(self, mapping: Mapping[str, Any]):
...
+      types = self.problem_definition.types_mlir_builder(
+          self.compile_time_problem_sizes_dict,
+          compiled_function_element_types_mlir_builder(self.np_types))
+
+      func = self.problem_definition.build_problem_under_context_manager(
+          fun_to_benchmark_name, types, zero_at_each_iteration)
+      wrapper = emit_benchmarking_function(entry_point_name, func)
+
+  # Must be called under ContextManager with Context() and Location()
+  def _compile(
```
ftynse, line 270: Nit: `_compile_to_execution_engine` so the difference with `compile` is clear?

python/examples/core/transforms.py
```diff
@@ -536,6 +536,25 @@ def __init__(self, fun_name: str, op_name: str, result_func_name: str,
...
+  def __init__(self):
+    pass
+
+  def __call__(self, module: Module, **kwargs):
+    PassManager.parse('linalg-interp-transforms').run(module)
+    self.drop_schedule_from_module(module)
+    return module
+
+  # TODO: we should not need that; PDLToLLVM and LinalgTransformToLLVM should
+  # just drop automatically?
```
ftynse, line 550: There's no PDLToLLVM or LinalgTransformToLLVM :)  Only the interpreter pass. I saw value in having the interpreter pass keep the transformation sequence as a sort of documentation/debugging aid, but it can be hidden under an option.

python/examples/matmul/test_schedule_dialect.py
```diff
@@ -0,0 +1,69 @@
...
+
+# CHECK-NOT: FAILURE
+
+
+def add_tile_schedule(module):
+  with InsertionPoint(module.body):
+    pdl_pattern_name = 'pdl_pattern_name'
+    pdl_pattern = pdl.PatternOp(\
+      1, #=benefit 
+      pdl_pattern_name#=sym_name
```
ftynse, line 31: Nit: this is Python, so we should be able to write `benefit=1, sym_name=pdl_pattern_name`.

python/examples/matmul/test_schedule_dialect.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+    with ir.InsertionPoint(pdl_pattern.body):
+      args = pdl.OperandsOp()
+      types = pdl.TypesOp()
+      pdl_op = pdl.OperationOp('linalg.generic', args=[args], types=[types])
+      pdl.RewriteOp(pdl_op, 'linalg_transform.apply')
+
+    sequence = transform.SequenceOp()
+    with ir.InsertionPoint(sequence.body.blocks[0]):
+      transform.TileOp(pdl_pattern_name, sizes=[12, 32, 1], pad=False)
```
ftynse, line 41: Side note: If you want the same behavior as codegen strategy wrt padding, you'd have to write two independent `TileOp`s -- one to tile, another to pad the result of the previous tiling. Otherwise, the enabler pass won't run between them which _may_ produce different padded shapes.

ftynse (approved):

python/examples/matmul/test_schedule_dialect.py
nicolasvasilache: ack

python/examples/matmul/test_schedule_dialect.py
nicolasvasilache: interestingly, this complains about unexpected `sym_name` which makes me think we are not calling the function I am expecting .. could you patch and try on your end?

nicolasvasilache: sigh .. thanks!

PR #247, opened 2022-02-07
Refactoring ALP

giuseros:
This is a big refactor of our experimental codebase to follow latest developments:
- Generate functions through the Python DSL interface
- Use the trasnformation infra common in the sandbox
- Generalize tuner/mlirc to be program agnostic (although transformations are still GEMM-specific)
- Add a python transformation to save the IR
- Make `configure.py` compatible with python 3.7


giuseros:
cc @nicolasvasilache @stevenvar

configure.py
```diff
@@ -36,36 +36,36 @@ def parse_arguments():
       help="Build with ENABLE_LLD=ON (optional)",
       dest="enable_lld",
       default=True,
-      action=argparse.BooleanOptionalAction,
+      action="store_true",
```
nicolasvasilache, line 39: could you briefly explain why you need this ?

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
```diff
@@ -192,147 +225,140 @@ std::unique_ptr<mlir::OperationPass<ModuleOp>> mlir::createExtractKernelPass() {
...
+    Block *tail = rewriter.splitBlock(loop->getBlock(),
+                                      Block::iterator(loop->getNextNode()));
+    extract_function_2("kernel_tail", tail, module, rewriter, loop->getBlock(),
+                       module.getLoc());
   }
 };
 
-std::unique_ptr<mlir::OperationPass<ModuleOp>> mlir::createExtractKernelTailPass() {
+std::unique_ptr<mlir::OperationPass<ModuleOp>>
+mlir::createExtractKernelTailPass() {
```
nicolasvasilache, line 362: I would be interested in synchronizing on what this does and whether it can be subsumed by peeling + the recent transform to outline some parent loop given an op.
My hope is we could replace this file with more standard stuff but I am not sure I have the details right.

experimental/alp/include/alp/Transforms/Passes.td
```diff
@@ -52,4 +54,14 @@ def LegalizePass: FunctionPass<"alp-legalize"> {
   ];
 }
 
+def ForToDoWhileLoop: Pass<"alp-for-to-dowhile", "FuncOp"> {
+  let summary = "Pass to legalize vector operations.";
```
nicolasvasilache, line 58: It is interesting to see such a transformation is needed.
Could you please document and explain why it is necessary?

experimental/alp/lib/Transforms/legalize_vector.cpp
```diff
@@ -98,103 +107,108 @@ struct ForOpVectorProgates : public OpRewritePattern<scf::ForOp> {
...
         UnrollVectorOptions().setNativeShapeFn(getShape).setFilterConstraint(
             filter));
-    (void)applyPatternsAndFoldGreedily(getFunction(), std::move(patterns));
+    (void)applyPatternsAndFoldGreedily(getOperation(), std::move(patterns));
 
     // Now there will be a lot of:
     // - transfer_read/insert_strided_slices
     // - transfer_write/extract_strided_slices
-    // Next step is to remove as many {insert,extract}_strided_slices as we can, especially across loop blocks
+    // Next step is to remove as many {insert,extract}_strided_slices as we can,
```
nicolasvasilache, line 158: This feels like it should be covered by union of canonicalization and folding patterns.
Could you share examples of IR before and after applying this transformations?
This way we can track down what is missing here.

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
```diff
@@ -170,6 +172,7 @@ namespace{
 };
 } // namespace
 
-std::unique_ptr<mlir::FunctionPass> mlir::createModuloSchedulingPass() {
+std::unique_ptr<mlir::OperationPass<FuncOp>>
+mlir::createModuloSchedulingPass() {
```
nicolasvasilache, line 176: I would similarly hope we use the pipelining pas, could you share IR before and after so it is easier to follow what is going on?

experimental/alp/python/alp/backend/codegen.py
```diff
@@ -0,0 +1,25 @@
...
+
+def codegen(source, dest, scheduler="ilpmax"):
+    cmd = ["$IREE_LLVM_SANDBOX_BUILD_DIR/bin/llc"]
+    cmd.append(source)
+    cmd.append("-filetype=asm")
+    cmd.append("-O3")
+
+    # Scheduling options
+    if scheduler == "ilpmax":
+        cmd.append("--misched=ilpmax")
```
nicolasvasilache, line 17: interesting! I'd be interested in some details here, is this for your specific HW or something generic?

experimental/alp/python/alp/benchmark/blas/gemm.py
```diff
@@ -0,0 +1,141 @@
...
+    size = [args.M, args.N, args.K]
+    dest = args.output if args.output else args.function_name + ".bench.mlir"
+    generate_benchmark_mlir(args.function_name, args.trA, size, args.reps, dest)
+
+    if args.asm_program:
+        exe = build(args.asm_program, dest)
+        print(run(exe))
+
+
+if __name__ == "__main__":
```
nicolasvasilache, line 140: Seems a bit weird that you have to reproduce this file, I'd hope we could reuse.
I'll try to plug in opentuner see if we can reduce the amount of duplication.

experimental/alp/python/alp/benchmark/infra.py
```diff
@@ -0,0 +1,73 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+#  See https://llvm.org/LICENSE.txt for license information.
+#  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import os
```
nicolasvasilache, line 5: Same comment here re duplication of infra, esp as we move to the scheduling dialect.

python/examples/core/transform.py
```diff
@@ -10,306 +11,377 @@
...
-        created by passing all parameters required by individual
-        transformations as keyword arguments to the list constructor.
-        """
-        assert issubclass(other_cls, (Transform, TransformationList))
-        name = objtype.__name__ + "Then" + other_cls.__name__
-        transforms = ([objtype, other_cls] if issubclass(other_cls, Transform)
-                      else [objtype] + other_cls._transform_classes)
-        return TransformListMetaclass(name, (TransformationList,), {},
-                                      transforms=transforms)
+    """Python descriptor dispatching `then` on the `Transform` class as either
```
nicolasvasilache, line 14: This file has unfortunately too many trivial spacing changes that makes it hard to review more deeply.

Please note the autoformatting setup that we've been using:
https://github.com/nicolasvasilache/nicolas.vasilache.github.io/blob/master/vscode_settings/settings.json#L43

Maybe that will help you reduce the amount of changes so this load-bearing file becomes reviewable?

nicolasvasilache (commented):
There seems to be a lot of duplication that could be trimmed down, but as per the guidelines it is not a blocker.

Changes to non aALP parts have some issues though.

configure.py
giuseros: This is to support python 3.7. If there is any strong reason to have `argparse.BooleanOptionalAction`, I can revert. But I would say if we can support 3.9 and 3.7 that would be better.

experimental/alp/python/alp/backend/codegen.py
giuseros: This is generic. We basically found out that using different scheduler algorithms gives different performance (and many times the simpler shuffle algorithm works better). There are also some specific settings that apply to our particular case, but they were not generic enough to publish (e.g., disabling some optimizations in LLVM made things work better, somehow). My theory is that an HPC compiler like the one we are doing needs a tailored scheduler (that can spend a bit more time to schedule very large blocks), but many disagree with me. This file is the place where I do these experiments

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
giuseros: We don't use this pass anymore. Basically, using a different scheduling algorithm, this pass can be (fortunately discarded). Once I'll do a bit of c++ refactoring + testing I will remove it

python/examples/core/transform.py
giuseros: Good to know, I was using Black and that file was reformatted like that. I will try your vscode settings. Does it apply the formatting every time you modify a python file?

experimental/alp/python/alp/benchmark/infra.py
giuseros: I will collectively reply here to the comments about the infrastructure. So my idea was to have
*  the compiler (mlirc) to accept a text file (program.mlir) and output an assembly file (program.s)
* the tuner to accept two text files (benchmark.mlir and program.mlir) and output an optimized assembly file (program.s) or the configuration to generate the assembly file
I guess that this is the main difference, right? Your infra is more python-to-python and dynamic. I think we can have a common tool that does both(?). Why the scheduling dialect would be a problem?

python/examples/core/transform.py
giuseros: Also, I never heard of YAPF :) Let me try it out !

configure.py
nicolasvasilache: ok with me but please extract this as a standalone PR that should land separately.

experimental/alp/lib/Transforms/extract_kernel_pass.cpp
nicolasvasilache: Great!

python/examples/core/transform.py
nicolasvasilache: yes, the config I shared reformats on save

experimental/alp/python/alp/benchmark/infra.py
nicolasvasilache: Yes, this makes sense but it needs to be rearchitected a bunch to get it to live side by side in core.
No need to do it now but please consider iterating to a common best possible state.

experimental/alp/include/alp/Transforms/Passes.td
nicolasvasilache: Extracted as a standalone #257 but I was not able to get extra perf from it.
Would be nice to have a use case that exhibits the benefit.
Is there a specific configuration you could share ?

experimental/alp/python/alp/backend/codegen.py
nicolasvasilache: Would be great to factor this in a unified way that we could also experiment with in JIT mode.
Ideally we'd also want to experiment with techniques such as superoptimization but we are far from it atm.

experimental/alp/include/alp/Transforms/Passes.td
giuseros: In some cases it does not make much difference. We found it particularly useful when working with dynamic shapes: the difference is about 30%, so quite an impact. It might also boil down to how llvm applies passes. Maybe for x86 it always converts, while it does not for AArch64?

experimental/alp/python/alp/backend/codegen.py
giuseros: You guys always teach me new terms :) I was studying a lot of this, but I didn't know the technical term was "superoptimization"! Cool! Anyway, any suggestion to how to integrate in the JIT?

configure.py
giuseros: Done! https://github.com/google/iree-llvm-sandbox/pull/258

experimental/alp/python/alp/backend/codegen.py
nicolasvasilache: I don't atm, it may be that a "JIT"-like flow needs to be built by passing files around and explicitly calling llvm, llc etc.
This is not a blocker for this revision to land, just something to think about.

python/examples/core/transform.py
giuseros: I tried YAPF out, but that file still seems to suffer for reformatting. Should I try to add this as an independent change as well?

experimental/alp/lib/Transforms/legalize_vector.cpp
giuseros: I added a test in `experimental/alp/test/test-alp-legalize.mlir`. See if you are able to disable the additional pattern and still don't end up with lots of insert/extract slices, I tried hard but wasn't able to do that

giuseros:
@nicolasvasilache I removed the change to `transform.py` so that we can deal with them in parallel

experimental/alp/lib/Transforms/legalize_vector.cpp
nicolasvasilache: ok, will take a look separately, thanks!

python/examples/core/transform.py
nicolasvasilache: Could you please slice this out in a separate PR, address and rebase ?
This way we can land this bigger refactoring.

experimental/alp/python/alp/benchmark/blas/gemm.py
nicolasvasilache: My tuning experiments have led me to nevergrad, much better !

experimental/alp/lib/Transforms/modulo_scheduling_pass.cpp
nicolasvasilache: Thanks! Added it to my TODO stack.

experimental/alp/include/alp/Transforms/Passes.td
nicolasvasilache: Dynamic is prob the keyword here.
We can experiment more later.

nicolasvasilache:
> @nicolasvasilache I removed the change to `transform.py` so that we can deal with them in parallel

Great, let's just do the same for configure.py and land this.
Alternatively, if you're not getting the same formatting, you can disable it temporarily to avoid all the github-unreviewable whitespaces.
I can format when I land.

giuseros:
@nicolasvasilache , I think there are no more space-formatting changes around. The only change that affects the main repo is in `configure.py`, but it is guarded by an `if args.enable_alp`, so I think it should be ok (please, let me know if it isn't). 

Please, let me know if there is anything else I need to change/update

nicolasvasilache:
>  it is guarded by an if args.enable_alp, so I think it should be ok (please, let me know if it isn't).

Looks good thanks!

PR #249, opened 2022-02-07
Support transpose_padding in transfer dialect TileOp

ftynse:
This has been added after the op was introduced and not reflected in the
op before now.


nicolasvasilache (approved):
Note: you may want to call this a "tiling expert" and impl. it as a sequence of tile pad etc in the scheduling dialect.
This will allow simplifying core which had to do things a certain way under a "no-liener-rewriter" doctrine.

ftynse:
Ack.

PR #251, opened 2022-02-07
Simplify execute_on_lane_0 lowering: One alloc per value.

matthias-springer:
No need to allocate a single buffer and indexing into to it. We can
allocate multiple buffers.


ThomasRaoux (approved):
Nice improvement!

nicolasvasilache (approved):
+1, thanks!

PR #253, opened 2022-02-08
Distribute vector.reduction ops

matthias-springer:
Only vector<32x...> are supported. The reduction is implemented via lane shuffles and element-wise reductions.


matthias-springer:
Only review the top-most commit.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -173,6 +176,119 @@ struct WarpOpElementwise : public OpRewritePattern<WarpSingleLaneOp> {
...
+    Value laneVal =
+        rewriter.create<vector::ExtractOp>(yieldLoc, laneValVec, 0);
+
+    // Parallel reduction: Every thread reduces two values. The result is
+    // stored at the lower thread. Requires log_2(kWarpSize) many parallel
+    // reductions.
+    for (int i = kWarpSize/2; i > 0; i /= 2) {
+      Value offset = rewriter.create<arith::ConstantOp>(
+          reductionOp.getLoc(), rewriter.getI32IntegerAttr(i));
+      Value shuffled = rewriter.create<gpu::ShuffleOp>(
```
nicolasvasilache, line 276: These should have better builders so we don't need to explicitly create useless constants at all call sites.
Could you update and bump the LLVM commit ? 
Ideally they'd also have a static form but that may be too much to ask..

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -173,6 +176,119 @@ struct WarpOpElementwise : public OpRewritePattern<WarpSingleLaneOp> {
...
+/// ...
+/// %r4, %s4 = gpu.shuffle down %e, %c1, %c32 : f32
+/// %a4 = arith.addf %a3, %r4 : f32
+/// %r = gpu.shuffle idx %a4, %c0, %c32 : f32
+/// ```
+struct WarpOpReduction : public OpRewritePattern<WarpSingleLaneOp> {
+  using OpRewritePattern<WarpSingleLaneOp>::OpRewritePattern;
+
+  // Reduce the two given scalars.
+  static Value reduceScalars(OpBuilder &b, Location loc, Value v1, Value v2,
```
nicolasvasilache, line 210: This has a bad smell to it .. can we update core to use CombiningKind and reuse / expose a builder with CombiningKind ?

nicolasvasilache (commented):
Looks great! (modulo the stuff that should be better sliced in core)

PR #256, opened 2022-02-08
Add loop transformations to the transform dialect

ftynse:
Introduce transform dialect operations for loop transformations
currently supported by the transformation strategy. To target
transformation on loops, also introduce an auxiliary operation that
takes a handle to a Linalg operation and produces a handle to
one of its surrounding loops.

### DIFF
--- a/include/Dialects/LinalgTransform/LinalgTransformOps.td
+++ b/include/Dialects/LinalgTransform/LinalgTransformOps.td
@@ -174,6 +174,44 @@ def LowerToLLVMOp : Linalg_Transform_Operation<"lower_to_llvm"> {
   let assemblyFormat = "attr-dict";
 }
 
+def GetParentLoopOp : Linalg_Transform_Operation<"get_parent_loop"> {
+  let description = [{Obtains a handle to a parent loop of the given
+  operation.}];
+
+  let arguments = (ins PDL_Operation:$target,
+                   DefaultValuedAttr<Confined<I64Attr, [IntPositive]>,
+                                     "1">:$num_loops);
+  let results = (outs PDL_Operation:$transformed);
+
+  let assemblyFormat = "$target attr-dict";
+}
+
+def UnrollLoopOp : Linalg_Transform_Operation<"unroll_loop"> {
+  let description = [{Unrolls the given loop with the given unroll factor.}];
+
+  let arguments = (ins PDL_Operation:$target,
+                   Confined<I64Attr, [IntPositive]>:$factor);
+  
+  let assemblyFormat = "$target attr-dict";
+}
+
+def PipelineLoopOp : Linalg_Transform_Operation<"pipeline_loop"> {
+  let arguments = (ins PDL_Operation:$target,
+                   DefaultValuedAttr<I64Attr, "1">:$iteration_interval,
+                   DefaultValuedAttr<I64Attr, "10">:$read_latency);
+  let results = (outs PDL_Operation:$transformed);
+
+  let assemblyFormat = "$target attr-dict";
+}
+
+def OutlineLoopOp : Linalg_Transform_Operation<"outline_loop"> {
+  let arguments = (ins PDL_Operation:$target,
+                   StrAttr:$func_name);
+  let results = (outs PDL_Operation:$transformed);
+
+  let assemblyFormat = "$target attr-dict";
+}
+
 //===----------------------------------------------------------------------===//
 
 def ExpertOp : Linalg_Transform_Operation<"expert"> {
--- a/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
+++ b/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
@@ -18,9 +18,12 @@
 #include "Dialects/LinalgTransform/SimplePatternRewriter.h"
 #include "Dialects/LinalgTransform/TrackingCSE.h"
 #include "Dialects/LinalgTransform/TrackingRewriteDriver.h"
+#include "Dialects/LinalgTransform/TransformOpMapping.h"
 #include "FunctionHelpers.h"
 #include "PDL.h"
+#include "TrackingListener.h"
 #include "Transforms/Functional.h"
+#include "Transforms/Listener.h"
 #include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
 #include "mlir/Conversion/LinalgToLLVM/LinalgToLLVM.h"
 #include "mlir/Conversion/LinalgToStandard/LinalgToStandard.h"
@@ -117,8 +120,10 @@ static FailureOr<LinalgOp> forwardOp(LinalgOp op, PatternRewriter &rewriter) {
 /// and store them in the operation map.
 static LogicalResult executeMatchOp(transform::MatchOp op, ModuleOp module,
                                     TransformOpMapping &operations) {
-  FailureOr<SmallVector<LinalgOp>> ops = findMatchingOps(op, module);
-  if (failed(ops)) return failure();
+  FailureOr<SmallVector<Operation *>> ops = findMatchingOps(op, module);
+  if (failed(ops))
+    return failure();
+  LLVM_DEBUG(DBGS() << "matched " << ops->size() << " ops\n");
   operations.try_emplace(op.target(), std::move(*ops));
   return success();
 }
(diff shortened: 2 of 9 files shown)


include/Dialects/LinalgTransform/LinalgTransformOps.td
```diff
@@ -166,6 +166,44 @@ def LowerToLLVMOp : Linalg_Transform_Operation<"lower_to_llvm"> {
...
+}
+
+def OutlineLoopOp : Linalg_Transform_Operation<"outline_loop"> {
+  let arguments = (ins PDL_Operation:$target,
+                   StrAttr:$func_name);
+  let results = (outs PDL_Operation:$transformed);
+
+  let assemblyFormat = "$target attr-dict";
+}
+
```
giuseros, line 206: Out of curiosity, is there also a ready-made "FuseLoopsOp"? As in:
```
for i = 0:10:
for j=0:10:
```
becomes 
```
for ij= 0:100
```

ftynse: Not at the moment, this isn't something we needed so far, but can be added. The transformation is called loop coalescing btw.

include/Dialects/LinalgTransform/LinalgTransformOps.td
```diff
@@ -166,6 +166,44 @@ def LowerToLLVMOp : Linalg_Transform_Operation<"lower_to_llvm"> {
...
+  let arguments = (ins PDL_Operation:$target,
+                   Confined<I64Attr, [IntPositive]>:$factor);
+  
+  let assemblyFormat = "$target attr-dict";
+}
+
+def PipelineLoopOp : Linalg_Transform_Operation<"pipeline_loop"> {
+  let arguments = (ins PDL_Operation:$target,
+                   DefaultValuedAttr<I64Attr, "1">:$modulo,
```
nicolasvasilache, line 192: This is not up to date anymore, we now have:
```
    Option<"II", "II", "unsigned", /*default=*/"1", "Iteration Interval.">,
    Option<"readLatency", "read-latency", "unsigned", /*default=*/"1",
```

@ThomasRaoux I believe I took the term "Iteration Interval" from your terminology.
I assume it is the same thing as the traditional "initiation interval" ?

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -514,18 +641,13 @@ static void noOpRewriter(ArrayRef<PDLValue> args, ArrayAttr constantParams,
...
-/// may be either an SSA value that corresponds to a list of ops produced by
-/// other transforms and stored in `operations` or a list of ops that match the
-/// PDL pattern specified by its name.
 template <typename OpTy>
-static FailureOr<SmallVector<LinalgOp>>
-findTransformTarget(OpTy transformOp, ModuleOp module,
-                    const DenseMap<Value, SmallVector<LinalgOp>> &operations) {
-  if (Value target = transformOp.target())
-    return operations.lookup(target);
+using has_target_matcher_t = decltype(std::declval<OpTy &>().targetMatcher());
```
nicolasvasilache, line 645: Should this be `target_matcher_t` ?
The `has_` prefix suggests a boolean type but I don't see it evaluating to a bool ..

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -546,26 +668,73 @@ findTransformTarget(OpTy transformOp, ModuleOp module,
...
 }
 
+template <typename OpTy>
+static std::enable_if_t<!llvm::is_detected<has_target_matcher_t, OpTy>::value,
+                FailureOr<SmallVector<Operation *>>>
+matchUsingPDL(OpTy transformOp, ModuleOp module) {
+  assert(false && "matching with PDL on an unsupported op");
+}
+
+/// Returns the operations that `transformOp` should be run on. The
```
nicolasvasilache, line 678: This sfinae genuflexion would go away once #218 lands right ?

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -546,26 +668,73 @@ findTransformTarget(OpTy transformOp, ModuleOp module,
...
+static void removeCurrentTarget(ConfigOpTy configOp,
+                                TransformOpMapping &operations) {
+  // Since there is only allowed use of the value in the transformation dialect,
+  // we can remove it from the mapping after processing its only user. This
+  // ensures we don't accidentally keep pointers to operations that may have
+  // been deleted by the current transformation.
+  if (!configOp.target())
+    return;
+  Value target = configOp.target();
+  assert(target.hasOneUse());
```
nicolasvasilache, line 703: The assert is already present in the `erase` operation.
Consider adding a more meaningful error message ?

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -546,26 +668,73 @@ findTransformTarget(OpTy transformOp, ModuleOp module,
...
+}
+
+/// Returns the operations that `transformOp` should be run on. The
+/// `transformOp` is expected to have the TargetableTransformOpTrait. Its target
+/// may be either an SSA value that corresponds to a list of ops produced by
+/// other transforms and stored in `operations` or a list of ops that match the
+/// PDL pattern specified by its name.
+template <typename OpTy>
+static FailureOr<SmallVector<Operation *>>
+findTransformTarget(OpTy transformOp, ModuleOp module,
```
nicolasvasilache, line 685: should this be using an interface and drop the templating?

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -546,26 +668,73 @@ findTransformTarget(OpTy transformOp, ModuleOp module,
...
-executeTransformOnEach(ModuleOp module, OpTy configOp, FnTy transform,
-                       DenseMap<Value, SmallVector<LinalgOp>> &operations) {
-  FailureOr<SmallVector<LinalgOp>> targets =
+template <typename ConfigOpTy, typename FnTy,
+          typename TransformedOpType =
+              typename llvm::function_traits<FnTy>::template arg_t<0>,
+          typename ReturnType = typename llvm::function_traits<FnTy>::result_t>
+static std::enable_if_t<!std::is_same<ReturnType, LogicalResult>::value,
+                        LogicalResult>
+executeTransformOnEach(ModuleOp module, ConfigOpTy configOp, FnTy transform,
```
nicolasvasilache, line 715: What is the value of the sfinae genuflexion here?
Can we have 2 functions with different names instead ?

include/Dialects/LinalgTransform/LinalgTransformOps.td
ftynse: Thanks. I would strongly suggest to give things more meaningful names than "II".

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: The `_t` suffix combined with the STL naming scheme indicates that name corresponds to a type. This is the normal idiom for `llvm::is_detected` (https://llvm.org/doxygen/namespacellvm.html#a5222d74738fef771d06ccec12e74cf97), which is used just below.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: Most of it, not all.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: This can be detemplatized with #218 AFAICS.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: You are probably confusing this with erasing the operation itself (`Operation::erase`), rather than removing the value from the dense map called `operations`, that does indeed contain an assertion but for it having _zero_ uses, not one. This assert reflects what we have started to enforce in the verifier recently.

include/Dialects/LinalgTransform/LinalgTransformOps.td
nicolasvasilache: `iteration_interval` seems fine here, the old one can then go away.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
nicolasvasilache: I would have expected `has_xxx_t` to correspond to either `true_type` or `false_type` from type_traits (see e.g.  https://en.cppreference.com/w/cpp/header/type_traits).

The example you post says: `has_copy_assign_t = decltype(std::declval<T&>() = std::declval<const T&>()); `.
I would think that this evals to true_type / false_type, though didn't check and not 100% sure as I don't know what `std::declval<T&>() = std::declval<const T&>()` evaluates to.

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
nicolasvasilache: great!

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
nicolasvasilache: ah yes .. I read this as `builder.eraseOp` for some reason.
Please disregard.

nicolasvasilache (approved):

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
ftynse: No, that does not evaluate to `std::true/false_type`, unless the copy-assignment of T returns an instance of that type. (Technically speaking, it does not evaluate to anything at all since decltype is an unevaluated context.) `std::declval<T&>() = std::declval<const T&>() ` is equivalent to `std::declval<T &>().operator =(std::declval<const T&>())`, so this is an expression that calls the copy-assignment operator. Putting this in `decltype` specifies the type that the expression would have had when evaluated. In this case, it is the return type of `T::operator =(const T&)`. In most cases, it is just `T &` to allow for chained assignments, but that is actually irrelevant. The important part is the expression inside `decltype` being well-formed, i.e. the copy-assignment operator existing on `T` in the example case. The type alias defined with this `decltype` is used a s template argument for `llvm::is_detected` so if the expression inside `decltype` is ill-formed, template argument substitution fails in the specialized version (that aliases `is_detected` to `true_type`), the non-specialized version (that aliases `is_detected` to `false_type`) can be used since substitution failure is not an error.

nicolasvasilache: I see, thanks for spelling it out.
So the true/false interpretation is actually in TM^2P land: error or not and this gets resolved by SFINAE .. gah

include/Dialects/LinalgTransform/LinalgTransformOps.td
ThomasRaoux: initiation interval is probably the right type but `iteration_interval` is fine as well.

PR #259, opened 2022-02-09
Fix build and bump LLVM version

matthias-springer:

PR #260, opened 2022-02-09
Fix remaining uses of -convert-scf-to-std

shabalind:


ftynse:
Thanks!

ftynse (approved):

shabalind:
Rebased on top of #259.

PR #265, opened 2022-02-10
Target conv 2d vectorization at specific ops

ftynse:
Vectorization for 2d convolutions has not been filtered for specific
ops, presumably to avoid exposing the name of the op after
decomposition. Target it at decomposed 1d convolutions instead in
preparation for a better transfromation targeting scheme.

Enable DecomposeOp transformation in the transform dialect.


nicolasvasilache (approved):
LGTM

PR #267, opened 2022-02-11
Add possibility to save IR in transform.py

giuseros:
cc @nicolasvasilache

### DIFF
--- a/python/examples/core/transform.py
+++ b/python/examples/core/transform.py
@@ -134,6 +134,35 @@ class PrintIR(Transform):
     return module
 
 
+class SaveIR(Transform):
+  """Save intermediate IR.
+
+    Save the module and do not change it. The transform can be configured as
+    follows:
+    * `file_name`: Base path to use to save the intermediate IRs.
+    * `id`: Suffix identifier to differentiate intermediate transforms
+    """
+
+  def __init__(self, file_name, t, id=0, **kwargs):
+    assert hasattr(t, "pipeline") or hasattr(t, "pipelines"), "missing attr"
+    self.pipelines = [t.pipeline] if hasattr(t, "pipeline") else t.pipelines
+    self.file_name = file_name
+    self.transform = t
+    self.id = id
+
+  def __call__(self, module: Module, fun_name: str):
+    base = os.path.splitext(self.file_name)[0]
+    transform_name = type(self.transform).__name__
+    base_transform = f"{base}.{self.id}" if self.id else f"{base}"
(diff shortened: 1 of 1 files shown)


python/examples/core/transform.py
```diff
@@ -209,6 +239,25 @@ def print_ir(self,
...
+  def save_ir(self,
+              file_name,
+              after_all: bool = False,
+              llvm: bool = False) -> TransformationList:
+    """Return a new transformation list that save the IR at the given points."""
+    transforms = []
+    id = 0
+    for t in self.transforms:
+      if (len(transforms) >= 2 and
+          type(transforms[-2]).__name__ == type(t).__name__):
```
nicolasvasilache, line 251: Do you see value in this type of granularity?
I would have thought that set by step saving and incrementing would be simpler and nicer.

I.e. atm you'd get something like:
```
foo.transA.0
foo.transA.1
foo.transB.0
foo.transB.1
```
which does not tell you how A and B are interleaved.

How about 
```
foo.1.transA.
foo.2.transB
foo.3.transA
foo.4.transB
```

instead ?

python/examples/core/transform.py
```diff
@@ -134,6 +134,36 @@ def __call__(self, module: Module, fun_name: str):
     return module
 
 
+class SaveIR(Transform):
```
nicolasvasilache, line 137: Note that the codegen strategy pipeline will soon be deprecated in favor of the scheduling dialect.
@ftynse may prefer you hold off on this?

giuseros: Well if it is really a problem I can try to move it in the experimental/ folder as well. Please, let me know!

nicolasvasilache: I think it is really about reducing extra work for @ftynse he should say what he prefers.

ftynse (commented):
The addition is not problematic for me in itself.

python/examples/core/transform.py
nicolasvasilache: coming back to this, I don't think this has been addressed ?

nicolasvasilache:
> The addition is not problematic for me in itself.

Ok then let's address the outstanding comment and land this please!

python/examples/core/transform.py
giuseros: No, sorry, I didn't notice @ftynse comment. I am going to push it now

giuseros: Done :)

nicolasvasilache (approved):

PR #268, opened 2022-02-13
Add correctness check to fusion tests and make experts identical.

hanhanW:


hanhanW:
@gysit @MaheshRavishankar could you help review the PR? I don't have permissions to add you to reviewers.

python/examples/fusion/test.py
```diff
@@ -39,47 +43,42 @@ def fill_matmul_fusion():
...
+              'linalg.generic',
+              pad=True,
+          )).then(
+              Tile(
+                  fun_name,
+                  'linalg.matmul',
+                  tile_sizes=[0, 0, 24],
+                  pad=True,
+                  pack_paddings=[1, 1, 0],
+              )).then(Vectorize(fun_name, '', vectorize_paddings=True)).then(
```
gysit, line 70: I like the approach! It makes sure padding happens in the order we want!

python/examples/fusion/definitions.py
```diff
@@ -165,7 +165,60 @@ def shapes_builder(self, sizes: Mapping[str, Any]) -> List[List[int]]:
...
+    return tensors
+
+  def check_np(self, A: np.dtype, B: np.dtype, C: np.dtype, D: np.dtype,
+               E: np.dtype) -> None:
+    """NumPy checking function.
+
+    Given a list of NumPy values, check the precomputed results matches those of
+    the expected reference implementation.
+    """
+    res = np.dot(A, B) + C
```
gysit, line 217: This is assuming the bias is set to zero right? Using a non zero bias may make sense at some point.

gysit (approved):
Thanks!

python/examples/fusion/definitions.py
hanhanW: No, I don't think so. I think the bias values are in `C`?

gysit: OK, I thought the bias is in the 1D array which supposedly is the tensor argument D of check_np?

hanhanW: It is in `C` which is a 1 D array.

```
    M, N, K = sizes["M"], sizes["N"], sizes["K"]
    return [
        [M, K],
        [K, N],
        [N],
        [M, N],
        [M, N],
    ]
```

`D` is sort of useless. The result of matmul is stored to `D` IIRC.

gysit: Ok makes sense! Thanks for explaining

PR #269, opened 2022-02-14
Hook up a minimal register tiling example to nevergrad.

nicolasvasilache:
This is expected to need significant evolution but for now this is enough to get off the ground.

Repro:
```
python -m python.examples.tuning.test_nevergrad_small_matmul_noisy --dynamic_at_compile_time_list '[]' --problem_sizes_list 60,60,60 --search-budget 100 --n_iters 100 --num-compilation-processes 30 --search-strategy NGOpt
```

### DIFF
--- a/python/examples/tuning/test_nevergrad_small_matmul_noisy.py
+++ b/python/examples/tuning/test_nevergrad_small_matmul_noisy.py
@@ -1,0 +1,278 @@
+from argparse import ArgumentParser
+import multiprocessing as mp
+import nevergrad as ng
+import numpy as np
+
+import mlir.iree_sandbox as sandbox
+import mlir.ir as ir
+import mlir.dialects.pdl as pdl
+import mlir.dialects.linalg_transform as transform
+
+from ..core.experts import *
+from ..core.harness import *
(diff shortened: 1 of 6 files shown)


python/examples/tuning/test_nevergrad_small_matmul_noisy.py
```diff
@@ -0,0 +1,285 @@
...
+    print(f'Register tile sizes: {search_sizes}')
+
+    # TODO: this is necessary to force-load the dialect, otherwise op creation
+    # complains about "unregistered dialect" despite the registration called.
+    register_sandbox_passes_and_dialects(module.context)
+    module.context.dialects["linalg_transform"]
+
+    nnz = np.count_nonzero(search_sizes)
+    peel = [x for x in range(nnz)]
+    peel = []  # TODO: enable peeling once handles are fixed.
```
ftynse, line 103: Could you elaborate what's the issue with handles that is blocking here?

nicolasvasilache: I experienced crashes and did not investigate further, will give it another show once this is refactored.

PR #270, opened 2022-02-14
Add an untargeted version of the vectorization op

ftynse:
In some cases, a global vectorization is desirable to ensure a proper
composition of vectorization and canonicalization patterns in the
process that is not achievable with the scoped targeted vectorization.
Introduce an untargeted version of the vectorization op in addition to
the targeted one to support such cases.

### DIFF
--- a/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
+++ b/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
@@ -100,6 +100,100 @@ static Value getNeutralOfLinalgOp(OpBuilder &b, OpOperand &op) {
                                      b.getZeroAttr(t));
 }
 
+template <typename ConfigOpTy>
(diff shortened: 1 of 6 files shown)


lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -100,6 +100,100 @@ static Value getNeutralOfLinalgOp(OpBuilder &b, OpOperand &op) {
                                      b.getZeroAttr(t));
 }
 
+template <typename ConfigOpTy>
+static void removeCurrentTarget(ConfigOpTy configOp,
+                                TransformOpMapping &operations) {
+  // Since there is only allowed use of the value in the transformation dialect,
```
nicolasvasilache, line 106: typo: only "one" allowed

nicolasvasilache (approved):

PR #276, opened 2022-02-14
Mark pointers as noalias when lowering to LLVM

ftynse:
This matches the ugly hack from the codegen strategy and makes the
generated code to perform on par with it for convolutions.

### DIFF
--- a/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
+++ b/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
@@ -559,7 +559,19 @@ executeLowerToLLVMOp(ModuleOp module,
   pm.addPass(createMemRefToLLVMPass());
   pm.addPass(createLowerToLLVMPass());
   pm.addPass(createReconcileUnrealizedCastsPass());
-  return pm.run(module);
+  if (failed(pm.run(module)))
+    return failure();
+
+  // Make all arguments noalias for now.
+  // FIXME: this is a terrible hack!
+  module.walk([](LLVM::LLVMFuncOp funcOp) {
+    for (int64_t i = 0; i < funcOp.getNumArguments(); ++i) {
+      if (!funcOp.getType().getParamType(i).isa<LLVM::LLVMPointerType>())
+        continue;
+      funcOp.setArgAttr(i, "llvm.noalias", UnitAttr::get(funcOp.getContext()));
+    }
+  });
+  return success();
 }
 
 static FailureOr<scf::ForOp>


lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
```diff
@@ -559,7 +559,19 @@ executeLowerToLLVMOp(ModuleOp module,
   pm.addPass(createMemRefToLLVMPass());
   pm.addPass(createLowerToLLVMPass());
   pm.addPass(createReconcileUnrealizedCastsPass());
-  return pm.run(module);
+  if (failed(pm.run(module)))
+    return failure();
+
+  // Make all arguments noalias for now.
+  // FIXME: this is a terrible hack!
```
nicolasvasilache, line 566: Why not make it a proper transformation on the funcOp ?
The transform is then responsible for turning it on.

ftynse: What do you mean by "proper transformation" here? We can make it a option in the lowering. To make it a transform dialect op, we would need a handle to the function.

In any case, this remains a terrible hack because it puts `noalias` on pointers that most certainly do alias (allocated and aligned). It just gets away with it because the allocated pointer is never read in the function.

nicolasvasilache: Right, so in practice we have some weak guarantees because the base pointer is only used for alloc/dealloc but this is obviously far from good. 
Do you see another solution than implementing alias sets?
If we really must prioritize this we could but experience has shown the cost/benefit tradeoff is not worth it atm.

Additionally, let's put this into the larger context, as @MaheshRavishankar is working on making it easier to plug custom codegen into IREE, a lot of these crutches will plainly be deleted.

nicolasvasilache (approved):

lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp
MaheshRavishankar: Handling no-alias is tricky in general (the LLVM model is just plain not good enough). For context, in IREE we dont have function arguments. ABI is represented differently. So we need something more involved to be able to get the alias sets that LLVM needs. Here, marking all arguments as `no_alias` makes sense to me. 

Have looked into this a bit more within IREE. Alias and alignment are both sort of related. In LLVM both the alignment information and the alias information needs to be annotated on loads/stores (assuming you dont do this on function arguments). This means we need a way in MLIR to annotate this. My current thought here is to have a transformation in MLIR that linearizes all memref accesses, i.e. a conversion pass that converts all `memref`s into 1D memrefs. This has been something I'd like to have for a long time. Apart from reducing the complexity of lowering to MLIR dialect, it will also get the IR in the form that allows us to annotate loads/stores with alignment information appropriately. Attaching noalias information on load/stores would be a trivial step at that time.

That was a bit of detour, but as Nicolas said none of this has gotten to "we need to do this", (as opposed to "we should eventually get to this").

PR #278, opened 2022-02-16
Expand the nevergrad search with a few more parameters.

nicolasvasilache:
First example of non-trivial schedule found:

```
Best solution: 125 GUnits/s (peak is 192 GUnits/s)
Problem sizes: [60, 60, 60]
Register tile sizes: [12, 16, 4]
Register interchange: [1, 2, 0]
Tile partial by one: [False, True, True]
```


nicolasvasilache:
```
Best solution: 154 GUnits/s (peak is 192 GUnits/s)
Problem sizes: [60, 60, 60]
Register tile sizes: [6, 16, 1]
Register interchange: [0, 2, 1]
Tile partial by one: [True, False, False]
```

PR #280, opened 2022-02-17
Bump llvm to d1e3235f604d65a62d25842305f54e43bd36681f

shabalind:


python/examples/core/compilation.py
```diff
@@ -137,8 +137,10 @@ def emit_benchmarking_function(name: str,
     loop = scf.ForOp(zero, n_iterations, one, iter_args)
     with InsertionPoint(loop.body):
       start = std.CallOp(nano_time, [])
-      call = std.CallOp(
-          func, wrapper.arguments[:-num_results - 1] + loop.inner_iter_args)
```
shabalind, line 141: @ftynse " + " here was returning wrong results and failing the test. I've replaced the code with a workaround but it might be a bug in python bindings upstream.

python/examples/core/compilation.py
```diff
@@ -137,8 +137,10 @@ def emit_benchmarking_function(name: str,
...
+      print(f"loop inner iter args: {loop.inner_iter_args[0]}")
```
ftynse, line 140: Leftover debug spew

shabalind: Good catch, I'll drop it once the CI passes since it takes a long while.

python/examples/core/compilation.py
ftynse: What kind of failures?

shabalind: This function call accepts arguments from the outer wrapper arguments + one argument through loop carried arguments. The expected result of the `+` here would be `a, b, l` where `a, b` are outer arguments and `l` is loop carried. But after the bump of llvm version it turned into `a, b, a`.

PR #284, opened 2022-02-18
Print some timings

nicolasvasilache:


nicolasvasilache:
@matthias-springer FYI I added some timings here.

Just run (and redirect to /tmp/foo) something like:
```
python -m python.examples.tuning.test_nevergrad_small_matmul_noisy --search-budget 1000 --n_iters 100 --num-compilation-processes 32
```

The just grep through the file to see the timings printed.
Basically on this example, the main dispatching process is much too slow and can only sustain ~10-20 processes.

This is probably related to the current naive use of multiprocessing and could be solved by moving to mp.Pool as you suggested. Probably avoiding to stupidly fork and terminate new processes every time would help a bunch.

Side note, we don't want to move to multi-thread, we need a MP solution.

matthias-springer:
How did you conclude that the main process is too slow? I measured the running time of the above command with different process numbers.

* 1 process: 2218 s
* 18 processes: 141s
* 24 processes: 107s, 119s (tried twice to check variance)
* 31 processes: 106s, 110s
* 64 processes: 99s
* 71 processes: 101s

My machine has 72 cores, but half of them are hyper-threading, so running more than 35 maybe 36 computations does not give much speedup. So we're definitely hitting some kind of bottleneck once we reach ~20 processes.

`mp.Pool` seemed like a good alternative at first but turns out it cannot terminate processes that timed out.

nicolasvasilache:
First I start with the cpuset setup at the end of the README.md, I get a cpuset of 32 cores and migrate all tasks out of that cpuset.

The I get htop in low latency refresh mode (e.g. htop -d 0.1 I believe).
Note that the python script uses taskset to pin each process to each core.

Looking at the processor load, I can clearly see waves of activity from 0 to 32 that indicate work is properly scheduled in a round-robin fashion but that the compile + run cycle is on average much faster that 32 * dispatches.

I confirm that by adding the print and verifying that the dispatch, tell and synchronize times are much higher than I'd like.
The compile and run times are comparatively faster.

Basically the round-robin pipeline of tasks is not filled quickly enough by the main process.

nicolasvasilache:
> mp.Pool seemed like a good alternative at first but turns out it cannot terminate processes that timed out.

If mp.Pool is indeed more efficient at start/join without terminating processes, we can move the timeout logic.

One easy way would be to have another fork inside [compile_and_run_checked_mp](https://github.com/google/iree-llvm-sandbox/blob/main/python/examples/core/nevergrad_parallel_utils.py#L53:5) which distributes the fork/busy-wait/terminate logic to each child.

A more lightweight way could be to use a SIGALARM signal https://stackoverflow.com/questions/492519/timeout-on-a-function-call.

nicolasvasilache:
This has served its purpose.

PR #286, opened 2022-02-21
README: Fix path to test case in mlir-proto-opt example command.

ingomueller-net:


nicolasvasilache:
Thanks @ingomueller-net

PR #287, opened 2022-02-21
Add RW lock support.

nicolasvasilache:
This allows decoupling the taskset for compilation (even supporting multiple compilation processes per cpu),
from the evaluation process which is now better isolated.
This allows setting up tasksets that let us run on multiple cores in each evaluation.
This is a prerequisite for running in parallel without prohibitive interferences.

### DIFF
--- a/python/examples/core/nevergrad_parallel_utils.py
+++ b/python/examples/core/nevergrad_parallel_utils.py
@@ -1,10 +1,15 @@
 from contextlib import redirect_stdout, redirect_stderr
 import io
-import numpy as np
+import math
 import multiprocessing as mp
+import numpy as np
 import os
+from prwlock import RWLock
 import signal
 import sys
+import threading
+import time
+import traceback
 import typing as tp
 
 from ..core.harness import *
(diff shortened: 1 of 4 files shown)


python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -39,71 +40,219 @@ def __init__(self, proposal, problem_instance):
...
+      for process_idx in processes_joined:
+        throughput = tell_joined_process(self.evaluation_tasks, process_idx,
+                                         scheduler, optimizer, throughputs,
+                                         parsed_args)
+        throughput = int(throughput)
+        best = throughput if throughput > best else best
+        self.evaluation_tasks[process_idx] = None
+
+    # We are sure there is at least one empty slot.
+    return self.evaluation_tasks.index(None), best
```
matthias-springer, line 142: note: This will get easier with `pool`.

python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -39,71 +40,219 @@ def __init__(self, proposal, problem_instance):
...
+      lock.release()
+      if t.is_alive():
+        os.system('kill %d' % os.getpid())
+        return
 
     # TODO: redirect to a file if we want this information.
     f.flush()
 
-    ipc_dict['result'] = IPCState(success=True, throughputs=throughputs)
+    ipc_dict['result'] = IPCState(success=True,
```
matthias-springer, line 242: I think this must also be set in case of a failure? Before killing the process

matthias-springer (approved):

python/examples/core/nevergrad_parallel_utils.py
nicolasvasilache: NGMPEvaluation checks whether result is set.
Doing it here will be tricky because the process is not joined and information does not pass easily.

python/examples/core/nevergrad_parallel_utils.py
nicolasvasilache: Cool

PR #288, opened 2022-02-21
Remove optimizer from NGScheduler

matthias-springer:
This was the cause for slow process creation, which was the reason for the master node bottleneck.

The size of optimizer grows with every sample and we were copying the optimizer state to each process.


nicolasvasilache (approved):
Great, thanks for catching this!

PR #289, opened 2022-02-22
Introduce TransformOpInterface

ftynse:
This interface generalizes the behavior of ops from the LinalgTransform
dialect. These ops now implement the interface instead of having the
interpreter switch based on the op kind. The interpreter is
significantly reduced and becomes extensible, supporting transformations
defined by other dialects than LinalgTransform.

This requires quite a bit of code motion in the repository, and some
further cleanups will be performed by separate commits.


include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -0,0 +1,226 @@
...
+/// transformation to a single operation handle and producing a single operation
+/// handle. The op must implement:
+///   - FailureOr<convertible-to-Operation*> applyToOne(OpTy) - a transformation
+///   that is applied in turn to all payload IR operations that correspond to
+///   the handle of the transform IR operation.
+/// In the functions above, OpTy is either Operation * or a concrete payload IR
+/// Op class that the transformation is applied to (NOT the class of the
+/// transform IR op).
+template <typename OpTy>
+class TargetableOpTrait : public OpTrait::TraitBase<OpTy, TargetableOpTrait> {
```
nicolasvasilache, line 146: nit: should `SingleOperandSingleResult` or something similar appear in the trait name ?

include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -0,0 +1,226 @@
...
+/// transformation to a single operation handle and not producing any handles.
+/// The op must implement:
+///   - LogicalResult applyToOne(OpTy) - a transformation that is applied in
+///   turn to all payload IR operations that correspond to the handle of the
+///   transform IR operation.
+/// In the functions above, OpTy is either Operation * or a concrete payload IR
+/// Op class that the transformation is applied to (NOT the class of the
+/// transform IR op).
+template <typename OpTy>
+class TargetableNoResultOpTrait
```
nicolasvasilache, line 186: nit: should `SingleOperandZeroResult` or something similar appear in the trait name ?

include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -0,0 +1,226 @@
...
+  /// Applies the transformation to each op from the only target.
+  LogicalResult apply(TransformResults &transformResults,
+                      TransformState &state) {
+    using TransformOpType = typename llvm::function_traits<
+        decltype(&OpTy::applyToOne)>::template arg_t<0>;
+    ArrayRef<Operation *> targets =
+        state.getTargets(this->getOperation()->getOperand(0));
+    SmallVector<Operation *> results;
+
+    for (Operation *target : targets) {
```
nicolasvasilache, line 198: Looking a bit above, it isn't clear to me that the duplication is needed.
Would it be too convoluted to have only 1 trait a and conditional based on:
```
OpTy::template hasTrait<OpTrait::OneResult>
```
to trigger the `push_back`.

I get the impl of the for loop pushes back in the 1 result case, it still seems like a lot of duplication for a relatively simple switch.

lib/Dialects/LinalgTransform/Transforms/TrackingCSE.cpp
```diff
@@ -7,7 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "Dialects/LinalgTransform/TrackingCSE.h"
-#include "TrackingListener.h"
+#include "../IR/TrackingListener.h"
```
nicolasvasilache, line 10: This feels annoying, any include_dirs you can add to the cmake?

lib/Dialects/LinalgTransform/IR/LinalgTransformOps.cpp
```diff
@@ -103,5 +388,339 @@ void transform::VectorizeOp::print(OpAsmPrinter &printer) {
...
+  return scf::ForOp();
+}
+
+//===---------------------------------------------------------------------===//
+// OutlineLoopOp
+//===---------------------------------------------------------------------===//
+
+static scf::ExecuteRegionOp outlineInExecuteRegion(RewriterBase &b,
+                                                   Operation *op) {
+  op->dump();
```
nicolasvasilache, line 674: spew

lib/Dialects/LinalgTransform/IR/LinalgTransformOps.cpp
```diff
@@ -7,17 +7,51 @@
 //===----------------------------------------------------------------------===//
 
 #include "Dialects/LinalgTransform/LinalgTransformOps.h"
+#include "Dialects/LinalgTransform/ScopedTransform.h"
```
nicolasvasilache, line 10: Note: I haven't reviewed this file yet as changes seem to be mostly moves.
Since github does not have the nice phab-like "code was moved" highlighting, I am skipping for now.

Please lmk if this is fundamentally more than a move and I'll look deeper.

include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -0,0 +1,226 @@
...
+/// values defined by these ops in the transform IR (the part of the IR
+/// containing the ops that implement TransformOpInterface and guide the
+/// transformation) and payload IR (the part of the IR that is being
+/// transformed) ops. Assumes that each value in the transform IR can be used at
+/// most once (since transformations are likely to change the payload IR ops the
+/// value corresponds to). Checks that transform IR values correspond to
+/// disjoint sets of payload IR ops throughout the transformation.
+class TransformState {
+public:
+  /// Creates a state for the transformation rooted at the given op.
```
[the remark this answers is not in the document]
ftynse: Doesn't seem to be the case in MLIR...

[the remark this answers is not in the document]
nicolasvasilache: depends where you look, that's why I use "trying"; e.g. randomly picking on stuff:

mlir/Rewrite/FrozenRewritePatternSet.h does
mlir/Dialect/Vector/Transforms/VectorTransforms.h partially does
mlir/Dialect/Linalg/Transforms/Transforms.h does
mlir/Dialect/Affine/LoopFusionUtils.h is 50-50 ...

My recollection is this was an actual debate with somewhat strong opinions involved.

include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -0,0 +1,226 @@
...
+    unsigned length;
+  };
+  SmallVector<Range, 2> segments;
+  SmallVector<Operation *> operations;
+};
+
+/// Applies a one-to-one transform to each of the given targets. Puts the
+/// results of transforms in `results` in the same order. Fails if any of the
+/// application fails. Individual transforms must be callable with signature
+///   FailureOr<convertible-to-Operation*>(OpTy)
```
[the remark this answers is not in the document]
ftynse: Not really. Default template deduction rules are not strong enough to deduce `ReturnTy` and `ArgTy` when passing a lambda to the following.

```
template <typename ReturnTy, typename ArgTy>
void foo(function_ref<ReturnTy (ArgTy)>)
```

One would need to explicitly construct a `function_ref` object before calling the function. C++17 deduction guides might help, but we are targeting LLVM with C++14. MLIR has many similar APIs, type conversion for example. I added static asserts though.

include/Dialects/LinalgTransform/TransformOpInterface.h
ftynse: You literally asked me to make the non-trait version of this use less templates, but sure.

include/Dialects/LinalgTransform/TransformOpInterface.h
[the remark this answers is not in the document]
ftynse: It never made it to the coding guidelines so I suppose there is no limitation one way or another. I usually avoid nitpicking on these in reviews to avoid resuscitating that bikeshed discussion. If you do feel strongly about these, I can change the comments.

lib/Dialects/LinalgTransform/Transforms/TrackingCSE.cpp
ftynse: I can just move the header.

include/Dialects/LinalgTransform/TransformOpInterface.h
ftynse: My template-fu is stronger :)

nicolasvasilache (approved):

PR #291, opened 2022-02-22
Fix running bench targets.

hanhanW:
The order was broken because of https://github.com/google/iree-llvm-sandbox/commit/72730fc5b8f41b6faf50ae1fb733838802c9cd1c
The PR passes arguments using keyword arguments to fix and prevent this
issue in the future.


nicolasvasilache (approved):
Thanks @hanhanW !

PR #293, opened 2022-02-23
Introduce an extension mechanism for TransformState

ftynse:
This mechanism allows one to attach arbitrary typed extension objects to
the TransformState object, shared by ops implementing the
TransformOpInterface. Thus transform ops may communicate more
information than available in the state.

Use this mechanism instead of manually creating tracking listeners in
the Linalg Transform dialect interpreter. The current use is slightly
awkward as it requires to add and remove an extension in order to ensure
it reflects the most recent state of the payload IR, but it already
avoids reconstructing the internal structure of the tracking listener
for every canonicalization/CSE/enabler transformation that run
consecutively. Further work should remove the need to remove the
extension by notifying it about changes to the state.

### DIFF
--- a/include/Dialects/LinalgTransform/TransformOpInterface.h
+++ b/include/Dialects/LinalgTransform/TransformOpInterface.h
@@ -46,9 +46,58 @@ public:
   /// the state accordingly.
   LogicalResult applyTransform(TransformOpInterface transform);
 
-  // FIXME: we shouldn't be exposing this as it allows for modification of the
-  // internal mapping state.
-  TransformOpMapping &getMapping() { return operations; }
+  /// The extension mechanism for TransformState. Extensions are expected to
+  /// derive this class and may use its methods to access the state. Extensions
+  /// are identified by their type and a state can only have one extension of
+  /// a particular type.
+  class Extension {
+  public:
+    // Out-of-line implementation to ensure vtable and metadata are emitted in
+    // a single .o file.
+    virtual ~Extension();
+
+  protected:
+    // FIXME: this provides direct write access to the state internals, replace
+    // with more transactional methods.
+    TransformOpMapping &getMapping(TransformState &state) {
+      return state.operations;
+    }
+  };
+
+  /// Adds a new extension of the type specifeid as template parameter,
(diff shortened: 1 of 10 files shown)


gysit (approved):

include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -46,9 +46,50 @@ class TransformState {
...
+      return state.operations;
+    }
+  };
+
+  /// Adds a new extension of the type specifeid as template parameter,
+  /// constructing it with the arguments provided. The extension is owned by the
+  /// TransformState. It is expected that the state does not already have an
+  /// extension of the same type.
+  template <typename Ty, typename... Args>
+  std::enable_if_t<std::is_base_of<Extension, Ty>::value, Ty &>
```
nicolasvasilache, line 72: why do you need SFINAE here ?

include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -64,6 +105,10 @@ class TransformState {
 
   /// The mapping between payload IR values and transform IR ops.
   TransformOpMapping operations;
+
+  /// Extensions attached to the TransformState, identified by the TypeID of
+  /// their type. Only one extension of any given type is allowed.
+  DenseMap<TypeID, std::unique_ptr<Extension>> extensions;
```
nicolasvasilache, line 111: I do not understand the need for this complexity, AFAIU `TypeID` is always `TypeID::get<TrackingState>`.

Is this premature generalization?

include/Dialects/LinalgTransform/TransformOpInterface.h
ftynse: replaced with static_assert

include/Dialects/LinalgTransform/TransformOpInterface.h
ftynse: The same as for having all the `functional::` stuff: separation of concerns and preparation for upstreaming.

PR #295, opened 2022-02-24
Add default attribute values to fix conv benchmarks.

hanhanW:
We're not able to run benchmarks if the default values are not added.
It's one of opdsl requrements.


gysit (approved):
Thanks a lot for fixing! I missed this when updating OpDSL.

gysit (approved):
Thanks a lot for fixing this!

PR #296, opened 2022-02-24
Use a multiprocessing.Pool to manage compilation+benchmark processes

matthias-springer:
This gives a 33% speedup on my machine when running with 35 cores. Also
simplifies the code.

### DIFF
--- a/python/examples/core/nevergrad_parallel_utils.py
+++ b/python/examples/core/nevergrad_parallel_utils.py
@@ -17,166 +17,73 @@ from ..core.nevergrad_tuner_utils import NGSchedulerInterface
 from ..core.problem_definition import ProblemDefinition
 from ..core.utils import compute_quantiles
 
+class SearchJobResult():
+  """The result of a search job."""
 
-class IPCState:
-  """State shared across parent and children processes in a MP run.
-  
-  All contained information must "pickle" (i.e. serialize across processes).
+  def __init__(self, proposal, throughputs):
+    self.proposal = proposal
+    self.throughputs = throughputs
+
+class ProcessState():
+  """State that is shared among all jobs that are scheduled to run on a pool
+  process.
   """
 
   def __init__(self,
-               success: bool,
-               throughputs: tp.Sequence[float],
-               problem=None):
-    self.success = success
-    self.throughputs = throughputs
-    self.problem = problem
+               available_cpus_queue,
+               parsed_args,
+               results_queue,
+               rw_locks,
+               shutdown_event):
+    self.available_cpus_queue = available_cpus_queue
+    self.parsed_args = parsed_args
+    self.results_queue = results_queue
+    self.rw_locks = rw_locks
+    self.shutdown_event = shutdown_event
 
-
-class NGEvaluation:
-  """Handle to a nevergrad evaluation."""
-
-  def __init__(self, proposal, problem_instance):
-    self.proposal = proposal
-    self.problem_instance = problem_instance
-
-
-class NGMPEvaluation(NGEvaluation):
-  """Handle to a multiprocess nevergrad evaluation."""
-
-  def __init__(self, proposal, problem_instance, process, ipc_dict):
-    super().__init__(proposal, problem_instance)
-    self.process = process
-    self.ipc_dict = ipc_dict
-    self.joined_with_root = False
-
-  def ipc_state(self):
-    return self.ipc_dict['result'] if 'result' in self.ipc_dict else \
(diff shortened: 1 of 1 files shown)


python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -17,166 +17,52 @@
...
+  benchmark may be run on multiple CPUs.
 
   Upon SIGTERM triggered by the parent, locks are simply dropped on the floor
   and everything stops.
   """
+
+  global results_queue, parsed_args, rw_locks
+
+  # Assigned CPU of this process.
+  compilation_cpu = os.sched_getaffinity(0).pop()
```
nicolasvasilache, line 55: please add a comment to explain what this does

nicolasvasilache (approved):
Very nice, thanks for these improvements!

nicolasvasilache:
I can confirm the speedup!

python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -17,166 +17,52 @@
...
+  This is run in interruptible multiprocess mode. Result are returned from this
+  process via `result_queue`.
 
-  ipc_dict must be provided, and it is used to return information across the
-  root / children process boundary:
-    - 'throughputs': the measured throughputs.
-    - 'success': the return status.
+  Before starting compilation, acquire a lock in read mode. Wait for a certain
+  amount of time and kill the process if compilation is still running. The
+  compilation is always run on a single CPU. (The process is alrady pinned to a
```
nicolasvasilache, line 41: nit: typo already

python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -17,166 +17,52 @@
...
-  The benchmark may be run on multiple CPUs, set by a taskset command.
+  Before starting a benchmark, acquire a lock in write mode. Wait for a certain
+  amount of time and kill the process if the benchmark is still running. The
+  benchmark may be run on multiple CPUs.
 
   Upon SIGTERM triggered by the parent, locks are simply dropped on the floor
   and everything stops.
   """
+
+  global results_queue, parsed_args, rw_locks
```
nicolasvasilache, line 52: do we need these to be globals?

python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -201,123 +87,92 @@ def schedule_and_save(module):
...
         problem.run(
-            n_iters=n_iters,
+            n_iters=parsed_args.n_iters,
             entry_point_name=scheduler.entry_point_name,
             runtime_problem_sizes_dict=problem.compile_time_problem_sizes_dict))
 
-  try:
+  # Function called when compilation or benchmark times out.
+  def kill_process():
+    # Put the compilation CPU back into the queue, so that a newly spawned
```
nicolasvasilache, line 96: much more dynamic and better, thanks!

python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -351,120 +206,145 @@ def finalize_parallel_search(scheduler: NGSchedulerInterface, \
...
+                    rw_locks_arg: tp.Sequence[RWLock],
+                    shutdown_event_arg):
+    """Initialize an mp.Pool process.
+
+    Store frequently used objects in global variables and select a pinned
+    compilation CPU for this process.
+    """
+
+    # Store arguments in global variables.
+    global available_cpus_queue, parsed_args, results_queue, rw_locks, shutdown_event
```
nicolasvasilache, line 256: eeesh .. do we really need this ?

nicolasvasilache (approved):

python/examples/core/nevergrad_parallel_utils.py
matthias-springer: `results_queue` and `parsed_args` could be passed each time we enqueue a search job. Setting it up during process initialization and storing them in global variables is a bit more efficient.

The `rw_locks`, we actually need a global variable. I ran into a weird bug where pool processes kept crashing. `RWLock` should pickle according to the documentation, but there must be a bug in their implementation somewhere. When setting up stuff during process initialization, we can pass stuff that does not pickle.

So since we need global variables anyway, I'd keep the others as well for efficiency reasons.

python/examples/core/nevergrad_parallel_utils.py
matthias-springer: I could put them into a single objects, so that we only have one global variable.

PR #297, opened 2022-02-24
Put vector.transfer_write distribution into a pattern.

matthias-springer:
Just a cleanup, since most other functionality is also a pattern.

### DIFF
--- a/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
+++ b/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
@@ -374,26 +374,29 @@ static bool hasSideEffect(Operation &op) {
 // TODO: Move to the op.
 static unsigned distributionRatio = 32;
 
-void mlir::vector_ext::distributeTransferWrite(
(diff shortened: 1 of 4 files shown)


lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -374,75 +374,94 @@ static bool hasSideEffect(Operation &op) {
...
+        distributionMapFn(fn) {}
+
+  LogicalResult matchAndRewrite(vector::TransferWriteOp writeOp,
+                                PatternRewriter &rewriter) const override {
+    auto warpOp = dyn_cast<WarpSingleLaneOp>(writeOp->getParentOp());
+    if (!warpOp)
+      return failure();
+
+    // There must be no op with a side effect after writeOp.
+    // TODO: What about ops with regions?
```
ThomasRaoux, line 390: `hasSideEffect` checks for recursive side effects so I don't think it should be problem

ThomasRaoux (approved):

PR #301, opened 2022-02-25
Add pattern to forward warpSingleLaneOp operands

ThomasRaoux:
Add builder to create warpOp with operands and add a pattern to
propagate the operands when they are directly returned out of the region

### DIFF
--- a/include/Dialects/VectorExt/VectorExtOps.td
+++ b/include/Dialects/VectorExt/VectorExtOps.td
@@ -148,8 +148,9 @@ def VectorExt_WarpSingleLaneOp : VectorExt_Op<"warp_execute_on_lane_0",
   let skipDefaultBuilders = 1;
   let builders = [
     OpBuilder<(ins "Value":$laneid)>,
-    // TODO: Add arguments support in the builder.
-    OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid)>
+    OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid)>,
+    OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid,
+                  "ValueRange":$operands, "TypeRange":$argTypes)>
   ];
 
   let extraClassDeclaration = [{
--- a/include/Dialects/VectorExt/VectorExtOps.td
+++ b/include/Dialects/VectorExt/VectorExtOps.td
@@ -149,8 +149,11 @@ def VectorExt_WarpSingleLaneOp : VectorExt_Op<"warp_execute_on_lane_0",
   let builders = [
     OpBuilder<(ins "Value":$laneid)>,
     OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid)>,
+    // `blockArgTypes` are different than `args` types as they are they
+    // represent all the `args` instances visibile to lane 0. Therefore we need
+    // to explicit pass the type.
     OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid,
-                  "ValueRange":$operands, "TypeRange":$argTypes)>
+                   "ValueRange":$args, "TypeRange":$blockArgTypes)>
   ];
 
   let extraClassDeclaration = [{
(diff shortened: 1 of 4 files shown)


include/Dialects/VectorExt/VectorExtOps.td
```diff
@@ -148,8 +148,9 @@ def VectorExt_WarpSingleLaneOp : VectorExt_Op<"warp_execute_on_lane_0",
   let skipDefaultBuilders = 1;
   let builders = [
     OpBuilder<(ins "Value":$laneid)>,
-    // TODO: Add arguments support in the builder.
-    OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid)>
+    OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid)>,
+    OpBuilder<(ins "TypeRange":$resultTypes, "Value":$laneid,
+                  "ValueRange":$operands, "TypeRange":$argTypes)>
```
matthias-springer, line 153: nit: indent

matthias-springer: nit: I would call them `$args` instead of `$operands`

matthias-springer: I would rename `argTypes` to `blockArgTypes` and add a comment that this cannot be inferred from `args` and is needed to distinguish between broadcasting/gathering from lanes (are these the right terms?) semantics.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -354,6 +355,37 @@ struct WarpOpDeadResult : public OpRewritePattern<WarpSingleLaneOp> {
...
+      Value warpOperand = warpOp.args()[arg.getArgNumber()];
+      if (result.use_empty() || result.getType() != warpOperand.getType())
+        continue;
+      valForwarded = warpOperand;
+      resultIndex = operand.getOperandNumber();
+      break;
+    }
+    if (!valForwarded)
+      return failure();
+    warpOp.getResult(resultIndex).replaceAllUsesWith(valForwarded);
```
matthias-springer, line 384: Is this safe? I always thought we would have to go through the rewriter API or do something like `updateRootInplace` (which does not work with `replaceAllUsesWith`).

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -354,6 +355,37 @@ struct WarpOpDeadResult : public OpRewritePattern<WarpSingleLaneOp> {
...
+        warpOp.getBodyRegion().getBlocks().begin()->getTerminator());
+    Value valForwarded;
+    unsigned resultIndex;
+    for (OpOperand &operand : yield->getOpOperands()) {
+      auto arg = operand.get().dyn_cast<BlockArgument>();
+      if (!arg || arg.getOwner()->getParentOp() != warpOp.getOperation())
+        continue;
+      Value result = warpOp.getResult(operand.getOperandNumber());
+      Value warpOperand = warpOp.args()[arg.getArgNumber()];
+      if (result.use_empty() || result.getType() != warpOperand.getType())
```
matthias-springer, line 376: Is this something that we support?

```mlir
%0 = vector_ext.warp_execute_on_lane_0(%laneid) args(%v0 : vector<1xf32>) -> (vector<1xf32>) {
 ^bb0(%arg0 : vector<1xf32>) :
  vector_ext.yield %arg0 : vector<1xf32>
}
```

(With the assumption that every lane has the same value.)

Or is this forbidden and we would just capture the `%v0` without going through `args`?

matthias-springer (approved):
Nice, I needed this for one of my changes.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: That's a good point. I think having warp_execute_on_lane_0 argument and the block argument have the same type should only be legal if the value is uniform. (or maybe we should disallow this completely). I would go with the assumption that this is correct to convert this example to `%v0` but we should think about it more and make sure the semantic is coherent with the rest.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: This doesn't modify the root so I though we didn't need `updateRootInplace` here. I see some other examples upstream which seem to do something similar so I assumed it was correct:
https://github.com/llvm/llvm-project/blob/5d47332783d0421ba0da2a723354fb523eb85d0b/mlir/lib/Dialect/SCF/SCF.cpp#L1449

Although I'm not completely clear if this case is allowed

include/Dialects/VectorExt/VectorExtOps.td
ThomasRaoux: done.

PR #302, opened 2022-02-25
Add pattern to sink ForOp out of WarpSingleLane region

ThomasRaoux:
This adds a pattern that can sink a for op region. This creates a new
WarpSingleLane region within the scf.for which will trigger further
lowering of this op.

The first commit is from https://github.com/google/iree-llvm-sandbox/pull/301 no need to review it here.

### DIFF
--- a/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
+++ b/lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
@@ -509,11 +509,105 @@ struct WarpOpTransferWrite : public OpRewritePattern<vector::TransferWriteOp> {
   DistributionMapFn distributionMapFn;
 };
 
+/// Sink scf.for region out of WarpSingleOp. This can be done only if the
+/// scf.ForOp is the last operation in the region so that it doesn't change the
+/// order of execution. This creates a new scf.for region after the
+/// WarpSingleOp. The new scf.for region will contain a new WarpSingleLaneOp
+/// region. Example:
+/// ```
+/// %w = vector_ext.warp_execute_on_lane_0(%laneid) -> (vector<4xf32>) {
+///   ...
+///   %v1 = scf.for %arg3 = %c0 to %c128 step %c1 iter_args(%arg4 = %v)
+///   -> (vector<128xf32>) {
+///     ...
+///     scf.yield %r : vector<128xf32>
+///   }
+///   vector_ext.yield %v1 : vector<128xf32>
+/// }
+/// ```
+/// To:
+/// %w0 = vector_ext.warp_execute_on_lane_0(%arg0) -> (vector<4xf32>) {
+///   ...
+///   vector_ext.yield %v : vector<128xf32>
+/// }
+/// %w = scf.for %arg3 = %c0 to %c128 step %c1 iter_args(%varg = %q0)
+///   -> (vector<4xf32>) {
+///     %iw = vector_ext.warp_execute_on_lane_0(%laneid)
+///     args(%varg : vector<4xf32>) -> (vector<4xf32>) {
+///     ^bb0(%arg: vector<128xf32>):
+///       ...
+///       vector_ext.yield %ir : vector<128xf32>
+///     }
+///     scf.yield %iw : vector<4xf32>
+///  }
+/// ```
+struct WarpOpScfForOp : public OpRewritePattern<WarpSingleLaneOp> {
+  using OpRewritePattern<WarpSingleLaneOp>::OpRewritePattern;
+  LogicalResult matchAndRewrite(WarpSingleLaneOp warpOp,
+                                PatternRewriter &rewriter) const override {
+    auto yield = cast<vector_ext::YieldOp>(
+        warpOp.getBodyRegion().getBlocks().begin()->getTerminator());
+    // Only pick up forOp if it is the last op in the region.
+    Operation *lastNode = yield->getPrevNode();
+    auto forOp = dyn_cast_or_null<scf::ForOp>(lastNode);
+    if (!forOp)
+      return failure();
+    SmallVector<Value> newOperands;
+    SmallVector<unsigned> resultIdx;
+    // Collect all the outputs coming from the forOp.
+    for (OpOperand &yieldOperand : yield->getOpOperands()) {
+      if (yieldOperand.get().getDefiningOp() != forOp.getOperation())
+        continue;
+      auto forResult = yieldOperand.get().cast<OpResult>();
+      newOperands.push_back(warpOp.getResult(yieldOperand.getOperandNumber()));
+      yieldOperand.set(forOp.getIterOperands()[forResult.getResultNumber()]);
+      resultIdx.push_back(yieldOperand.getOperandNumber());
+    }
+    OpBuilder::InsertionGuard g(rewriter);
+    rewriter.setInsertionPointAfter(warpOp);
+    // Create a new for op outside the region with a WarpSingleOp region inside.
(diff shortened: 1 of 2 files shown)


lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -482,10 +482,105 @@ struct WarpOpTransferWrite : public OpRewritePattern<vector::TransferWriteOp> {
...
+      return failure();
+    SmallVector<Value> newOperands;
+    SmallVector<unsigned> resultIdx;
+    // Collect all the outputs coming from the forOp.
+    for (OpOperand &yieldOperand : yield->getOpOperands()) {
+      if (yieldOperand.get().getDefiningOp() != forOp.getOperation())
+        continue;
+      newOperands.push_back(warpOp.getResult(yieldOperand.getOperandNumber()));
+      yieldOperand.set(
+          forOp.getIterOperands()[yieldOperand.getOperandNumber()]);
```
matthias-springer, line 536: I think this won't work. E.g.

```mlir
execute_on_lane_0(...) {
  %v1 = "some_op"() : () -> vector<32xf32>
  %v2 = scf.for ... iter_args(%arg = %v1) ...
  vector_ext.yield %v1, %v2
}
```

(array access will be out of bounds)

Also, what if the `scf.for` is yielding vectors in a different order? E.g.,
```mlir
scf.for ... iter_args(%arg0 = %v0, %arg1 = %v1) {
  scf.yield %arg1, %arg0
}
```

Will this still work?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -482,10 +482,105 @@ struct WarpOpTransferWrite : public OpRewritePattern<vector::TransferWriteOp> {
...
+    BlockAndValueMapping mapping;
+    mapping.map(forOp.getInductionVar(), newForOp.getInductionVar());
+    for (auto args : llvm::zip(forOp.getRegionIterArgs(),
+                               innerWarp.getBody()->getArguments())) {
+      mapping.map(std::get<0>(args), std::get<1>(args));
+    }
+    rewriter.setInsertionPoint(innerWarp.getBody(),
+                               innerWarp.getBody()->begin());
+    for (Operation &innerOp : forOp.getBody()->without_terminator())
+      rewriter.clone(innerOp, mapping);
```
matthias-springer, line 559: I would use `RewriterBase::mergeBlocks`. You can move over the entire loop body to a new loop. See `Dialect/SCF/Transforms/BufferizableOpInterfaceImpl.cpp:370` for an example.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux:
```
execute_on_lane_0(...) {
  %v1 = "some_op"() : () -> vector<32xf32>
  %v2 = scf.for ... iter_args(%arg = %v1) ...
  vector_ext.yield %v1, %v2
}
```
is not legal as `scf.for` requires the number of arguments to match the number of results (as all the values are loop carried dependency.

I believe the second case will work but let me add a test so that we are both convinced :)

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: That's what I add tried originally but the blocks don't exactly match as the scf.for as the induction variable as extra block argument.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: I added a test where the variable are swapped, I think the IR generated is correct, let me know if you are not convinced

matthias-springer: Note, in the above example the `scf.for` has one iter_arg and one result. The yield in the example belongs to the `execute_on_lane_0` region.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
matthias-springer: I think it should still work. You can use `argValues` to pass replacement values. In the same way you are using the BlockAndValueMapping at the moment.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
matthias-springer: The problem is that `yieldOperand.getOperandNumber()` belongs to the `vector_ext.yield` op, but you're using it to index into `forOp.getIterOperands()`, which can have a different number of iter_args (and return values).

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
matthias-springer: I think an scf::ForOp block always starts with the IV, followed by all iter_args. So you should be able to pass `/*argValues=*/{newForOp.getInductionVar(), innerWarp.getBody()->getArguments() /*expanded*/}`.

Actually, it's not going to be one-liner as I hoped... So not sure if it's worth it.

matthias-springer (approved):

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -482,10 +482,105 @@ struct WarpOpTransferWrite : public OpRewritePattern<vector::TransferWriteOp> {
...
+///     %iw = vector_ext.warp_execute_on_lane_0(%laneid)
+///     args(%varg : vector<4xf32>) -> (vector<4xf32>) {
+///     ^bb0(%arg: vector<128xf32>):
+///       ...
+///       vector_ext.yield %ir : vector<128xf32>
+///     }
+///     scf.yield %iw : vector<4xf32>
+///  }
+/// ```
+struct WarpOpScfForOp : public OpRewritePattern<WarpSingleLaneOp> {
```
nicolasvasilache, line 517: This seems to do 2 things at the same time and feels like it would benefit from being split.
Step 1: splice the vector_ext.warp_execute_on_lane_0 just before the last for loop.
Step 2: swap the unique scf.for with the enclosing vector_ext.warp_execute_on_lane_0.

The first functionality in step 1 seems generally useful; don't we already have something like this?

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: you're absolutely right. I had read too fast. Thanks for catching this! I fixed the wrong index and updated the order of yield operand in the last test to make sure this case is handled correctly.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: This is a great point. I had not seen it that way but it is much simpler indeed. The transformation of slicing the `vector_ext.warp_execute_on_lane_0` will be needed in more cases as you assumed. For instance this is also what is done in Matthias' PR: https://github.com/google/iree-llvm-sandbox/pull/303
What I'm wondering now is whether it should be a helper or a pattern on its own. Starting with a common helper is probably the way to go.
I'll push the code this way and then refactor this to create a helper that slice `vector_ext.warp_execute_on_lane_0`

PR #303, opened 2022-02-26
Extract transfer_write with vector<1x> from warp op

matthias-springer:


test/VectorExt/warp_distribution.mlir
```diff
@@ -50,13 +28,45 @@ func @warp(%laneid: index, %arg1: memref<1024xf32>, %arg2: memref<1024xf32>,
...
+// CHECK-D-LABEL: func @warp_extract(
+//       CHECK-D:   %[[C1:.*]] = arith.constant 1 : index
+//       CHECK-D:   %[[WARPOP:.*]] = vector_ext.warp_execute_on_lane_0(%{{.*}}) -> (vector<1xf32>)
+//       CHECK-D:     "test.dummy_op"
+//       CHECK-D:     vector_ext.yield %{{.*}} : vector<1xf32>
+//       CHECK-D:   }
+//       CHECK-D:   vector_ext.warp_execute_on_lane_0(%{{.*}}) args(%[[WARPOP]] : vector<1xf32>) {
+//       CHECK-D:     ^bb0(%[[BROADCASTED:.*]]: vector<32xf32>):
+//       CHECK-D:     %[[MASK:.*]] = vector.create_mask %[[C1]] : vector<32xi1>
+//       CHECK-D:     vector.transfer_write %[[BROADCASTED]], %{{.*}}[%{{.*}}], %[[MASK]]
```
ThomasRaoux, line 64: As mentioned in the previous comment I don't think we need to create a masked transfer_write with vector<32xf32> arg here.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -446,6 +437,83 @@ struct WarpOpTransferWrite : public OpRewritePattern<vector::TransferWriteOp> {
...
+    rewriter.setInsertionPointAfter(newWarpOp);
+
+    // Create a second warp op that contains only writeOp (and a mask). Pass in
+    // the vector that should be written.
+    auto secondWarpOp = rewriter.create<WarpSingleLaneOp>(
+        loc, TypeRange(), newWarpOp.laneid());
+    secondWarpOp.argsMutable().assign(
+        newWarpOp.getResult(newWarpOp.getNumResults() - 1));
+    Block &body = secondWarpOp.getBodyRegion().front();
+    VectorType broadcastedType =
```
ThomasRaoux, line 466: I don't think we need to do that since we know we broadcast the value from the previous region we can just use the value directly as a uniform defined outside of the region

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -380,44 +380,35 @@ struct WarpOpTransferWrite : public OpRewritePattern<vector::TransferWriteOp> {
...
+                                WarpSingleLaneOp warpOp) const {
     AffineMap map = distributionMapFn(writeOp);
     SmallVector<int64_t> targetShape(writeOp.getVectorType().getShape().begin(),
                                      writeOp.getVectorType().getShape().end());
     assert(map.getNumResults() == 1 &&
            "multi-dim distribution not implemented yet");
     for (unsigned i = 0, e = map.getNumResults(); i < e; i++) {
       unsigned position = map.getDimPosition(i);
+      if (targetShape[position] < distributionRatio)
+        return failure();
```
ThomasRaoux, line 396: Is this needed? This seems like a subset of the what is tested right below: (`targetShape[position] % distributionRatio != 0`)

ThomasRaoux (approved):

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
matthias-springer: you're right!

PR #304, opened 2022-02-28
Split vector reduction

matthias-springer:


matthias-springer:
Only review the last commit.

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -265,6 +267,48 @@ struct WarpOpReduction : public OpRewritePattern<WarpSingleLaneOp> {
...
+      else
+        accumulator = vector::makeArithReduction(
+            rewriter, loc, reductionOp.kind(), accumulator, warpReduction);
+    }
+
+    rewriter.replaceOp(reductionOp, accumulator);
+    return success();
+  }
+};
+
```
ThomasRaoux, line 311: This is fine for now but I think eventually we should make this use the same pattern and interface as the rest of the unroll vector patterns:
https://github.com/llvm/llvm-project/blob/1768cb3a674ad657dd3eccd001f0a3e9a7675af6/mlir/lib/Dialect/Vector/Transforms/VectorUnrollDistribute.cpp#L210

This way it would be independent of Warp ops and can be used for other things. (and can be upstreamed independently.)
It is fine if you prefer to leave it like that for now to make progress on other stuff but I would at least add a comment.

ThomasRaoux:
Looks good to me, just a comment on how I think it should look like eventually.

PR #306, opened 2022-02-28
Fixing LinalgTransform bug

giuseros:
This is fixing the bug I raised in https://github.com/google/iree-llvm-sandbox/issues/300 and adding a test case

### DIFF
--- a/lib/Dialects/LinalgTransform/IR/TransformOpInterface.cpp
+++ b/lib/Dialects/LinalgTransform/IR/TransformOpInterface.cpp
@@ -79,12 +79,13 @@ transform::TransformState::applyTransform(TransformOpInterface transform) {
   if (failed(transform.apply(results, *this)))
     return failure();
 
+  for (Value target : transform->getOperands())
+    removePayloadOps(target);
+
   for (auto en : llvm::enumerate(transform->getResults()))
     if (failed(setPayloadOps(en.value(), results.get(en.index()))))
       return failure();
 
-  for (Value target : transform->getOperands())
-    removePayloadOps(target);
   return success();
 }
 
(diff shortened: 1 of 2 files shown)


giuseros:
cc @ftynse

lib/Dialects/LinalgTransform/IR/TransformOpInterface.cpp
```diff
@@ -79,12 +79,15 @@ transform::TransformState::applyTransform(TransformOpInterface transform) {
   if (failed(transform.apply(results, *this)))
     return failure();
 
-  for (auto en : llvm::enumerate(transform->getResults()))
+  for (Value target : transform->getOperands()) {
```
ftynse, line 82: Nit: drop trivial braces.

test/LinalgTransform/double_tiling_alp.mlir
```diff
@@ -0,0 +1,33 @@
+// RUN: mlir-proto-opt -linalg-interp-transforms %s | FileCheck %s
+
+// CHECK-LABEL: func @matmul_tensors(
+// CHECK-SAME:    %[[TA:[0-9a-z]+]]: tensor<128x128xf32>
+// CHECK-SAME:    %[[TB:[0-9a-z]+]]: tensor<128x128xf32>
+// CHECK-SAME:    %[[TC:[0-9a-z]+]]: tensor<128x128xf32>
+// CHECK-SAME:  -> tensor<128x128xf32> {
```
ftynse, line 7: I'm not sure these check lines are necessary. The test seems to ensure the interpreter doesn't assert, so just having one label line and ensure some IR was produced is enough, the rest will just create churn should the syntax change in the future. Also consider adding a simple comment stating that only the success of the transformation is relevant here.

ftynse (approved):
LGTM with changes

giuseros:
@ftynse , all the changes have been applied (sorry for the delay)

PR #309, opened 2022-03-01
Add a minimal skeleton for a library callable from MLIR

nicolasvasilache:
Assuming mlir-opt and mlir-cpu-runner are built and in the path.

```
~/github/iree-llvm-sandbox/iterators_build$ >

mlir-opt ../experimental/iterators/test/simple.mlir -convert-std-to-llvm | mlir-cpu-runner -e main -entry-point-result=void -shared-libs=$(pwd)/lib/Utils/libruntime_utils.so
```

Prints:
```
Create a dummy pointer 0xdeadbeef
Use a dummy pointer 0xdeadbeef
```

### DIFF
--- a/experimental/iterators/CMakeLists.txt
+++ b/experimental/iterators/CMakeLists.txt
@@ -1,7 +1,14 @@
 cmake_minimum_required(VERSION 3.0.0)
 project(iterators VERSION 0.1.0)
 
+# TODO: Atm LLVM is still in C++14 land.
 set(CMAKE_CXX_STANDARD 17)
+
+include_directories(include)
+add_subdirectory(lib)
+
+
+
 
 add_subdirectory(docs)
 
@@ -23,6 +30,8 @@ target_include_directories(iterators PUBLIC include)
 include(CTest)
 enable_testing()
 
+# TODO: move these to the tests director.
+# TODO: split tests into C++ and MLIR
 add_executable(iterator_tests
         tests/column_scan_operator_test.cc
         tests/reduce_operator_test.cc
--- a/experimental/iterators/test/simple.mlir
+++ b/experimental/iterators/test/simple.mlir
@@ -1,0 +1,17 @@
+// TODO: this currently does not propertly connect to an actual execution
+
+// R-UN: mlir-opt %s -convert-std-to-llvm -reconcile-unrealized-casts | \
+// R-UN: mlir-cpu-runner -e main -entry-point-result=void \
+// R-UN:   -shared-libs=$(pwd)/lib/Utils/libruntime_utils.so \
+// R-UN: | FileCheck %s
+
+func private @oneWay() -> (i64)
+func private @otherWay(i64)
+
+func @main() {
+  // CHECK: Create a dummy pointer 0xdeadbee
+  %p = call @oneWay() : () -> (i64)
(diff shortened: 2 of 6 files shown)


experimental/iterators/test/simple.mlir
```diff
@@ -0,0 +1,15 @@
...
+// R-UN: mlir-opt %s -convert-std-to-llvm -reconcile-unrealized-casts | \
+// R-UN: mlir-cpu-runner -e main -entry-point-result=void \
+// R-UN:   -shared-libs=$(pwd)/lib/Utils/libruntime_utils.so \
+// R-UN: | FileCheck %s
+
+func private @oneWay() -> (i64)
+func private @otherWay(i64)
+
+func @main() {
+  %p = call @oneWay() : () -> (i64)
```
ingomueller-net, line 12: Maybe add a dummy check?

```suggestion
  %p = call @oneWay() : () -> (i64)
  // CHECK: Create a dummy pointer 0xdeadbee
```

experimental/iterators/test/simple.mlir
```diff
@@ -0,0 +1,15 @@
+// TODO: this currently does not propertly connect to an actual execution
+
+// R-UN: mlir-opt %s -convert-std-to-llvm -reconcile-unrealized-casts | \
+// R-UN: mlir-cpu-runner -e main -entry-point-result=void \
+// R-UN:   -shared-libs=$(pwd)/lib/Utils/libruntime_utils.so \
+// R-UN: | FileCheck %s
```
ingomueller-net, line 6: I have managed to run this manually. How is this normally run? I guess by some CMake infrastructure? I'd have to include this project into the parent build to achieve this, right? (I guess I'll have to do that sooner or later, anyways.)

experimental/iterators/CMakeLists.txt
```diff
@@ -23,6 +30,8 @@ target_include_directories(iterators PUBLIC include)
 include(CTest)
 enable_testing()
 
+# TODO: move these to the tests director.
+# TODO: split tests into C++ and MLIR
```
ingomueller-net, line 34: Will do. I saw that LLVM uses googletests for some things (in the unittest directories). I guess I can adopt that pattern.

ingomueller-net (commented):
Cool, thanks a lot! Only minor comments and question and one suggestion.

experimental/iterators/test/simple.mlir
nicolasvasilache: In LLVM proper, this is using [lit](https://llvm.org/docs/CommandGuide/lit.html) and [FileCheck](https://llvm.org/docs/CommandGuide/FileCheck.html).

@gysit connected these pieces in a python script here: https://github.com/google/iree-llvm-sandbox/blob/main/run_tests.py; you should be able to copy-paste and reuse and then s/R-UN/RUN in the example.

experimental/iterators/CMakeLists.txt
```diff
@@ -1,8 +1,15 @@
 cmake_minimum_required(VERSION 3.0.0)
 project(iterators VERSION 0.1.0)
 
+# TODO: Atm LLVM is still in C++14 land.
```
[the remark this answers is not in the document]
nicolasvasilache: sgtm!

PR #310, opened 2022-03-02
POC: Different ways of mondo-building IREE and sandbox.

stellaraccident:


nicolasvasilache:
the iree-integrate branch should supersede this.

PR #311, opened 2022-03-02
Fix bug in distribution of element-wise ops

matthias-springer:


lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -154,10 +154,7 @@ struct WarpOpElementwise : public OpRewritePattern<WarpSingleLaneOp> {
     SmallVector<Value> yieldValues;
     SmallVector<Type> retTypes;
     for (OpOperand &operand : elementWise->getOpOperands()) {
-      auto targetType = VectorType::get(
-          distributedVal.getType().cast<VectorType>().getShape(),
-          operand.get().getType().cast<VectorType>().getElementType());
-      retTypes.push_back(targetType);
+      retTypes.push_back(distributedVal.getType());
```
ThomasRaoux, line 157: So that won't work for operations like cast, or select that may have different element type. It would be good to extend to those kind of ops.

ThomasRaoux (approved):

PR #312, opened 2022-03-02
Bump llvm version after func dialect landing.

gysit:


nicolasvasilache (approved):
Thanks for the fast turnaround!

PR #316, opened 2022-03-04
Iterators: Integrate into main CMake infra.

ingomueller-net:
To the best of my understanding, this PR makes the (experimental) "Iterators" subproject follow the project structure of other subprojects in this repository, maybe with the exception of two quirks:

(1) Since the code relies on some C++17 features, the headers are defined as a header-only library target, and the standard version is set to C++17 just for that target. Most (or all) CMake set-ups I have seen don't bother with targets and just use `include_directories`.
(2) I didn't spend time making the compilation of doxygen target in line with the remainder of the repo. It is built when the Iterator library is built using some simple recipe, which should be good enough for a sanity check.

Since LLVM/MLR/the sandbox repo is built with higher warning levels, I now have some cleaning up to do.


nicolasvasilache (approved):
Looks good, thanks!

Merging as is, you may want to address the nit separately and just land it without further approval.

PR #318, opened 2022-03-04
Drop support for pass pipeline in the Tile transform and follow consequences.

nicolasvasilache:
From now on, the Tile tranform can only be used with the transform dialect.
Ripping this bandaid has a number of implications and triggers a bunch of deletions that this PR implements.
    
A few bugs are exposed in the FIXME, they can be distributed and addressed as followups.
    
Followup cleanups can now easily occur on top of this PR, all the way to retiring most of the codegen strategy driver.


FIXME
```diff
@@ -0,0 +1,2 @@
+python -m python.examples.fusion.test is currently deactivated until fusion has a transform dialect impl.
+python -m python.examples.reduction.reduction_1d_bench fails in certain cases.
```
gysit: nit: newline missing

nicolasvasilache:
There are a few regressions but nothing that I would consider serious, this is part of ripping the bandaid.
I saved the hash before this commit so we can refer back and investigate if something is hard in the future.

nicolasvasilache:
This has landed as `de7c1b0c451207b54324cea50705f3812cb6f4ab` and not as [efd7728](https://github.com/google/iree-llvm-sandbox/commit/efd772820f1a4d6c22afc62c0781798fee1498a2) (which was a mistake commit).

PR #321, opened 2022-03-07
Add option to plot throughput analysis after search

matthias-springer:
Take the best configuration and plot how the throughput is changing when
changing one feature at a time.


python/examples/core/nevergrad_parallel_utils.py
```diff
@@ -231,6 +232,79 @@ def finalize_parallel_search(scheduler: NGSchedulerInterface, \
                                     module_save_filename=final_module_filename,
                                     benefit=best)
 
+class Plotting:
```
nicolasvasilache, line 235: should this go into the "non-paralllel" NG utils?
Generally we also want to refactor utils a bit and have less of a flat file structure.

nicolasvasilache (approved):

PR #323, opened 2022-03-07
Iterators: Fix some issues found by clang warnings

ingomueller-net:
These surfaced after compiling as part of the top-level project (cf #316).

I am not too sure what the best practices are for signed/unsigned comparisons. Input welcome!


ingomueller-net:
I should say that this is meant to me merged only after #316.

nicolasvasilache (approved):

PR #324, opened 2022-03-07
Iterators: Configure lit for compiler tests and unittests

ingomueller-net:
Not 100% if this follows best practices; this is really just some copy'n'paste + trial'n'error from existing folders...

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -57,13 +57,13 @@ jobs:
     - name: Build
       run: |
         cd ${GITHUB_WORKSPACE}/sandbox
-        python configure.py --llvm-path=../llvm-project
+        python configure.py --llvm-path=../llvm-project --iterators
         ccache -s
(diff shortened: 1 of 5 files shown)


ingomueller-net:
This is meant to be merged after #323.

ingomueller-net:
My first attempt made the tests fail. The reason is that, if I change `run_tests.py`, I need to make everything in CI work, which means I need to build the `iterators` subproject. I just added another commit that does that; now the tests pass. What is a bit suspicious is that I don't see the `IteratorsTests` target being built, nor do I see the output of the lit tests, so I am not 100% sure if things are actually run...

nicolasvasilache (approved):
Looks good assuming the tests go back to green after rebase.

ingomueller-net:
I did small changes since your review: (1) make running the iterator tests in `run_tests.py` optional (because building it is optional), and (2) add a new commit that follows the renaming of `-convert-std-to-llvm` to `-convert-func-to-llvm`.

@nicolasvasilache: You want to check again or should I go ahead and merge?

nicolasvasilache:
`@nicolasvasilache: You want to check again or should I go ahead and merge?`

I guess you have your answer :) thanks!

PR #325, opened 2022-03-08
Bump llvm

chelini:


chelini:
Dear All, unrelated to this I also have a question regarding `createLinalgTransformInterpreterPass`. Why the pass is not exposed to the users? I have a use case where I would like to create the pass and add it to the pass manager. Do you feel it is a good idea to expose the pass?

gysit:
> Dear All, unrelated to this I have a question regarding `createLinalgTransformInterpreterPass`. Why the pass is not exposed to the users? I have a use case where I would like to create the pass and add it to the pass manager. Do you feel it is a good idea to expose the pass?

Others may correct me if I am wrong. I assume there was no use case so far and it is fine exposing the method in the header.

chelini:
> > Dear All, unrelated to this I have a question regarding `createLinalgTransformInterpreterPass`. Why the pass is not exposed to the users? I have a use case where I would like to create the pass and add it to the pass manager. Do you feel it is a good idea to expose the pass?
> 
> Others may correct me if I am wrong. I assume there was no use case so far and it is fine exposing the method in the header.

Thanks.

gysit (approved):
Thanks!

nicolasvasilache (changes requested):
Can we please split this up in multiple PRs? 

There seem to be multiple unrelated changes all inside a single PR atm.

chelini:
Ok, I will close this, and create multiple smaller PRs.

PR #334, opened 2022-03-08
Add execution test for tile reduction

ThomasRaoux:


lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -368,12 +368,25 @@ struct WarpOpForwardOperand : public OpRewritePattern<WarpSingleLaneOp> {
...
+          continue;
+        // Assume all the values coming from above are uniform.
+        if (result.getType() != operand.get().getType())
+          continue;
+        valForwarded = operand.get();
+        resultIndex = operand.getOperandNumber();
+        continue;
+      }
       auto arg = operand.get().dyn_cast<BlockArgument>();
       if (!arg || arg.getOwner()->getParentOp() != warpOp.getOperation())
```
matthias-springer, line 386: Not sure if there's a use case for it, but it would be easy to also support bbArgs from a parent op (not just the warpOp).

matthias-springer (approved):

lib/Dialects/VectorExt/Transform/VectorWarpUtils.cpp
ThomasRaoux: good, point, change the logic to handle both

PR #336, opened 2022-03-09
Delete dead transforms code

nicolasvasilache:


python/examples/core/transforms.py
```diff
@@ -262,15 +246,6 @@ class Generalize(Transform):
   def __init__(self, fun_name: str, op_name: str, **kwargs):
     self.fun_name = fun_name
     self.op_name = op_name
-    self._parse_variables_in_kwargs(kwargs)
-    interchange_str = _get_size_list_as_str(name='iterator-interchange',
```
gysit, line 266: is _get_size_list_as_str dead code as well?

gysit (approved):

python/examples/core/transforms.py
nicolasvasilache: Not yet, we still have uses in Fuse and in LinalgExtTile but both are arguably broken.
Will cleanup in a later incantation as I bring up LinalgExtTile

PR #337, opened 2022-03-09
Iterators: Add missing new lines at end of several files.

ingomueller-net:


ingomueller-net:
Landing right away as requested by @nicolasvasilache in #316.

PR #338, opened 2022-03-09
Split padding from tiling.

gysit:
- Remove the padding option from the tile operation.
- Update the tests to use a separate Pad step instead of padding right after tiling.


python/examples/dialects/linalg_transform_test.py
```diff
@@ -45,8 +45,8 @@ def tile_twice():
   sequence = transform.SequenceOp()
   with ir.InsertionPoint(sequence.body.blocks[0]):
     target = transform.MatchOp("foo")
-    tiled1 = transform.TileOp(target, sizes=[128, 32], pad=True)
-    tiled2 = transform.TileOp(tiled1, sizes=[32, 16], pad=True)
+    tiled1 = transform.TileOp(target, sizes=[128, 32])
```
nicolasvasilache, line 48: is this test now missing padding ?

test/LinalgTransform/invalid.mlir
```diff
@@ -12,14 +12,6 @@ linalg_transform.sequence {
 
 // -----
 
-linalg_transform.sequence {
```
nicolasvasilache, line 15: isn't there some positive mlir test to update that exercises padding chaining after tile?

nicolasvasilache (approved):

PR #341, opened 2022-03-09
Iterators: Replace dummy test with first iterator-style MLIR code.

ingomueller-net:
This PR creates two `.mlir` test files:

1. `simple-plan-input.mlir`, which represents a program in a to-be-created "iterators" dialect,
2. `simple-plan-target.mlir`, which is what I currently imagine the previous one should be lowered to.

The second file can be lowered to LLVM IR, linked to the runtime library, and executed.


experimental/iterators/test/simple-plan-input.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: mlir-opt -allow-unregistered-dialect %s
```
nicolasvasilache, line 1: This seems premature in this revision?

experimental/iterators/include/iterators/Utils/RuntimeUtils.h
```diff
@@ -23,7 +23,13 @@
 //===---------------------------------------------------------------------===//
 // Exposed C API.
 //===---------------------------------------------------------------------===//
-extern "C" RUNTIME_UTILS_EXPORT int64_t oneWay();
-extern "C" RUNTIME_UTILS_EXPORT void otherWay(int64_t);
+
+extern "C" RUNTIME_UTILS_EXPORT int8_t *makeSampleInputOperator();
+
+// XXX(ingomueller): resolve name clash with namespace
```
nicolasvasilache, line 29: Please do it now, we don't want to carry this "disambiguation by naming" forward.

experimental/iterators/lib/Utils/RuntimeUtils.cpp
```diff
@@ -1,18 +1,51 @@
...
-extern "C" int64_t oneWay() {
-  void *p = reinterpret_cast<void *>(0xDEADBEEF);
-  fprintf(stdout, "Create a dummy pointer %p\n", p);
-  return reinterpret_cast<int64_t>(p);
+
+extern "C" int8_t *makeSampleInputOperator() {
+  std::vector<int32_t> sampleInput = {1, 2, 3};
+  using ScanOperatorType = decltype(makeColumnScanOperator(sampleInput));
+  auto const ptr =
+      new ScanOperatorType(sampleInput); // XXX(ingomueller) this is a leak!
```
nicolasvasilache, line 20: You def. want a `free` counterpart operation that you call explicitly from MLIR.
I see the reliance on `decltype` and the need to pass that type information across make/free boundaries.

For now I think I would just go with `makeSampleInputVectorInt32Operator` and `SampleInputVectorInt32Operator` and do the proper cast in the free.

experimental/iterators/test/simple-plan-target.mlir
```diff
@@ -0,0 +1,16 @@
...
+func private @makeSampleInputOperator() -> !llvm.ptr<i8>
+func private @makeReduceOperator2(!llvm.ptr<i8>) -> !llvm.ptr<i8>
+func private @comsumeAndPrint(!llvm.ptr<i8>) -> ()
+
+func @main() {
+  %input = call @makeSampleInputOperator() : () -> (!llvm.ptr<i8>)
+  %reduce = call @makeReduceOperator2(%input) : (!llvm.ptr<i8>) -> (!llvm.ptr<i8>)
+  // CHECK: (6)
+  call @comsumeAndPrint(%reduce) : (!llvm.ptr<i8>) -> ()
+  return
```
nicolasvasilache, line 15: you should have some `free` here

experimental/iterators/lib/Utils/RuntimeUtils.cpp
```diff
@@ -1,18 +1,51 @@
...
+  return std::make_tuple(std::get<0>(t1) + std::get<0>(t2));
+};
+
+extern "C" int8_t *makeReduceOperator2(int8_t *const upstream) {
+  using ScanOperatorType =
+      decltype(makeColumnScanOperator(std::declval<std::vector<int32_t>>()));
+  using ReduceOperatorType =
+      decltype(makeReduceOperator(std::declval<ScanOperatorType *>(), sum));
+  // XXX(ingomueller) this is a leak!
+  auto const ptr = new ReduceOperatorType(
```
nicolasvasilache, line 36: same, you want a free that typecasts to the right thing on the MLIR side.

nicolasvasilache (approved):

experimental/iterators/test/simple-plan-input.mlir
ingomueller-net: I thought checking that the file has MLIR syntax could be useful, no?

nicolasvasilache: In practice this is only testing that the core MLIR custom parser and printer form works. 
This is already well tested upstream.

experimental/iterators/include/iterators/Utils/RuntimeUtils.h
ingomueller-net: Fixed with #344 and latest commit, which also uses "poor-man's namespaces" (prefixing the C-functions with `iterators*`).

PR #342, opened 2022-03-09
Port tiling to linalg_ext::TileOp to the transform dialect.

nicolasvasilache:
In the process, retire the TileToTileOp pass.

### DIFF
--- a/lib/Dialects/LinalgExt/Transforms/TilingToTileOp.cpp
+++ b/lib/Dialects/LinalgExt/Transforms/TilingToTileOp.cpp
@@ -5,9 +5,11 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 //===----------------------------------------------------------------------===//
+
 #include "Dialects/LinalgExt/LinalgExtOps.h"
 #include "Dialects/LinalgExt/PassDetail.h"
 #include "Dialects/LinalgExt/Passes.h"
+#include "Dialects/LinalgExt/Transforms/Transforms.h"
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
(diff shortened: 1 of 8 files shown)


lib/Dialects/LinalgExt/Transforms/TilingToTileOp.cpp
```diff
@@ -52,101 +59,49 @@ static SmallVector<Value> tileToTileOp(PatternRewriter &rewriter,
...
+  SmallVector<Value> tileSizes =
+      options.tileSizeComputationFunction(rewriter, op);
+  int64_t dim = -1;
+  for (auto en : llvm::enumerate(tileSizes)) {
+    Optional<int64_t> maybeTileSize = getConstantIntValue(en.value());
+    if (maybeTileSize && *maybeTileSize == 0)
+      continue;
+    if (maybeTileSize && *maybeTileSize < 0)
+      return rewriter.notifyMatchFailure(op, "Negative tile size");
+    if (dim >= 0)
```
gysit, line 85: sorry the comment should be here. This checks may be implemented by a transform op verifier.

nicolasvasilache: Right there are a few tradeoffs wrt matchers vs transform op verifier vs matchAndRewrite.
I'll add a TODO because I do not know what the end state should look like; if we put everything in the verifier, it will be missing here.

@Mogball in case something comes to mind.

gysit: the existing ops duplicate some of that verification logic since the patterns are not exclusively used via transform dialect. A comment/TODO is fine...

PR #348, opened 2022-03-10
Bump llvm

gysit:
Update llvm version to af98b0af.

Disable two tests since tracking currently fails due to a pattern that does not use replace op (https://reviews.llvm.org/D121369).


nicolasvasilache:
this large stack looks very fishy

PR #352, opened 2022-03-10
Add python support for linalg_ext transforms that use the transform dialect.

nicolasvasilache:
This requires some unwelcome dependencies of the transform dialect on the linalg_ext and async dialects.
These dependencies should be cleaned up in the future when we have a proper transform interface.

### DIFF
--- a/python/sandbox/dialects/_linalg_transform_ops_ext.py
+++ b/python/sandbox/dialects/_linalg_transform_ops_ext.py
@@ -119,6 +119,7 @@ class LowerToLLVMOp:
                enable_arm_sve: BoolArg = None,
                enable_amx: BoolArg = None,
                enable_x86vector: BoolArg = None,
+               enable_async: BoolArg = None,
                loc=None,
                ip=None):
     super().__init__(_ensure_bool_attr(reassociate_fp_reductions, False),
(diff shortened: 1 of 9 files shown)


python/sandbox/dialects/_linalg_transform_ops_ext.py
```diff
@@ -336,3 +338,70 @@ class SequenceOp:
...
+
+  def __init__(self,
+               target: Union[ir.Value, ir.Operation, ir.OpView],
+               *,
+               sizes: IntListArg = None,
+               loc=None,
+               ip=None):
+    sizes = _ensure_array_attr(sizes, [])
+    operation_type = pdl.OperationType.get()
+    super().__init__(operation_type, target, sizes, loc=loc, ip=ip)
```
gysit, line 359: _ensure_array_attr(sizes, [])?

gysit (approved):

python/sandbox/dialects/_linalg_transform_ops_ext.py
nicolasvasilache: if you are suggesting inlining the expression, I kept it outlined for consistency with the rest.

PR #367, opened 2022-03-14
Use global FileCheck variables

ftynse:
Modern FileCheck resets variables after each CHECK-LABEL, so the test
could be broken by using an anyhow recent version of FileCheck. Use
$-prefixed global variables that are not reset instead.


gysit (approved):
Thanks for fixing!

PR #372, opened 2022-03-15
WIP - Enable timing for Transform Dialect

chelini:


chelini:
Hi All, 
I would like to add timing support for the Transform dialect using `-mlir-timing`. Currently, running:
`mlir-proto-opt -mlir-timing -linalg-interp-transforms some.file.mlir` returns the time spent it the `InterpreterPass` I would like to have more fine-grained information (i.e., see how much time is spent in the `BufferizeOp` or in other transform Ops).

My current proposal adds a `DefaultTimingManager` to the `TransformState` and a bit of logic in each op to set up the timer (e.g., bufferzieOp).

Questions:
1. Is this something useful to have?
2. Do you know better ways to achieve this?

Thanks a lot!

ftynse:
This may be something useful to have, but not in the way it is currently implemented. Only two transformations rely on passes, we just haven't had time to clean them up yet. The plan is not to have any pass managers inside. So you just need to add timing facility to the interpreter itself, around the call to interface `apply` function.

include/Dialects/LinalgTransform/TransformOpInterface.h
```diff
@@ -206,6 +207,9 @@ class TransformState {
   /// Extensions attached to the TransformState, identified by the TypeID of
   /// their type. Only one extension of any given type is allowed.
   DenseMap<TypeID, std::unique_ptr<Extension>> extensions;
+
+public:
```
ftynse, line 211: I don't see a reason for this to be public.

lib/Dialects/LinalgTransform/IR/TransformOpInterface.cpp
```diff
@@ -104,9 +104,12 @@ void transform::TransformState::updatePayloadOps(
 
 LogicalResult
 transform::TransformState::applyTransform(TransformOpInterface transform) {
+  TimingScope timing = tm.getRootScope();
+  TimingScope transformOp = timing.nest(transform->getName().getStringRef());
```
ftynse, line 108: What happens when there's more than one op with the same name (double tiling is a thing, so is having multiple sequences based on different matches)?

ftynse (changes requested):
It's slightly closer, but not there yet. I really don't think this should interfere with the interface, which should be kept as clear as possible. Some place like https://github.com/google/iree-llvm-sandbox/blob/d868c9df3ccc1cf0de0b62c22d4e4a4d70f6bc89/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp#L166 would be more appropriate. It would also let you report timing connected to specific locations of ops, rather than just their name. Note that the interpreter also runs pseudo-canonicalization, CSE and some other enabler transformations between ops, which take substantial time and are unaccounted for by the current approach. Also consider https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Support/Timing.h#L334-L339.

chelini:
> It's slightly closer, but not there yet. I really don't think this should interfere with the interface, which should be kept as clear as possible. Some place like
> 
> https://github.com/google/iree-llvm-sandbox/blob/d868c9df3ccc1cf0de0b62c22d4e4a4d70f6bc89/lib/Dialects/LinalgTransform/Transforms/TransformInterpreter.cpp#L166
> 
> would be more appropriate. It would also let you report timing connected to specific locations of ops, rather than just their name. Note that the interpreter also runs pseudo-canonicalization, CSE and some other enabler transformations between ops, which take substantial time and are unaccounted for by the current approach. Also consider https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Support/Timing.h#L334-L339.

Regarding the instantiation of the `DefaultTimeManager`. I read the documentation, but I did not understand *why* instantiating the object is discouraged. The `Timer` and `TimingScope` classes need a `TimingManager`; otherwise, they are tombstones. `TimingManger` allows passing some handles to measure time. The one I would like to have is the one implemented for `DefaultTimeManager` in `TimerImpl` and `TimerImpl`, for some reason, is not exposed in the interface.

Aside from that, a more inconvenience is the following: Ideally, I would like to nest my timings under the "top-level" `TimingScope` associated with the passManager (https://github.com/llvm/llvm-project/blob/14d1a582093951973a5b94e4aab4d91ef4a0b80d/mlir/lib/Pass/PassTiming.cpp#L145), but I cannot access it from the pass as far as I know. So currently, it seems that passes need to have their custom flag for timing (instead of using `mlir-timing`) if they want to provide more finer-grained execution reports.

include/Dialects/LinalgTransform/TransformOpInterface.h
chelini: right.

lib/Dialects/LinalgTransform/IR/TransformOpInterface.cpp
chelini: If they have the same name they end up in the same bucket. For example, the report for a double tiling will report only one 'tiling' entry where the value is the sum of the execution of the two tiling op. We can fix this by appending location information.

PR #377, opened 2022-03-17
Add print op to transformation dialect.

gysit:
- the commit revives the old PrintIR transform and with print after all etc.


matthias-springer (approved):

nicolasvasilache:
Rebased and landed in #385

PR #381, opened 2022-03-18
Rename drop-schedule-from-module to drop-schedule

nicolasvasilache:
This better reflects what the pass actually does.


chelini (approved):
Thanks.

PR #405, opened 2022-03-24
Support distribution of larger vector.reduction ops.

matthias-springer:
Vector reductions greater than 32xf32 can be distributed. Still needs to
be a multiple of 32.


lib/Dialect/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -673,6 +674,12 @@ void mlir::vector_ext::populatePropagateVectorDistributionPatterns(
   pattern.add<WarpOpElementwise, WarpOpTransferRead, WarpOpDeadResult,
               WarpOpReduction, WarpOpBroadcast, WarpOpForwardOperand,
               WarpOpScfForOp>(pattern.getContext());
+  vector::populateVectorUnrollPatterns(
```
nicolasvasilache, line 677: The warm feeling of composition :)

nicolasvasilache (approved):

PR #410, opened 2022-03-24
Split peeling from TileOp transform

matthias-springer:
* `peel` is no longer an attribute of `TileOp`.
* `TileOp` returns the tiled LinalgOp and each loop as results.
* The parser/printer of `TileOp` are hand-written to avoid having to specify the loop handle results in the type signature. Instead, the number of results is inferred from the size of the `sizes` ArrayAttr in the parser.
* Add a test case for tiling + peeling.

### DIFF
--- a/lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
+++ b/lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
@@ -151,27 +151,54 @@ LogicalResult transform::MatchOp::apply(TransformResults &results,
 // TileOp
 //===---------------------------------------------------------------------===//
 
-FailureOr<LinalgOp> transform::TileOp::applyToOne(LinalgOp target) {
+LogicalResult transform::TileOp::apply(TransformResults &transformResults,
+                                       TransformState &state) {
   LinalgTilingOptions tilingOptions;
   SmallVector<int64_t> tileSizes = extractI64Array(sizes());
+  size_t numExpectedLoops = 0;
+  for (int64_t i : tileSizes)
+    if (i)
+      ++numExpectedLoops;
+
   // "scalarize_dyn_dims" actually sets the same lambda as the tile sizes and
   // asserts that it is not already set.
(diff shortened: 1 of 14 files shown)


lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
```diff
@@ -151,27 +151,46 @@ LogicalResult transform::MatchOp::apply(TransformResults &results,
...
+        functional::applyAt(linalgOp, functionalTile);
+    if (failed(tiled))
+      return failure();
+
+    tiledLinalgOps.push_back(tiled->op);
+    assert(tiled->loops.size() == tileSizes.size() &&
+           "number of generated loops not same as requested");
+    for (unsigned int i = 0; i < tileSizes.size(); ++i) {
+      loops[i].push_back(tiled->loops[i]);
+    }
```
ftynse, line 186: `llvm::append_range(loops[i], tiled->loops[i])`

lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
```diff
@@ -151,27 +151,46 @@ LogicalResult transform::MatchOp::apply(TransformResults &results,
...
+  }
+
+  transformResults.set(tiled_linalg_op().cast<OpResult>(), tiledLinalgOps);
+  for (unsigned int i = 0; i < tileSizes.size(); ++i) {
```
ftynse, line 190: Don't evaluate `size()` on every iteration: `for (unsigned i = 0, e = tileSizes.size(); i < e; ++i)`

ftynse (approved):
Nice!

lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
matthias-springer: I think I actually need `push_back` here because `tiled->loops[i]` is not a container.

nicolasvasilache:
Integrated in #419 after LLVM bump and synchronization with IREE

PR #411, opened 2022-03-25
Update the sandbox to use the new padding options (WIP).

gysit:
Test cl to verify the changes in https://reviews.llvm.org/D122309 work well with the transform dialect.


nicolasvasilache:
Integrated in #418

PR #412, opened 2022-03-27
Iree integrate

nicolasvasilache:


nicolasvasilache:
This revision rolls back the last 4 commits, resynchronizes properly with iree-dialects and replays them back.
As a result we are back to sync'ed with some version of IREE.

gysit (approved):

PR #415, opened 2022-03-29
SQL: prototypes for dialects, frontend and lowering

webmiche:
This PR adds the following:
1. a dialect that mirrors the ibis datastructures
2. a dialect for relational algebra
3. a frontend for ibis to ibis dialect conversion
4. a lowering for a simple query from the ibis dialect to the relational dialect

The only really working query currently is `table.filter(table['a'] == 'AS')`. This is one of the examples of the ibis user guide.

### DIFF
--- a/.github/workflows/testSQLDialect.yml
+++ b/.github/workflows/testSQLDialect.yml
@@ -1,0 +1,39 @@
+name: SQLDialectTest
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Test SQL Dialect
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+
+    - name: Checkout project
+      uses: actions/checkout@v2
+      with:
+        path: sandbox
+
+    - name: Install Python depends
+      run: |
+        python -m pip install -r ${GITHUB_WORKSPACE}/sandbox/experimental/sql/requirements.txt
+
+    - name: Check format with yapf
+      run: |
+        # Returns en error code if the files are not formatted
+        yapf --diff -r ${GITHUB_WORKSPACE}/sandbox/experimental/sql
+
+    - name: Test
+      run: |
+        cd ${GITHUB_WORKSPACE}/sandbox
+        lit -v ${GITHUB_WORKSPACE}/sandbox/experimental/sql/test
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -1,0 +1,198 @@
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from dataclasses import dataclass
+from typing import Any
+from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
+from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr
+from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder
+
+# This file define a dialect that mirrors the internal structure of ibis.
+# The file contains an `Attribute` for all types encountered in the simple
+# query. Types are defined as being in defined `ibis/expr/types` in the
+# ibis source repo. All other nodes encountered in this translation are
+# encoded as Operations.
+
+
+@irdl_attr_definition
+class TableExpr(ParametrizedAttribute):
+  name = "ibis.table_expr"
+
+
+@irdl_attr_definition
+class Column(ParametrizedAttribute):
+  name = "ibis.column"
+
+
+@irdl_attr_definition
+class StringColumn(Column):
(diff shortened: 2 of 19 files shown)


ingomueller-net (changes requested):
I have done a first pass and added a few comments, mainly about formatting and similar. I am missing a few things to do a more in-depth review:

1. How can I run this code? Maybe add a short guide for setting up and running things to the README. Maybe a simple test case would also be good.
2. What's the overall rationale of the different files? I can guess many things but a few comments, possibly with links to resources from Ibis and xDSL would help a lot.

**Update:** I promised that the reviews aren't meant to stop you from merging things but selected "request changes." Unfortunately, I can't change that now, so just ignore it. I won't select that review decision in the future... ;)

webmiche:
Thanks for the configs, I used them to format properly. As for your questions:

1. I have a driver, but it relies on a sqlite database that ibis distributes with their tutorial. I am uncertain how licensing works with respect to that since the database is not actually mine, so I decided to not upload it currently. I think we should spin this into a discussion on which backend we actually want to support and then provide running testcases for that.
2. I mainly decided to split different dialects and different passes in the current pipeline. I will add some comments to each file what it does exactly.

I think it makes sense to keep this as a PR until we have a running testcase in.

ingomueller-net:
Cool, thanks for the explanation!

For a long-term solution for input data, we might want to look how the pandas backend does things. From the [documentation](https://ibis-project.org/backends/Pandas/#direct-execution-backends):

```python
import ibis
ibis.pandas.connect({"t": pd.DataFrame({"a": [1, 2, 3]})})
```

That should allow to produce test data easily. Also, pandas data frames can be wrapped into structs of arrays without copies, i.e., by using the original allocations of the different columns, so we could essentially process pandas data frames with our code directly. (That's what CVM does.)

About the SQLite: is it possible to split the driver into code that you wrote, which you can commit, and external code, which one has to download (and patch)? How about other dependencies such as xDSL? A `requirements.txt` shouldn't be a problem with licenses; the same is probably true for git submodules. The objective of all of that is to enable us to run your code...

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2021 The IREE Authors
```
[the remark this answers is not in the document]
nicolasvasilache: we can drop it here and everywhere.

webmiche:
I included a testcase that uses Pandas instead of Sqlite. Additionally, I added comments to the files explaining what they contain/do. I will now setup the requirements and take a look at CI integration.

experimental/sql/tools/rel_opt.py
```diff
@@ -0,0 +1,98 @@
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
```
webmiche, line 1: This file is supposed to basically mirror the functionality of MlirOpt, so I want to run it directly from the command line with `./tools/rel_opt.py ...`. However, this requires me to have `#!/usrbin/env python3` as the first line. This conflicts with having the license as the first lines. Is there a way to circumvent this? Maybe @nicolasvasilache has an input?

webmiche:
I added filecheck tests and a file that mirrors MlirOpt basically. xDSL has nice support for this since the last version.

ingomueller-net:
I went over the README. Installing and running went exactly like described. Nice!

I think that this is good enough for a first iteration. Ready to merge from my side.

ingomueller-net (approved):

ingomueller-net:
LGTM (looks good to me), but I think @nicolasvasilache should at least check the CI integration.

.github/workflows/buildAndTest.yml
```diff
@@ -59,4 +59,4 @@ jobs:
     - name: Test
       run: |
         cd ${GITHUB_WORKSPACE}/sandbox
-        python run_tests.py --iterators-tests
+        python run_tests.py --sql-dialect-tests --iterators-tests
```
ingomueller-net, line 62: Nit: trailing space.

.github/workflows/testSQLDialect.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    - name: Check format with yapf 
+      run: |
+        # Returns en error code if the files are not formatted
+        yapf --diff -r ${GITHUB_WORKSPACE}/sandbox/experimental/sql
+
+    - name: Test
+      run: |
+        cd ${GITHUB_WORKSPACE}/sandbox
+        lit -v ${GITHUB_WORKSPACE}/sandbox/experimental/sql/test
```
ingomueller-net, line 39: Nit: There are a few trailing white spaces in this file...

experimental/sql/simple.py
```diff
@@ -0,0 +1,44 @@
...
+from src.ibis_to_rel import ibis_dialect_to_relational
+
+connection = ibis.pandas.connect({"t": pd.DataFrame({"a": ["AS", "EU", "NA"]})})
+
+# Get the table.
+table = connection.table('t')
+
+# Define the query.
+query = table.filter(table['a'] == 'AS')
+print('Pandas AST: ' + '-' * 60)
```
ingomueller-net, line 25:
```suggestion
print('Ibis AST: ' + '-' * 60)
```

ingomueller-net (commented):
LGTM. Two minor things in the comments.

ingomueller-net:
Looks still good to me! A final stamp by @nicolasvasilache would be great, in particular for the CI targets.

nicolasvasilache:
I see it does a lot of stuff at once with XDSL that I am still not familiar with. 
In particular, this single PR introduces 2 dialects with 2 classes of tests, a frontend and a conversion.
Can we split all this into smaller CLs that do 1 thing at a time ?
I think this would be a few simple git/github commands to stage the PRs and a little more documentation. 
This way it will be easier to review without getting lost between things that are really simple and things that require more attention

PR #420, opened 2022-04-01
Add support for non-string padding values.

gysit:
Previously only string padding values, that later on would be parsed to the element type of the padded operation, were supported. After the change, it is possible to hand in a floating point attribute directly.

### DIFF
--- a/lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
+++ b/lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
@@ -86,16 +86,6 @@ static SmallVector<int64_t> extractI64Array(ArrayAttr attr) {
   return result;
 }
 
-/// Extracts a vector of StringRef from an array attribute. Asserts if the
-/// attribute contains values other than StringRefs.
-static SmallVector<StringRef> extractStringArray(ArrayAttr attr) {
-  SmallVector<StringRef> result;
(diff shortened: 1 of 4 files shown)


lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
```diff
@@ -334,14 +324,26 @@ FailureOr<LinalgOp> transform::PadOp::applyToOne(LinalgOp target) {
...
+    if (auto stringAttr = attr.dyn_cast<StringAttr>()) {
+      paddingValues.push_back(
+          parseAttribute(attr.cast<StringAttr>(), elementType));
+      if (!paddingValues.back()) {
+        return target->emitOpError("expects a padding value ")
+               << std::get<0>(it) << " that parses to " << elementType;
+      }
+      continue;
+    }
+    // Othwise, add the attribute directly.
```
chelini, line 341: minor typo

gysit: Thanks, should be fixed!

nicolasvasilache:
This has landed in IREE : https://github.com/google/iree/commit/08dfc09e181624815c01da12d7191017e90c4621.

Closing here as it will be mirrored on the next IREE -> sandbox pull.

PR #423, opened 2022-04-01
Add peeling support to fusion.

gysit:
Adapt FuseOp in the same way https://github.com/google/iree-llvm-sandbox/pull/419 adapted TileOp. The FuseOp now returns a handles to the generated tile loops, which then may be used to peel of loops from the fused loop nest.


nicolasvasilache:
This has landed in IREE : https://github.com/google/iree/commit/6eac24d15ae44f6a400da6c97989861787851be0

Closing here as it will be mirrored on the next IREE -> sandbox pull.

PR #432, opened 2022-04-05
Add integration test for warp op to scf.if lowering

matthias-springer:
Instead of allocation shared memory buffers with `memref.alloc`,
generate a global. Memory space 3 indicates shared memory.

### DIFF
--- a/lib/Dialect/VectorExt/Transform/VectorWarpUtils.cpp
+++ b/lib/Dialect/VectorExt/Transform/VectorWarpUtils.cpp
@@ -25,6 +25,8 @@ using namespace mlir::vector_ext;
 
 static const unsigned kWarpSize = 32;
 
+static const int64_t kSharedMemorySpace = 3;
+
 // Clones `op` into a new operations that takes `operands` and returns
 // `resultTypes`.
 static Operation *cloneOpWithOperandsAndTypes(OpBuilder &builder, Location loc,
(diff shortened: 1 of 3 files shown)


lib/Dialect/VectorExt/Transform/VectorWarpUtils.cpp
```diff
@@ -720,15 +722,50 @@ static LogicalResult rewriteWarpOpToScfFor(RewriterBase &rewriter,
...
     } else {
-      return rewriter.create<memref::AllocOp>(loc, MemRefType::get({1}, type));
+      memrefType = MemRefType::get({1}, type, {}, kSharedMemorySpace);
     }
+
+    // Get symbol table holding all shared memory globals.
+    ModuleOp moduleOp = warpOp->getParentOfType<ModuleOp>();
+    SymbolTable symbolTable(moduleOp);
+
+    // Create a pretty name. Make sure that the name is unique.
```
matthias-springer, line 739: Note: This is copied from BufferUtils.cpp. Could be a helper in MemrefOps.

ThomasRaoux: I'm pretty sure the names doesn't have to be unique at this point and we don't need all this. I would skip all this logic.
Look at how iree does it:
https://github.com/google/iree/blob/45fc889981a3b3b88bbc11066bcfcf92f91d7c77/iree/compiler/Codegen/LLVMGPU/ConvertToLLVM.cpp#L149

test/Integration/Dialect/VectorExt/GPU/test-warp.mlir
```diff
@@ -9,6 +9,32 @@
...
+// RUN: mlir-proto-opt %s -test-vector-warp-distribute="hoist-uniform distribute-transfer-write rewrite-warp-ops-to-scf-if" -canonicalize | \
+// RUN: mlir-opt -convert-scf-to-cf -convert-cf-to-llvm -convert-vector-to-llvm -convert-arith-to-llvm \
+// RUN:  -gpu-kernel-outlining \
+// RUN:  -pass-pipeline='gpu.module(strip-debuginfo,convert-gpu-to-nvvm,reconcile-unrealized-casts,gpu-to-cubin)' \
+// RUN:  -gpu-to-llvm -reconcile-unrealized-casts |\
+// RUN: mlir-cpu-runner -e main -entry-point-result=void \
+// RUN:   -shared-libs=%mlir_runner_utils_dir/libmlir_cuda_runtime%shlibext \
+// RUN:   -shared-libs=%mlir_runner_utils_dir/libmlir_c_runner_utils%shlibext \
+// RUN:   -shared-libs=%mlir_runner_utils_dir/libmlir_runner_utils%shlibext | \
+// RUN: FileCheck %s
```
ThomasRaoux, line 36: Awesome!

ThomasRaoux (approved):

lib/Dialect/VectorExt/Transform/VectorWarpUtils.cpp
matthias-springer: You're right, `SymbolTable::insert` already makes the symbol name unique. Dropped that part.

PR #434, opened 2022-04-07
SQL: ibis dialect + tests (+ background)

webmiche:
This PR adds the ibis dialect and a simple roundtrip testcase, that shows how an example IR can look like. The ibis dialect mirrors the structure of the [ibis](https://ibis-project.org/docs/dev/) internal data structures. There are `Attribute`s for all types encountered in the simple query. Types are defined as being in defined `ibis/expr/types` in the ibis source repo. All other nodes encountered in this query are encoded as Operations.

### DIFF
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -1,0 +1,259 @@
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from dataclasses import dataclass
+from typing import Any
+from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
+from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr
+from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder
+
+# This file defines a dialect that mirrors the internal data structures of the
+# ibis framework (https://ibis-project.org/docs/dev/). In ibis, there exists an
+# additional layer of abstractions called `expressions`. Each operation in a
+# query references the expressions of its subqueries and is itself reference by
+# an expression. This is a remainder of the ibis UI that allows
+# printing/executing/... of expressions. Notice that the additional layer of
+# indirection can make ibis queries DAGs rather than trees. This only happens if
+# the user enters the query in a specific way, so this is also only a left-over
+# of the API. This dialect represents a cleaned up version of queries without
+# the expression layer for each node. In this version, all queries are trees, so
+# every part of the query is modeled as an operation with regions for its
+# respective subqueries. Hence, every operation is a single region terminator
+# and there are no SSAValues in this dialect.
+#
+# DISCLAIMER: This design tries to avoid premature complexity and is subject to
+# change. In particular, comples predicates might involve operations that are
+# not single region terminators.
+
+
+@irdl_attr_definition
+class DataType(ParametrizedAttribute):
+  """
+  Models an ibis datatype.
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/datatypes.py#L34
+  """
+  name = "ibis.datatype"
+
+
+@irdl_attr_definition
+class int32(DataType):
+  """
+  Models the ibis int32 type.
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/datatypes.py#L294
+
+  Example:
+
+  ```
+  !ibis.int32
+  ```
+  """
+  name = "ibis.int32"
+
+
+@irdl_attr_definition
+class String(DataType):
+  """
+  Models the ibis string type. The Parameter `nullable` defines whether the
+  string can be null.
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/datatypes.py#L176
+
+  Example:
+
+  ```
+  !ibis.string<0 : !i1>
+  ```
+  """
+  name = "ibis.string"
+
+  nullable = ParameterDef(IntegerAttr)
+
+  @builder
+  @staticmethod
+  def get(val: int) -> 'String':
+    return String([IntegerAttr.from_int_and_width(val, 1)])
+
+
+@irdl_op_definition
+class TableColumn(Operation):
+  """
+  References a specific column with name `col_name` from `table`.
+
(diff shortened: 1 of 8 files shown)


experimental/sql/test/ibis_dialect_tests/simple_query.xdsl
```diff
@@ -0,0 +1,52 @@
+// RUN: rel_opt.py %s | filecheck %s
+
+// Query: table.filter(table['a'] == 'AS')
+module() {
```
nicolasvasilache, line 4: Thanks for splitting this up !

Upon discussion with Ingo, it appears that it is not very clear to me what purpose the SSA values serve in the ibis dialect.

I'd suggest making everything a terminator to encode a tree.

Later we can rewrite the tree to SSA form by looking at symbols (e.g. table name) which I understand carry the reuse semantics ?

If we were to only use terminators everywhere, the IR would look like:

```suggestion
module() {
  ibis.selection() {
    // selection input
    ibis.pandas_table() ["table_name" = "t"] {
      ibis.schema_element() ["elt_name" = "a", "elt_type" = !ibis.String<1 : !i1>]
    }
  } {
    // selection predicate
    ibis.equals() {
      // predicate column
      ibis.table_column() ["col_name" = "a"] {
        ibis.pandas_table() ["table_name" = "t"] {
          ibis.schema_element() ["elt_name" = "a", "elt_type" = !ibis.String<1 : !i1>]
        }
      }
    } {
      // predicate literal
      ibis.literal() ["val" = "AS", "type" = !ibis.String<1 : !i1>]
    }
  }
}
```

What do you think of this modeling ?

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -0,0 +1,247 @@
...
+# structures of the ibis framework (https://ibis-project.org/docs/dev/).
+# DataTypes are defined as Attributes, other important nodes are
+# operations. Everything that is an `expresssion` in ibis is not
+# modelled.
+
+
+@irdl_attr_definition
+class DataType(ParametrizedAttribute):
+  """
+  Models an ibis datatype.
```
ingomueller-net, line 21: There is a trailing white space in this and several other lines. The simplest is maybe to set `files.trimTrailingWhitespace": true` in `settings.json`.

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -0,0 +1,247 @@
...
+from typing import Any
+from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
+from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr
+from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder
+
+# This file defines a dialect that mirrors the internal data
+# structures of the ibis framework (https://ibis-project.org/docs/dev/).
+# DataTypes are defined as Attributes, other important nodes are
+# operations. Everything that is an `expresssion` in ibis is not
+# modelled.
```
ingomueller-net, line 15: I think I would elaborate a bit here:
* Describe the fact that the ibis query is sort of a tree but not really and that that's an artifact of the UI.
* Describe the expression vs operation weirdness in Ibis and that that's an artifact of the UI as well.
* Say that this dialect should represent a cleaned-up version of that tree and what exactly it cleans up.

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -0,0 +1,247 @@
...
+  @builder
+  @staticmethod
+  def get(val: int) -> 'String':
+    return String([IntegerAttr.from_int_and_width(val, 1)])
+
+
+@irdl_op_definition
+class TableColumn(Operation):
+  """
+  Gets a specific column with name `col_name` from `table`.
```
ingomueller-net, line 71: Maybe `Refers to` or `References` instead of `Gets`?

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -0,0 +1,247 @@
...
+  """
+  name = "ibis.datatype"
+
+
+@irdl_attr_definition
+class int32(DataType):
+  """
+  Models the ibis int32 type. 
+
+  https://github.com/ibis-project/ibis/blob/master/ibis/expr/datatypes/core.py#L346
```
ingomueller-net, line 33: Maybe its safer to link to the file at a particular tag (since file names and line numbers may change).

experimental/sql/dialects/ibis_dialect.py
ingomueller-net: Also, I can't really match "other important nodes are operations" with the structure of the IR... Aren't they (single-op) regions?

experimental/sql/tools/rel_opt.py
```diff
@@ -0,0 +1,65 @@
...
+#from src.ibis_to_rel import ibis_dialect_to_relational
+from dialects.ibis_dialect import Ibis
+#from dialects.relational_dialect import Relational
+
+
+class RelOptMain(xDSLOptMain):
+
+  #def register_all_passes(self):
+  #  self.available_passes[
+  #      'ibis-dialect-to-relational'] = ibis_dialect_to_relational
```
ingomueller-net, line 23: Remove the dead code of this file?

ingomueller-net (commented):
Removing the yields seems to have been a good idea -- the IR is much simpler! Removing the expressions is another step in the same direction; now, I think the IR looks really nice and intuitive.

Just a few minor things and some questions about the documentation. Otherwise: LGTM!

experimental/sql/dialects/ibis_dialect.py
nicolasvasilache: This, plus I'd also mention that we model a tree where every op is a single region terminator and without any SSA value in this dialect atm.

nicolasvasilache (approved):
Looks good once comments are addressed.

Also, tree in IR weirdness makes me look forward to SSA values.

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -0,0 +1,257 @@
...
+# subqueries and is itself reference by an expression. This is a
+# remainder of the ibis UI that allows printing/executing/... of
+# expressions. Notice that the additional layer of indirection can make
+# ibis queries DAGs rather than trees. This only happens if the user
+# enters the query in a specific way, so this is also only a left-over
+# of the API. This dialect represents a cleaned up version of queries
+# without the expression layer for each node. In this version, all
+# queries are trees, so every part of the query is modeled as an
+# operation with regions for its respective subqueries. Hence, every
+# operation is a single region terminator and there are no SSAValues
```
webmiche, line 24: The fact that every operation is a single region terminator might not hold throughout the entire lifetime of this dialect. Imagine there is a selection with multiple predicates (T.a = 5 and T.b = 2). We could model this by having two equals operations in the predicate region of the selection. This might be a problem for later on tough.

nicolasvasilache: yes, I would put your wording as a disclaimer, everything is subject to evolution, we just want to avoid premature complexity.

nicolasvasilache:
Once the CI turns green, feel free to land!

webmiche:
@nicolasvasilache the CI fails for a part that I did not touch. Can I merge anyway?

ingomueller-net:
Rerunning fixed the build. Merge away! :)

nicolasvasilache:
Thanks!

PR #437, opened 2022-04-07
Iterators: Rename IteratorType to StreamType

ingomueller-net:
I realized that the current operations we have don't *work on* iterators; they *are* iterators, and what they work on are *streams* (of elements of some type).


nicolasvasilache (approved):
Nice, thanks for the carification!

PR #438, opened 2022-04-08
Iterators: Add iterator state types and ops for Open/Next/Close.

ingomueller-net:


experimental/iterators/lib/Dialects/Iterators/IR/Iterators.cpp
```diff
@@ -31,13 +32,98 @@ void IteratorsDialect::initialize() {
...
+#include "iterators/Dialect/Iterators/IR/IteratorsTypeInterfaces.cpp.inc"
+
 //===----------------------------------------------------------------------===//
 // Iterators operations
 //===----------------------------------------------------------------------===//
 
 #define GET_OP_CLASSES
 #include "iterators/Dialect/Iterators/IR/IteratorsOps.cpp.inc"
 
+LogicalResult OpenOp::verify() {
```
nicolasvasilache, line 49: Some of these verifiers could be removed if you adopted standard Tablegen predicates on types (you may need to create your own CPred-based stuff).
The way we go for such cases is to test 1 case and then all the rest are considered auto-generated and now correct by construction.

OTOH we also need to provision for the future; if you feel the verification rules are going to change in the future and will require visting the type when it becomes nested then this is fine for now and until we invent the generic rules that we will want to apply.

experimental/iterators/test/Dialect/Iterators/iterator-states-invalid.mlir
```diff
@@ -0,0 +1,75 @@
...
+  return
+}
+
+// -----
+
+!sampleInputState = type !iterators.sampleinputstate<i32>
+!reduceState = type !iterators.reducestate<!sampleInputState>
+func @testOpenStateTypeMismatch() {
+  %initialState = "iterators.createSampleInputState"() : () -> !sampleInputState
+  // expected-error@+1 {{Opening iterator of type '!iterators.sampleinputstate<i32>' should return the same type but returns '!iterators.reducestate<!iterators.sampleinputstate<i32>>'}}
```
nicolasvasilache, line 39: This message and other similar ones in the file are generally considered "standard Tablgen do not reinvent".
Look around `SameOperandsElementType` in OpBase.td and see if you find something that suits you.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsTypes.td
```diff
@@ -25,4 +26,36 @@ def Iterators_Stream : Iterators_Type<"Stream", "stream"> {
...
 }
 
+//===----------------------------------------------------------------------===//
+// State of low-level iterators
+//===----------------------------------------------------------------------===//
+
+class Iterators_Iterator<string name, string typeMnemonic>
+    : Iterators_Type<name, typeMnemonic,
+                     [DeclareTypeInterfaceMethods<Iterators_IteratorInterface>]> {
+  let summary = "State of a Volcano-style iterator";
```
nicolasvasilache, line 36: Since we're playing games with visibility vs opacity boundaries, I'd go as far as spelling out the type in C++ and being explicit about what is opaque and visible at this level of abstraction.

nicolasvasilache (commented):
Generally looks good.
There are some opportunities for simplification by adopting existing Tablegen abstractions (or extending them a bit) to avoid manually spelling out all cases of the verifiers + tests.

experimental/iterators/test/Dialect/Iterators/iterator-states-invalid.mlir
ingomueller-net: Aha, this is a super interesting pointer! `SameOperandsElementType` only works on [`ShapedType`](https://mlir.llvm.org/docs/Dialects/Builtin/#shapedtype-shapedtypeinterface) (due to how `getElementTypeOrSelf` is [implemented](https://github.com/llvm/llvm-project/blob/09c2b7c35af8c4bad39f03e9f60df8bd07323028/mlir/lib/IR/TypeUtilities.cpp#L24)), but actually maybe `Iterators` *are*/*should be* `ShapedType`s. If that were the case, we might get some inference facilities etc. for free, which may be a great example of why implementing database-style computations in MLIR is a good idea. Definitely, Iterators have an element type; they also have a rank, which is always `1` in all cases I am currently considering; and they have a shape, which is mostly unknown statically, but not always. I am not sure, though, if a `ShapedType` can decide only to work for rank `1`. What should my `clone` method do if it is invoked with a shape of rank `!= 1`?

ingomueller-net: The more I think about this, the more I think that `ShapedType` may model too broad a concept: it models things that have elements *and* things that have a shape. Currently, I only need the former. There are `ContainerType` and `ShapedContainerType`, which do make that distinction, but all useful predicates (`SameOperandsElementType`) are based on `ShapedType` instead of those.

ingomueller-net: Maybe I understand why: `ContainerType` is just a `class`, not an interface, so unlike `ShapedType` you can `dyn_cast` to it...

ingomueller-net: Maybe what I propose can't be done: I am thinking of `ShapedType` (an interface which forces elements and a shape) that *is a* `Container` (which only forces elements, where `ShapedType` inherits all element-related properties from `Container`. I am not sure, but I think that that kind of inheritance it not supported with `OpInterface`...

Bottom line: it seems like I need to replicate `SameOperandsElementType` and similar for my purposes. Not very satisfying, TBH...

experimental/iterators/lib/Dialects/Iterators/IR/Iterators.cpp
ingomueller-net: I have left the test cases for now. We can remove those that check off-the-shelf behavior before merging.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsTypes.td
ingomueller-net: I am not sure what this means. What exactly are the visibility boundaries? What does "spelling out the type in C++" mean?

I am imagining things like this: eventually we will generate the implementation of `OpenOp` etc for the concrete type of its argument (which implement `IteratorInterface`). That implementation may in turn use `OpenOp`/ `NextOp`/`CloseOp` on the states of its uptream iterator(s), whose concrete types it knows (which implement `IteratorInterface`) but whose implementation is not generated yet.

ingomueller-net:
@nicolasvasilache: Thanks for the review. The last three commits use some more built-in constructs for verification. I am not sure how to remove the remainder (see inline discussion).

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsTypes.td
ingomueller-net: Offline discussion: this is meant as an addition in the documentation.

experimental/iterators/test/Dialect/Iterators/iterator-states-invalid.mlir
ingomueller-net: Offline discussion: yes, I need to roll my own thing.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsTypes.td
ingomueller-net: Done with latest commit.

ingomueller-net:
Addressed with latest commit. I'll merge when the tests completed.

PR #440, opened 2022-04-11
SQL: Ibis frontend

webmiche:
This patch adds a first version of an ibis frontend. The frontend currently supports exactly the query `table.filter(table['a'] = 'AS').

The frontend consists of a basic visitorpattern with a `visit()`-function, that matches on the given node and dispatches to the proper function to build the operation for that node.

### DIFF
--- a/experimental/sql/src/ibis_frontend.py
+++ b/experimental/sql/src/ibis_frontend.py
@@ -1,0 +1,104 @@
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from dataclasses import dataclass
+from modulefinder import Module
+from unittest import result
+from xdsl.dialects.builtin import ArrayAttr, StringAttr, ModuleOp, IntegerAttr
+from xdsl.ir import Operation, MLContext, Region, Block
+from typing import List, Type, Optional
+
+import ibis
+import ibis.expr.types
+import ibis.expr.datatypes
+import ibis.expr.operations.relations as rels
+import ibis.expr.operations.generic as gen_types
+from ibis.expr.operations.logical import Equals as EQ
+import ibis.backends.pandas.client as PandasBackend
+
+import dialects.ibis_dialect as id
+
+# This file contains the translation from ibis to the ibis_dialect.
+# It is built in a visitor-style, as this is one of the best designs
+# to crawl trees in my experience. The entry-point is the `visit`-
+# function just below. It dispatches to the appropriate translator.
+
+
+@dataclass
+class NodeVisitor:
+  ctx: MLContext
+
+  def visit(self, node: ibis.expr.types.Expr) -> Operation:
+    if isinstance(node, ibis.expr.types.TableExpr):
+      return self.visit_TableExpr(node)
+    if isinstance(node, ibis.expr.types.StringColumn):
+      return self.visit_StringColumn(node)
+    if isinstance(node, ibis.expr.types.BooleanColumn):
+      return self.visit_BooleanColumn(node)
+    if isinstance(node, ibis.expr.types.StringScalar):
+      return self.visit_StringScalar(node)
+    raise KeyError(f"Unknown nodetype {type(node)}")
+
+  def convert_datatype(self, type_: ibis.expr.datatypes) -> id.DataType:
+    if isinstance(type_, ibis.expr.datatypes.String):
+      return id.String.get(1 if type_.nullable else 0)
+    if isinstance(type_, ibis.expr.datatypes.Int32):
+      return id.int32()
+    raise KeyError(f"Unknown datatype: {type(type_)}")
+
+  def visit_Schema(self, schema: ibis.expr.schema.Schema) -> Region:
+    ops = []
+    for n, t in zip(schema.names, schema.types):
+      ops.append(id.SchemaElement.get(n, self.convert_datatype(t)))
+    return Region.from_operation_list(ops)
+
+  def visit_TableExpr(self, table: ibis.expr.types.TableExpr) -> Operation:
+    op = table.op()
+    if isinstance(op, PandasBackend.PandasTable):
+      schema = self.visit_Schema(op.schema)
+      new_op = id.PandasTable.get(op.name, schema)
+      return new_op
+    if isinstance(op, rels.Selection):
+      table = Region.from_operation_list([self.visit(op.table)])
+      # TODO: handle mulitple predicates
+      predicates = Region.from_operation_list([self.visit(op.predicates[0])])
+      new_op = id.Selection.get(table, predicates)
+      return new_op
+    raise KeyError(f"Unknown tableExpr: {type(op)}")
+
+  def visit_StringColumn(
+      self, stringColumn: ibis.expr.types.StringColumn) -> Operation:
+    op = stringColumn.op()
+    if isinstance(op, gen_types.TableColumn):
+      table = Region.from_operation_list([self.visit(op.table)])
+      new_op = id.TableColumn.get(table, op.name)
+      return new_op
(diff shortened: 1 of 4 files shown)


experimental/sql/tools/rel_opt.py
```diff
@@ -4,12 +4,33 @@
...
+from src.ibis_frontend import ibis_to_xdsl
 from dialects.ibis_dialect import Ibis
 
 
 class RelOptMain(xDSLOptMain):
 
+  def register_all_frontends(self):
+    super().register_all_frontends()
+
+    def parse_ibis(f: IOBase):
```
webmiche, line 18: This is still something I am not very proud of and it comes from the fact that this rel_opt tool wants to read one file and build an IR from that. I think it might make sense to extensively test using `pytest` further down the line (AFAIU, the same framework ibis uses for their test-suite).

ingomueller-net: How about a while loop on `f.readline()` that filters out lines that `startswith(#)`? Similar to [this SO post](https://stackoverflow.com/a/6985913/651937) and potentially as a [generator](https://stackoverflow.com/a/519653/651937).

ingomueller-net: Actually, if you prefix the `CHECK*` with ` #`, `eval`ing the whole file could be fine?

ingomueller-net (commented):
LGTM modulo minor comments.

experimental/sql/tools/rel_opt.py
webmiche: Still, I don't like this approach a lot, since the query actually isn't generated form just the file, but from some local context within rel_opt as well...

experimental/sql/src/ibis_frontend.py
```diff
@@ -0,0 +1,104 @@
...
+import ibis.expr.operations.generic as gen_types
+from ibis.expr.operations.logical import Equals as EQ
+import ibis.backends.pandas.client as PandasBackend
+
+import dialects.ibis_dialect as id
+
+# This file contains the translation from ibis to the ibis_dialect.
+# It is built in a visitor-style, as this is one of the best designs
+# to crawl trees in my experience. The entry-point is the `visit`-
+# function just below. It dispatches to the appropriate translator.
```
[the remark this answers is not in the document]
webmiche: I wrote something on that.

experimental/sql/src/ibis_frontend.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@dataclass
+class NodeVisitor:
```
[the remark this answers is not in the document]
webmiche: I rewrote everything using this package. It feels way nicer to me, particularly since the top-level visit, that just calls the others, is not needed anymore. Plus, I like shifting as much complexity to others as possible.

experimental/sql/src/ibis_frontend.py
```diff
@@ -0,0 +1,104 @@
...
+    if isinstance(op, gen_types.TableColumn):
+      table = Region.from_operation_list([self.visit(op.table)])
+      new_op = id.TableColumn.get(table, op.name)
+      return new_op
+    raise Exception(f"Unknown stringcolumn: {type(op)}")
+
+  def visit_BooleanColumn(
+      self, boolColumn: ibis.expr.types.BooleanColumn) -> Operation:
+    op = boolColumn.op()
+    if isinstance(op, gen_types.TableColumn):
```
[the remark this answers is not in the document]
webmiche: I think this should be handled by the fact that we have a @dispatch(ibis.expr.types.BooleanColumn) just before.

experimental/sql/src/ibis_frontend.py
```diff
@@ -0,0 +1,104 @@
...
+
+  def visit_TableExpr(self, table: ibis.expr.types.TableExpr) -> Operation:
+    op = table.op()
+    if isinstance(op, PandasBackend.PandasTable):
+      schema = self.visit_Schema(op.schema)
+      new_op = id.PandasTable.get(op.name, schema)
+      return new_op
+    if isinstance(op, rels.Selection):
+      table = Region.from_operation_list([self.visit(op.table)])
+      # TODO: handle mulitple predicates
```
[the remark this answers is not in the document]
webmiche: Yes indeed, but we will still have to think about how this will be represented in the IR and how ibis represents this currently. The most straightforward way I see is just having multiple operations in the region, each defining a condition, and the actual predicate is the `and` of all of those. But how do we then represent `or`s?

This will probably be heavily influenced by ibis for this layer, but it is certainly also a question for lower layers.

experimental/sql/test/frontend/selection.ibis
```diff
@@ -0,0 +1,19 @@
+# RUN: rel_opt.py -f ibis %s | filecheck %s
+
+table.filter(table['a'] == 'AS')
+
+# CHECK: ibis.selection() {
```
nicolasvasilache, line 5: nit: can you add a few whitespace betwee `#` and `CHECK` to align the `:` ?

experimental/sql/tools/rel_opt.py
nicolasvasilache: I am not sure I follow 100% but maybe you could add a fake test_harness op that would take l19-28 as a python string and interpret it on the fly.

nicolasvasilache (approved):

experimental/sql/tools/rel_opt.py
ingomueller-net: Maybe the context should be part of the test case? It could still be shortened slightly by moving some of it to some `test_utils` package or similar.

```python
import ibis
import pandas as pd

connection = ibis.pandas.connect(
  {"t": pd.DataFrame({"a": ["AS", "EU", "NA"]})})
table = connection.table('t')
```

But maybe that doesn't combine well with `eval`? I think that the root cause of this whole issue is the fact that Ibis isn't a language but an API...

Eventually, I think that we want to call `ibis_to_xdsl` from our own Ibis backend. But writing that could involve a serious amount of plumbing, so I wouldn't do it right now.

I am inclined to say to kick the can down the road: live it as is for now and worry about it later.

experimental/sql/src/ibis_frontend.py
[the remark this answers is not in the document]
ingomueller-net: Cool, yes, that looks better.

Question: is it such that *every* expression has *exactly one* `op`? Would it make sense to visit any expression type with the same visitor, take the op out, and then visit that op (which dispatches to op-specific implementations)?

experimental/sql/tools/rel_opt.py
webmiche: Let me rephrase: Fundamentally, I feel that we are not really generating IR from a file like a *file*check test usually does, but rather from a file and a context. The exact schema of a table for example is part of that context. A linked part would be where to find column "a" in a tuple of table "t" (at position 0? or 1?) or where "t" actually is in memory. This is information that we will need at some point to actually be able to lower, but I am not sure whether this should be part of the IR of a query.

I would suggest leaving it as is and reopening this discussion once we start lowering to less abstract representations. Then we might be able to see more easily if this information should be part of the IR, or if this should be passed into the lowering pass as a context argument.

experimental/sql/src/ibis_frontend.py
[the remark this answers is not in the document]
ingomueller-net: What I was thinking is that there is an `AndExpression`, such that `$c > 1 and $b == "AS"` is translated into `AndExpression(Lt(Col("c"), Lit(1)), Eq(Col("b"), Lit("AS")))`. More predicates can then be expressed as nested `AndExpression`s but there is always just a single root expression.

experimental/sql/tools/rel_opt.py
ingomueller-net: OK, I think I understand. And I agree: let's leave it as is for now.

nicolasvasilache: sgtm!

PR #441, opened 2022-04-11
Fix bufferization of in_parallel

matthias-springer:
Copies were inserted at the wrong place and the wrong buffer was used
during bufferization.


nicolasvasilache:
This has landed in IREE as https://github.com/google/iree/commit/795bb5344f952e0fba73c7ffa3f04078f92aeb81

Closing here as it will be mirrored in the next IREE -> sandbox pull.

PR #443, opened 2022-04-11
Warp verifier

ThomasRaoux:
This is on top of the parser changes, please only look at the second commit in this PR


matthias-springer (approved):
nice!

PR #444, opened 2022-04-12
Move shared memory allocation to a lambda set by user

ThomasRaoux:
The memory allocation should be handled by user. It will allow user to
handle cases with multiple warps within a group and will give more
freedom to decide where the temporary memory should be allocated.


include/Dialect/VectorExt/VectorExtWarpUtils.h
```diff
@@ -22,7 +22,18 @@ namespace vector_ext {
 /// Collect patterns to propagate warp distribution.
 void populatePropagateVectorDistributionPatterns(RewritePatternSet &pattern);
 
-void populateWarpSingleLaneOpToScfForPattern(RewritePatternSet &patterns);
+/// Lamdba function to let user allocate memory needed for the lowering of
```
matthias-springer, line 25: nit: `users`

include/Dialect/VectorExt/VectorExtWarpUtils.h
```diff
@@ -22,7 +22,18 @@ namespace vector_ext {
...
+/// WarpSingleLaneOp.
+/// The function needs to be return an allocation that the lowering can use as
```
matthias-springer, line 27: nit: `needs to return`

include/Dialect/VectorExt/VectorExtWarpUtils.h
```diff
@@ -22,7 +22,18 @@ namespace vector_ext {
...
+/// temporary memory. The allocation needs to match the shape of the type (the
+/// type may be VectorType or a scalar) and be availble for the current warp. If
+/// there are several warps running in parallel the allocation needs to be split
+/// so that each warp as its own allocation.
```
matthias-springer, line 31: nit: `has its own allocation`

matthias-springer (approved):

PR #445, opened 2022-04-12
Add fusion into parallel op.

gysit:


lib/Dialect/LinalgExt/Transforms/Fusion.cpp
```diff
@@ -0,0 +1,108 @@
...
+        getValueOrCreateConstantIndexOp(b, sliceOp->getLoc(), std::get<1>(it));
+
+  return tiledIterationDomain;
+}
+
+FailureOr<linalg::LinalgOp>
+mlir::iree_compiler::IREE::LinalgExt::LinalgExtFusionPattern::
+    returningMatchAndRewrite(linalg::LinalgOp producerOp,
+                             PatternRewriter &rewriter) const {
+  // Check the producer has one result consumed by one tensor::ExtractSliceOp.
```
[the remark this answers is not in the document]
gysit: I did the change in a separate commit to make it easier to compare and reversible. As far as I know a pattern always matches one operation. This attempt matches the producer and the containing operation is handed in as a pattern option.

lib/Dialect/LinalgExt/Transforms/Fusion.cpp
```diff
@@ -0,0 +1,121 @@
...
+  for (Operation *user : producerOp->getUsers()) {
+    auto sliceOp = dyn_cast<tensor::ExtractSliceOp>(user);
+    if (!sliceOp)
+      continue;
+    if (sliceOp->getParentOp() != containingOp)
+      continue;
+    sliceOps.push_back(sliceOp);
+  }
+  // Check for a non-empty list of fusion opportunities.
+  if (sliceOps.empty())
```
nicolasvasilache, line 82: Fun extension for the future that I realized when writing a design doc, we can relax this and get other forms of fusion than just producer-consumer by simply adding the result slices to the terminator :)

nicolasvasilache (approved):
Looks nice, thanks !

nicolasvasilache:
Feel free to land and ignore IREE, I'll have to integrate anyway.

lib/Dialect/LinalgExt/Transforms/Fusion.cpp
gysit: right the dummy op is not needed. Although tiling a dummy op may still be the most convenient way to get the slices.

PR #448, opened 2022-04-12
Rename WarpSingleLaneOp to WarpExecuteOnLane0

ThomasRaoux:
This is to be consistent with the op printing

### DIFF
--- a/include/Dialect/VectorExt/VectorExtOps.td
+++ b/include/Dialect/VectorExt/VectorExtOps.td
@@ -59,7 +59,7 @@ def VectorExt_YieldOp : VectorExt_Op<"yield", [
   let assemblyFormat = "attr-dict ($operands^ `:` type($operands))?";
 }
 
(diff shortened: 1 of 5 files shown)


include/Dialect/VectorExt/VectorExtOps.td
```diff
@@ -59,7 +59,7 @@ def VectorExt_YieldOp : VectorExt_Op<"yield", [
   let assemblyFormat = "attr-dict ($operands^ `:` type($operands))?";
 }
 
-def VectorExt_WarpSingleLaneOp : VectorExt_Op<"warp_execute_on_lane_0",
+def VectorExt_WarpExecuteOnLane0 : VectorExt_Op<"warp_execute_on_lane_0",
```
matthias-springer, line 62: Are you sure you want to drop the `Op` suffix? All ops in the vector dialect end with `Op`.

ThomasRaoux: Good catch, fixed it

matthias-springer (approved):

PR #449, opened 2022-04-13
SQL: rel_tree dialect

webmiche:
This patch adds the rel_tree dialect. This dialect models relational queries as trees. I tried to model something close to our general selection example in a way that I feel it should look like, when looking at relational algebra. I am not sure whether this is how relational query tress are represented in databases, maybe there is some better design that I do not know of. Maybe @ingomueller-net  has input on that.

Also, I am not very happy with the name currently. rel_tree looks not very informative, whereas relational_query_tree looks overblown. Maybe someone has a better idea of a name.

Lastly, I did not yet think about how to lower ibis to this dialect, because I feel like this dialect should not be heavily influenced by ibis and how their trees look but rather how relational algebra "should" look.


experimental/sql/dialects/rel_tree.py
```diff
@@ -0,0 +1,223 @@
...
+  nullable = ParameterDef(IntegerAttr)
+
+  @staticmethod
+  @builder
+  def get(val: int) -> 'String':
+    return String([IntegerAttr.from_int_and_width(val, 1)])
+
+
+@irdl_op_definition
+class Column(Operation):
```
webmiche, line 61: I was not 100% sure about the design of this operation. In particular, I am not sure if it should also contain the table name or not. At the moment, we are modeling just a selection, so it is clear which table a column comes from and ingo suggested that one can also write `$a = 5` instead of `T.a = 5`, so I went with just the column name.

But thinking forward to maybe modeling joins, I think we want to reuse this operation and there AFAIU it could be the case that something like `T.id = S.id` comes up. This would obviously make this operation not strong enough. What do you think @ingomueller-net

ingomueller-net: I think we should try to be close to "textbook" relational algebra, i.e., the math-like form we use in introductory classes. In that case, a join is defined only if the subset of the names left and right are non-overlapping in order to avoid such ambiguities. This has the benefit that each expression really only needs to understand its direct inputs; and I see it as one of the main added values of the conversion from Ibis to relational to create this locality.

I guess typical Bachelor courses teach you to rename one of the columns if there is, in fact, a name clash. Our conversion could do the same. Another possibility is to make `"T.id"` the (flat) name (i.e., without exposing the table.column structure).

experimental/sql/dialects/rel_tree.py
```diff
@@ -0,0 +1,223 @@
...
+
+  @builder
+  @staticmethod
+  def get(table: Region, col_name: str) -> 'Column':
+    return Column.build(attributes={"col_name": StringAttr.from_str(col_name)},
+                        regions=[table])
+
+
+@irdl_op_definition
+class Predicate(Operation):
```
ingomueller-net, line 83: The semantics of this operation rather seem to suggest the name `Comparison`, no? For me, a "predicate" is any expression that returns a Boolean, i.e., Boolean operators such as `And` and `Or` are also predicates.

experimental/sql/dialects/rel_tree.py
```diff
@@ -0,0 +1,223 @@
...
+  """
+  name = "rel_tree.selection"
+
+  table = SingleBlockRegionDef()
+  predicates = SingleBlockRegionDef()
+
+  @staticmethod
+  @builder
+  def get(table: Region, predicates: Region) -> 'Selection':
+    return Selection.build(regions=[table, predicates])
```
ingomueller-net, line 134: In my vision of the relational tree dialect, this predicate/these predicates *including all its/their child expressions* only operate on scalars (as opposed to relations). Maybe it makes sense to introduce two different base classes for this purpose: `RelationlExpression`, which operates on relations (such as selection, project, join, etc.) and `ScalarExpression`, which are used to parametrize some of the former (such as the predicates).

experimental/sql/dialects/rel_tree.py
```diff
@@ -0,0 +1,223 @@
...
+  Example:
+
+  ```
+  rel_tree.schema_element() ["elt_name" = "id", "elt_type" = !rel_tree.int32]
+  ```
+  """
+  name = "rel_tree.schema_element"
+
+  elt_name = AttributeDef(StringAttr)
+  elt_type = AttributeDef(DataType)
```
ingomueller-net, line 177: These *could* also be just `name` and `type`... Although you'd clash with the keyword `type`, right? Maybe this is a good enough work-around then...

experimental/sql/tools/rel_opt.py
```diff
@@ -8,6 +8,7 @@
 
 from src.ibis_frontend import ibis_to_xdsl
 from dialects.ibis_dialect import Ibis
+from dialects.rel_tree import RelationalTree
```
ingomueller-net, line 11: Another possible name is `rel_alg` as reference to the mathematical meaning of relational algebra, which is just about expressions. The SSA form could be called `rel_prog` or `rel_ssa` because it is a "program" that works on SSA values that are relations...

ingomueller-net (commented):
No real blocker for this PR, just thoughts that could be part of this or a next iteration. Apart from the column name, maybe the distinction between relational and scalar expressions is the most interesting one.

experimental/sql/dialects/rel_tree.py
webmiche: I agree with this destinction, but I struggle to see how to implement this in the rel_tree dialect as these things feel like types to me and we don't really have types in the rel_tree (other than data types). This destinction might be something we should make explicit in the dialect below this one.

experimental/sql/dialects/rel_tree.py
webmiche: name actually clashes as well, since xDSL expects the string that gets parsed to a certain op to be a field named name (i.e. the field on line 174).

experimental/sql/dialects/rel_tree.py
webmiche: I agree and I renamed it to `compare`, since it feels more like an action to me.

experimental/sql/dialects/rel_tree.py
webmiche: Staying close to the textbook is precisely what I wanted with this dialect, so I guess my suspcion is no problem.

For the flat name, I feel that `$a = 5` is a bit more general than writing `T.a = 5` and I would want to model something as general as possible, so i would keep it as is. What are you thoughts, @ingomueller-net ?

experimental/sql/tools/rel_opt.py
webmiche: I really like that, since I feel that we really are modeling relational algebra in this dialect, while the dialect one lower is already a kind of `implementation`.

PR #451, opened 2022-04-20
SQL: projection frontend

webmiche:
This patch adds support for projections in the ibis frontend. As it turns out, ibis uses the same node for selections and projections, so I mirrored this behavior by adding another region to the `selection` node.


experimental/sql/dialects/ibis_dialect.py
```diff
@@ -38,7 +38,7 @@ class DataType(ParametrizedAttribute):
 
 
 @irdl_attr_definition
-class int32(DataType):
+class Int32(DataType):
```
ingomueller-net, line 41: I think that this should be in a dedicated mini-PR (or at least a dedicated commit).

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -126,11 +142,13 @@ class Selection(Operation):
...
   table = SingleBlockRegionDef()
   predicates = SingleBlockRegionDef()
+  projections = SingleBlockRegionDef()
 
   @staticmethod
   @builder
-  def get(table: Region, predicates: Region) -> 'Selection':
-    return Selection.build(regions=[table, predicates])
+  def get(table: Region, predicates: Region,
+          projections: Region) -> 'Selection':
```
ingomueller-net, line 150: I guess this class also deserves some doc that describes its semantics.

BTW: The Ibis `Selection` also seems to do (some of) the sorting in that operation.

experimental/sql/src/ibis_frontend.py
```diff
@@ -3,8 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 from dataclasses import dataclass
-from modulefinder import Module
-from unittest import result
```
ingomueller-net, line 7: I guess these are also unrelated changes? I'd move them to the same "clean-up" mini-PR I mentioned above.

experimental/sql/dialects/ibis_dialect.py
ingomueller-net: Question: Is a projection really just a "projection" in the sense of a list of existing columns you allow in the output? Or can you also add expressions that compute new columns?

ingomueller-net: Question: Can you rename a column with a projection/selection?

experimental/sql/src/ibis_frontend.py
```diff
@@ -71,6 +78,17 @@ def visit(  #type: ignore
...
 
+@dispatch(ibis.expr.types.IntegerColumn)
+def visit(  #type: ignore
+    intColumn: ibis.expr.types.IntegerColumn) -> Operation:
+  op = intColumn.op()
+  if isinstance(op, gen_types.TableColumn):
+    table = Region.from_operation_list([visit(op.table)])
+    new_op = id.TableColumn.get(table, op.name)
+    return new_op
+  raise Exception(f"Unknown stringcolumn: {type(op)}")
```
ingomueller-net, line 89: What is this class for? Are you using it in this PR already? (If not, separate it?)

ingomueller-net (approved):
Looks good to me, modulo the minor comments! I'd also rebase and remove the "add support for single projection" and its revert in order to reduce the noise in the PR.

experimental/sql/src/ibis_frontend.py
webmiche: This class is very similar to the handling of stringcolumns, but occurs when the column type is an int and not a string. So the class is actually used in the PR, since I enriched the example database, but it doesn't really have something to do with the topic of the PR. I guess I can move this out to a different PR if you want.

experimental/sql/dialects/ibis_dialect.py
webmiche: The `sort_keys` field is used for queries such as `table.sort_by('a')`. So there are three things (we know of) that a selection does, but I did not manage to create a `selection` that does multiple. They basically always just do one thing, so `table.sort_by('a')['a', 'b']` becomes two `selection`s.

For the expression part, I found out that you can write `table[table['a']]`, this gets inlined by the frontend and looks like `table['a']`. In the source code, they have a comment saying that the expressions can be `array expression`s, but I did not figure out how to create those.

I also did no manage to create new columns using a selection, since I ran into the error: `value 'd' is not a field in ['a', 'b', 'c']`, which seems to me like you cannot create new columns.

What I am unsure right know is how you would express a query like:

```
Select * From (
   Select a, b, a+b from T)
```

I think something like this should be a possible, but I don't know how you would write that in ibis.

ingomueller-net: Thanks for investigating! Here are a few things I managed to do:

```python
t[t['a'].name('b')]  # Rename 'a' to 'b'
t[(t['a'] + "foo").name('b')]  # Compute new column 'b'
t[t]  # Select all columns of t (which is interesting if you want to add some columns and preserve all existing ones)
t.projection(['a'])  # Looks like pure syntactic sugar for t[...]
t.a  # Looks like pure syntactic sugar for t['a']
t.mutate(b=t['a'])  # Add a new column while preserving all existing ones
```

There are also a few mentions of arrays in [the docs](https://ibis-project.org/docs/dev/ibis-for-sql-programmers/#projections-selectaddremove-columns), all of which I believe are not what we are looking for:

```python
t.a.value_counts()  # Not sure if this is really an "array" rather than a column. There are two more mentions of this type.
t[t.a.isin(['AS', 'AX'])]  # Doesn't seem like an array function either
```

What I think "really" refers to arrays is this (see [docs](https://ibis-project.org/docs/dev/api/expressions/collections/#ibis.expr.types.arrays)):

```python
t[ibis.array([t.a, t.a]).name('arr')]
t[ibis.array([t.a, t.a]).length().name('l')]
```

ingomueller-net: Some more array expressions:

```python
t.groupby('a').aggregate(t.a.collect().name('arr'))  # Create array through aggregation
t[(ibis.array([t.a]) + ibis.array([t.a])).name('arr')]  # Concat arrays
```

And a list of expressions with examples: https://github.com/ibis-project/ibis/issues/2621

ingomueller-net: Anyways, I guess that array columns aren't on the top of our priority list ;)

webmiche: Wow, thanks a lot for your investigation! I guess these are a lot of test cases that show some flaws in our current modeling, e.g. the renaming shows that a column in a selection actually needs a "source" and  "destination" column name. 

AFAIU, `t.projection(['a'])` and `t.a` are not exactly the same in terms of ibis data structures, since `t.a` does not actually contain a `selection`, but is only a `TableColumn`. I agree that they should be translated to the same IR later on, but they don't seem to be the same in terms of the ibis infrastructure.

I am actually uncertain here how we should continue. As I said, your example show a lot of flaws, but I feel like fixing all of them within this PR might be overkill. I suggest we add testcases for all of these examples in this PR and make them explicit fails and then actually add support for the necessary functionality later on.

ingomueller-net: Aha, the distinction you say between `t.a` and `t.projection(['a'])` makes sense! I guess that `t['a']` falls into the first category and `t[['a']]` in the second?

Oh, sure, absolutely! Most (or even none) of this shouldn't go to this PR. I guess what should go in is a comment what (we currently think) projection does in Ibis and what of this we support. Even the test cases can go into a next PR.

webmiche: Yes exactly!

To me it feels like the `selection` class in ibis is used in a lot of contexts and feels more like a SQL `Select` statement and all (or most) of the things you can do with it and not like relational algebra `select` node. I will write something along these lines...

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -119,18 +138,32 @@ class Selection(Operation):
...
     // predicates
     ibis.equals() ...
+  } {
+    // projections
+  }
+
+  ibis.selection() {
+    // table
+    ibis.table() ...
+  } {
```
nicolasvasilache, line 148: nit: can the xdsl parser support more readable custom assembly form ?
e.g.
```
ibis.selection()  
   with table       = { ... }
   with predicates  = { ... } 
   with projections = { ... }
```

webmiche: While that is certainly useful, there is no support for this yet, particularly not autogenerated for each operation. I think it should be possible to define custom printers and parsers, but AFAIK we did not develop very far in that direction yet.

I opened an [issue](https://github.com/xdslproject/xdsl/issues/113) on the min xdsl repo, since i remember us discussing that we want to develop this further. Hopefully, we won't forget with this issue :)

PR #455, opened 2022-04-21
Replace usage of nested PassManager-s in BufferizeOp::apply

shabalind:


shabalind:
As an added bonus we get to avoid re-running cleanup transformations multiple times (might improve compilation time).

ftynse (approved):

lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
```diff
@@ -699,17 +699,23 @@ transform::LowerVectorsOp::apply(transform::TransformResults &results,
...
-
   bufferization::OneShotBufferizationOptions options;
   options.memCpyFn = [](OpBuilder &builder, Location loc, Value from,
                         Value to) {
     return success(linalg::makeMemRefCopyOp(builder, loc, from, to));
   };
-  pm.addPass(createLinalgComprehensiveModuleBufferizePass(options));
-  if (failed(pm.run(state.getTopLevel())))
+
+  auto moduleOp = dyn_cast<ModuleOp>(state.getTopLevel());
```
ftynse, line 716: Just `cast` if you don't verify the success status.

lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
```diff
@@ -699,17 +699,23 @@ transform::LowerVectorsOp::apply(transform::TransformResults &results,
 // BufferizeOp
 //===---------------------------------------------------------------------===//
 
+static void applyBufferizationEnablingTransformations(ModuleOp moduleOp) {
+  RewritePatternSet patterns(moduleOp.getContext());
+  patterns.add<GeneralizePadOpPattern>(moduleOp.getContext());
+  (void)applyPatternsAndFoldGreedily(moduleOp, std::move(patterns));
```
ftynse, line 705: Propagate failure maybe?

shabalind: This helper comes from comprehensive bufferization pass where the failure wasn't propagated either. @matthias-springer is that intentional? Do we want to propagate failures here?

lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
shabalind: 👍

lib/Dialect/LinalgTransform/IR/LinalgTransformOps.cpp
shabalind: I'm keeping this as is for now, not to change the behavior compared to prior nested PassManager-based implementation. We can amend this change later if it's safe to propagate failures.

PR #456, opened 2022-04-21
SQL: rel_ssa prototype

webmiche:
This patch adds the rel_ssa dialect, a dialect that represents relational algebra using SSAValues to represent intermediate results.

### DIFF
--- a/experimental/sql/dialects/rel_ssa.py
+++ b/experimental/sql/dialects/rel_ssa.py
@@ -87,6 +87,27 @@ class Boolean(ParametrizedAttribute):
 
 
 @irdl_attr_definition
+class SchemaElement(ParametrizedAttribute):
+  """
+  Models an element of a schema with name `elt_name` and type `elt_type`.
+
+  Example:
+  '''
+  !rel_ssa.schema_element<"id", !rel_ssa.int32>
+  '''
+  """
+  name = "rel_ssa.schema_element"
+
+  elt_name = ParameterDef(StringAttr)
+  elt_type = ParameterDef(DataType)
+
+  @staticmethod
+  @builder
+  def get(name: str, type_: DataType) -> 'SchemaElement':
+    return SchemaElement([StringAttr.from_str(name), type_])
+
+
+@irdl_attr_definition
 class Bag(ParametrizedAttribute):
   """
   Models a bag in a relational SSA query. The exact schema of the bag is part of
@@ -95,17 +116,18 @@ class Bag(ParametrizedAttribute):
   Example:
 
   '''
-  !rel_ssa.bag<[!rel_ssa.int32]>
+  !rel_ssa.bag<[!rel_ssa.schema_element<"id", !rel_ssa.int32>]>
   '''
   """
   name = "rel_ssa.bag"
 
-  schema = ParameterDef(ArrayOfConstraint(DataType))
+  schema = ParameterDef(ArrayOfConstraint(SchemaElement))
 
   @staticmethod
   @builder
-  def get(types: list[DataType]) -> 'Bag':
-    return Bag([ArrayAttr.from_list(types)])  #type: ignore
+  def get(types: list[DataType], names: list[str]) -> 'Bag':
+    schema_elts = [SchemaElement.get(n, t) for n, t in zip(names, types)]
+    return Bag([ArrayAttr.from_list(schema_elts)])
 
 
 #===------------------------------------------------------------------------===#
@@ -288,6 +310,7 @@ class RelSSA:
     self.ctx.register_attr(Int32)
     self.ctx.register_attr(String)
     self.ctx.register_attr(Boolean)
+    self.ctx.register_attr(SchemaElement)
 
     self.ctx.register_op(Select)
     self.ctx.register_op(PandasTable)
--- a/experimental/sql/dialects/rel_ssa.py
+++ b/experimental/sql/dialects/rel_ssa.py
@@ -1,0 +1,159 @@
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from dataclasses import dataclass
+from typing import Any, Union, List
+from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
+from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr
+from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder
+
+# This file contains the relational dialect. It is currently
+# only aimed at the simple query and mirrors a lot of the
+# ibis dialect. This is subject to change, once the translation
+# to the iterator dialect starts getting in shape.
+
+
+@irdl_attr_definition
+class DataType(ParametrizedAttribute):
+  name = "rel_ssa.datatype"
+
+
+@irdl_attr_definition
+class Int32(DataType):
+  name = "rel_ssa.int32"
+
+
+@irdl_attr_definition
+class String(DataType):
+  name = "rel_ssa.string"
+
+  nullable = ParameterDef(IntegerAttr)
+
+  @staticmethod
+  @builder
+  def get(val: Union[int, IntegerAttr]) -> 'String':
+    if isinstance(val, IntegerAttr):
+      return String([val])
+    return String([IntegerAttr.from_int_and_width(val, 1)])
+
+
+@irdl_attr_definition
+class Boolean(ParametrizedAttribute):
+  name = "rel_ssa.bool"
+
+
+@irdl_attr_definition
+class Bag(ParametrizedAttribute):
+  name = "rel_ssa.bag"
+
+  schema = ParameterDef(ArrayOfConstraint(DataType))
+
+  @staticmethod
+  @builder
+  def get(types: list[DataType]) -> 'Bag':
+    return Bag([ArrayAttr.from_list(types)])  #type: ignore
+
+
+@irdl_op_definition
+class Column(Operation):
+  name = "rel_ssa.column"
+
+  col_name = AttributeDef(StringAttr)
+  result = ResultDef(DataType)
+
+
+@irdl_op_definition
+class Compare(Operation):
+  name = "rel_ssa.compare"
+
+  left = OperandDef(AnyAttr())
+  right = OperandDef(AnyAttr())
+  comparator = AttributeDef(StringAttr)
+
+  result = ResultDef(AnyAttr())
+
+  @staticmethod
+  @builder
+  def get(left: Operation, right: Operation,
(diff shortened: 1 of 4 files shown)


experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,271 @@
...
+
+  '''
+  !rel_ssa.bool
+  '''
+  """
+  name = "rel_ssa.bool"
+
+
+@irdl_attr_definition
+class Bag(ParametrizedAttribute):
```
webmiche, line 80: In the meeting, we discussed having named tuples as part of this type. I have yet to investigate how this will work using xDSL.

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,271 @@
...
+  Interface class for all Expressions, i.e., operations that work on single
+  values.
+  """
+  ...
+
+
+@irdl_op_definition
+class Column(Expression):
+  """
+  References a specific column with name `col_name` and returns its value.
```
ingomueller-net, line 112: I think it would be helpful to briefly talk about the scope of these names. Here is probably a good place. In this dialect, names refer to the schema of the encompassing operator -- and *never* an operator that precedes it like in the Ibis dialect.

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,271 @@
...
+  def get(name: str, result_type: Attribute) -> 'PandasTable':
+    return PandasTable.build(
+        attributes={"table_name": StringAttr.from_str(name)},
+        result_types=[result_type])
+
+
+@irdl_op_definition
+class Select(Operator):
+  """
+  Selects all tuples of `input` according to `predicates`.
```
ingomueller-net, line 229:
```suggestion
  Selects all tuples of `input` according to the Expressions `predicates`.
```

or similar... (make the connection between expressions and operators explicit in the doc).

experimental/sql/dialects/rel_ssa.py
ingomueller-net: OK, cool! If it's not trivial, feel free to leave it as is in this PR and change later.

ingomueller-net (approved):

experimental/sql/dialects/rel_ssa.py
webmiche: I went with an approach that introduces an attribute `SchemaElement` that contains a `StringAttr` for the name and a `DataType` for the type. This makes the IR quite cluttered, but I feel we are only custom printer away from what we actually would want (i.e. syntax of the form `*name* : *type*`).

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+  Example:
+
+  ```
+  !rel_ssa.string<0: !i1>
+  ```
+  """
+  name = "rel_ssa.string"
+
+  # TODO: redefine nullable as a property of all fields
+  nullable = ParameterDef(IntegerAttr)
```
nicolasvasilache, line 59: I am not completely clear what this is .. in core MLIR parlance, is this an IntegerAttr "type attribute" that describes whether the string is nullable or not (i.e. optional) ?

If so, it seems quite unfortunate that xdsl renames MLIR concepts and we should make it clear and loud in the comments at least that this is an MLIR Type Attribute.

Also, if my interpretation is correct, should this be a BoolAttr ?

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+#===------------------------------------------------------------------------===#
+# Query types
+#===------------------------------------------------------------------------===#
+
+
+@irdl_attr_definition
+class Boolean(ParametrizedAttribute):
+  """
+  Models a type that can either be true or false to, e.g., show whether a tuple
+  is part of the result or not.
```
nicolasvasilache, line 78: it would be nice to have a sightly more fleshed out IR example to illustrate the "show whether a tuple..." part.

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+  !rel_ssa.bag<[!rel_ssa.schema_element<"id", !rel_ssa.int32>]>
+  '''
+  """
+  name = "rel_ssa.bag"
+
+  schema = ParameterDef(ArrayOfConstraint(SchemaElement))
+
+  @staticmethod
+  @builder
+  def get(types: list[DataType], names: list[str]) -> 'Bag':
```
nicolasvasilache, line 128: nit: your logic, printing and parsing use `(name, type)` but your API uses `(type, name)`.
Can you make the API follow the rest for consistency?

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+@irdl_op_definition
+class Column(Expression):
+  """
+  References a specific column with name `col_name` and returns its value. The
+  name 'col_name' must be part of the schema of an encompassing operator.
+
+  Example:
+
+  '''
+  %0 : rel_ssa.int32 = rel_ssa.column() ["col_name" = "id"]
```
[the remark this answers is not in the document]
nicolasvasilache: ah I see that in the comment at the top:
```
Operators work on Bags whereas expressions work on
single values and are used to specify the exact behavior of operators.
```

We should be very loud about putting the essence of this comment into every op and type documentation.

I am quite unclear about the modelling at this point.
Here it seems you want to create a new column named "id" and containing int32 values, so I would expect a type like `column<"id", int32>`?
The problem I see is that `%0` is locally an `int32` in all its uses but in practice it is globally a `column<int32>` and we need to traverse SSA use-def chains to know what is what.
The MLIR printer/parser logic will really not like that.

Am I misunderstanding the modeling semantics you have with XDSL ?

I would additionally separate the verification invariants into a separate section `Verification Invariants:` in the same way you have `Example:`.

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+  Example:
+
+  '''
+  ...
+  %0 : !rel_ssa.bool = rel_ssa.compare(%a : !rel_ssa.int32, %b : !rel_ssa.int32) ["comparator" = "="]
+  '''
+  """
+  name = "rel_ssa.compare"
+
+  left = OperandDef(DataType)
```
nicolasvasilache, line 178: Same comment as above, this one really operates on int32 types.
This begs the question of just using `int32` here and even just using a vanilla MLIR func + SymbolAttr that operates on int32.
This op is quite different from the Column op.

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+
+@irdl_op_definition
+class PandasTable(Operator):
+  """
+  Defines a table with name `table_name`.
+
+  Example:
+
+  '''
+  %0 : rel_ssa.bag<[!rel_ssa.int32]> = rel_ssa.pandas_table() ["table_name" = "t"]
```
nicolasvasilache, line 258: ok, this makes sense to me.
I guess I have trouble understanding the Column op ..

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+
+@irdl_op_definition
+class Select(Operator):
+  """
+  Selects all tuples of `input` according to the yielded expression in `predicates`.
+
+  Example:
+
+  '''
+  rel_ssa.select(%0: rel_ssa.bag<[!rel_ssa.int32]>) {
```
nicolasvasilache, line 282: I am quite unclear about wanting this modeling as it seems you could achieve the same with:
```
func.func some_fun(%0: i32) -> i1 {
  ...
  return %3 : i1
}

rel_ssa.select "some_fun" from %0: rel_ssa.bag<i32>
```

experimental/sql/dialects/rel_ssa.py
webmiche: AFAIU, this is exactly the [MLIR IntegerAttr](https://mlir.llvm.org/docs/Dialects/Builtin/#integerattr). It means that this type has a parameter that defines whether it is nullable or not. And this parameter stores an integer.

I agree that `BoolAttr` would be nicer, but this has not yet been introduced to our builtin dialect since its basically sugar for i1 anyhow. In a final version, I agree that this should be a `BoolAttr`.

experimental/sql/dialects/rel_ssa.py
```diff
@@ -0,0 +1,320 @@
...
+
+  # TODO: redefine nullable as a property of all fields
+  nullable = ParameterDef(IntegerAttr)
+
+  @staticmethod
+  @builder
+  def get(val: Union[int, IntegerAttr]) -> 'String':
+    if isinstance(val, IntegerAttr):
+      return String([val])
+    return String([IntegerAttr.from_int_and_width(val, 1)])
```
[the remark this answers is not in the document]
webmiche: This constructs a String from an IntegerAttr or from an int. Whether or not the string is nullable is decided by the int in one case or by the IntegerAttr (that internally has a value field).

In essence, the union type and the `isinstance` were a try to have an API that looks simple (always use `get(...)` to build something), while still being expressive (can use int and IntegerAttr to build). In the end, I feel that this kind of code should be removed, since it is quite ugly. The approach of having two functions `from_int` and `from_integerattr` would probably work better or using the mutipledispach module to dispatch to the right thing according to the type.

And again, I actually want a `Bool` but that is not supported yet, so i1 is used in this prototype.

experimental/sql/dialects/rel_ssa.py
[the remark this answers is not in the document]
webmiche: I think this might just be a naming problem. In the `rel_impl` dialect I named this op `IndexByName` since it indexes into the tuple using a name. This will produce an `int32`, since that type of the field called `id` is `int32` in the given table.

experimental/sql/dialects/rel_ssa.py
webmiche: Yes, I think you can achieve the same using this modeling. I must admit, I don't feel like I understand all the details well enough to really decide what makes most sense. To me, the current version is nice, since the operation and the function that the operation applies are very close to each other. That feels to me like it makes the IR more readable than having some functions at first and then all the actual `select` operations.

I think in a lowered setting, say the `iterators`-dialect, we might want to have this function version. But for this, we can just move the body to another function during the lowering, I feel.

webmiche:
Yes, I think the whole bags, columns, scalars, tuples... topic and how to model all of these and how they relate is certainly a fundamental question of this thesis.

experimental/sql/dialects/rel_ssa.py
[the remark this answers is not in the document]
ingomueller-net: Sorry for chiming in late but it might also help to rename `val`, for example, to `is_nullable`.

experimental/sql/dialects/rel_ssa.py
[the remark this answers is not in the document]
ingomueller-net: Other possibilities for naming this operation: `ColumnRef`, `FieldRef`, `ExtractValue`, `ExtractField`...

ingomueller-net (commented):
Adding a few additional thoughts to the conversations...

PR #461, opened 2022-04-25
SQL: alg to ssa conversion

webmiche:
This patch adds a first prototype of a conversion from rel_alg to rel_ssa.

### DIFF
--- a/experimental/sql/src/alg_to_ssa.py
+++ b/experimental/sql/src/alg_to_ssa.py
@@ -45,7 +45,7 @@ class RelAlgRewriter(RewritePattern):
   def find_type_in_parent_op(self, name: str, op: Operation):
     """
     Crawls through all parent_ops until reaching either a ModuleOp, in which
-    case the lookup failed or reaching an operation with an input bag, that the
+    case the lookup failed or reaching an operator with an input bag, that the
     type can be looked up in.
     """
     parent_op = op.parent_op()
--- a/experimental/sql/src/alg_to_ssa.py
+++ b/experimental/sql/src/alg_to_ssa.py
@@ -42,7 +42,7 @@ class RelAlgRewriter(RewritePattern):
         return s.elt_type
     return None
 
-  def lookup_type_in_parent_op(self, name: str, parent_op: Operation):
+  def find_type_in_parent_op(self, name: str, parent_op: Operation):
     """
     Crawls through all parent_ops until reaching either a ModuleOp, in which
     case the lookup failed or reaching an operation with an input bag, that the
@@ -56,7 +56,7 @@ class RelAlgRewriter(RewritePattern):
       if type_:
         return type_
       raise Exception(f"element not found in parent schema: {name}")
-    return self.lookup_type_in_parent_op(name, parent_op.parent_op())
+    return self.find_type_in_parent_op(name, parent_op.parent_op())
 
 
 #===------------------------------------------------------------------------===#
@@ -84,7 +84,7 @@ class ColumnRewriter(RelAlgRewriter):
 
   @op_type_rewrite_pattern
   def match_and_rewrite(self, op: RelAlg.Column, rewriter: PatternRewriter):
-    res_type = self.lookup_type_in_parent_op(op.col_name.data, op.parent_op())
+    res_type = self.find_type_in_parent_op(op.col_name.data, op.parent_op())
     new_op = RelSSA.Column.get(op.col_name.data, res_type)
     rewriter.insert_op_before_matched_op([new_op])
     rewriter.erase_matched_op()
--- a/experimental/sql/src/alg_to_ssa.py
+++ b/experimental/sql/src/alg_to_ssa.py
@@ -16,9 +16,9 @@ import dialects.rel_ssa as RelSSA
 # algebra dialect to the relational SSA dialect. The current design has a parent
 # class `RelAlgRewriter` that contains functions used for several `Rewriter`s.
 # All other `Rewriter`s inherit from that class. To introduce SSA, the passes
-# insert the children, that should become operands, before the matched op, then
-# use the `added_operations_before` functionality of the rewriter to link the
-# operands to the new operation.
+# insert the children (and the their whole block), that should become operands,
+# before the matched op, then use the `added_operations_before` functionality of
+# the rewriter to link the operands to the new operation.
 
 
 @dataclass
@@ -42,12 +42,13 @@ class RelAlgRewriter(RewritePattern):
         return s.elt_type
     return None
 
-  def find_type_in_parent_op(self, name: str, parent_op: Operation):
+  def find_type_in_parent_op(self, name: str, op: Operation):
     """
     Crawls through all parent_ops until reaching either a ModuleOp, in which
     case the lookup failed or reaching an operation with an input bag, that the
     type can be looked up in.
     """
+    parent_op = op.parent_op()
     if isinstance(parent_op, ModuleOp):
       raise Exception(f"element not found in parent schema: {name}")
     if isinstance(parent_op, RelAlg.Select):
@@ -56,14 +57,14 @@ class RelAlgRewriter(RewritePattern):
       if type_:
         return type_
       raise Exception(f"element not found in parent schema: {name}")
-    return self.find_type_in_parent_op(name, parent_op.parent_op())
+    return self.find_type_in_parent_op(name, parent_op)
 
 
 #===------------------------------------------------------------------------===#
 # Expressions
 #===------------------------------------------------------------------------===#
 """
-All expression rewriters implicitely assume that the last operation in a lock is
+All expression rewriters implicitly assume that the last operation in a block is
 the one to be yielded.
 """
 
@@ -84,7 +85,7 @@ class ColumnRewriter(RelAlgRewriter):
 
   @op_type_rewrite_pattern
   def match_and_rewrite(self, op: RelAlg.Column, rewriter: PatternRewriter):
-    res_type = self.find_type_in_parent_op(op.col_name.data, op.parent_op())
+    res_type = self.find_type_in_parent_op(op.col_name.data, op)
     new_op = RelSSA.Column.get(op.col_name.data, res_type)
     rewriter.insert_op_before_matched_op([new_op])
     rewriter.erase_matched_op()
--- a/experimental/sql/src/alg_to_ssa.py
+++ b/experimental/sql/src/alg_to_ssa.py
@@ -138,6 +138,11 @@ class PandasTableRewriter(RelAlgRewriter):
     rewriter.erase_matched_op()
 
 
+#===------------------------------------------------------------------------===#
+# Conversion setup
+#===------------------------------------------------------------------------===#
+
+
 def alg_to_ssa(ctx: MLContext, query: ModuleOp):
   walker = PatternRewriteWalker(GreedyRewritePatternApplier([
       PandasTableRewriter(),
--- a/experimental/sql/src/alg_to_ssa.py
+++ b/experimental/sql/src/alg_to_ssa.py
@@ -34,15 +34,38 @@ class RelAlgRewriter(RewritePattern):
 
   def lookup_type_in_schema(self, name: str,
                             bag: RelSSA.Bag) -> Optional[RelSSA.DataType]:
+    """
+    Looks up the type of name in the schema of bag.
+    """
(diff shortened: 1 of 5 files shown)


experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+@dataclass
+class ColumnRewriter(RelAlgRewriter):
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelAlg.Column, rewriter: PatternRewriter):
+    res_type = self.lookup_type_in_schema(
+        op.col_name.data,
+        op.parent_op().parent_op().input.op.results[0].typ)
```
webmiche, line 66: I am not happy with this particular line just yet. This depends on the fact that the parent_op of the parent_op is a select and it cannot be used otherwise. I am still considering options of how to circumvent this. 

I am thinking of adding a data structure that keeps track of the schemas of different tables. Then, this would just become a lookup in that data structure. But how the data structure looks like is not clear either. I feel that a dictionary mapping table_names (strings) to dictionaries mapping column_names (strings) to data types could be feasible, but then we need the table_name here, i.e. we basically have the exact same problem.

In the end, I feel that I cannot implement this without having to lookup the parent_op and get the schema from there, in a sense, that is the price we pay for not having the table_name in the column operation.

Obviously, the other option we have is going back through the entire pipeline and adding the table_name everywhere. Then the dictionary approach would work again.

I plan to cleanup this parent_lookup version in a cleaner way s.t. it generalizes better (i.e. lookup the next parent_op with an input table and lookup in its schema), but I would happy to hear suggestions.

experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,151 @@
...
+    type can be looked up in.
+    """
+    if isinstance(parent_op, ModuleOp):
+      raise Exception(f"element not found in parent schema: {name}")
+    if isinstance(parent_op, RelAlg.Select):
+      type_ = self.lookup_type_in_schema(name,
+                                         parent_op.input.op.results[0].typ)
+      if type_:
+        return type_
+      raise Exception(f"element not found in parent schema: {name}")
```
ingomueller-net, line 58: Why does one of the two functions raise an exception and the other one return `None`? The names have the same structure -- that suggests the same behavior, no?

experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,151 @@
...
+      if s.elt_name.data == name:
+        return s.elt_type
+    return None
+
+  def lookup_type_in_parent_op(self, name: str, parent_op: Operation):
+    """
+    Crawls through all parent_ops until reaching either a ModuleOp, in which
+    case the lookup failed or reaching an operation with an input bag, that the
+    type can be looked up in.
+    """
```
ingomueller-net, line 50: Can you convince me that this always does the right thing?

If I have understood the structure of this dialect correctly, "expressions" are nested inside of "operators," which, in turn, form a program in SSA form. I *guess* that at this point we know the schema of any operator in the top-level program, no? So, from any expression, the first containing *operator* -- not just the first containing `Select` -- should have the relevant schema, no?

If that's the case, I guess that the `Operator` class could/should have a schema property such that you can write this function against the `Operator` interface (with `isinstance(parent_op, Relalg.Operator)` or similar).

experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,151 @@
...
+        RelSSA.Literal.get(op.val, self.convert_datatype(op.type)))
+    rewriter.erase_matched_op()
+
+
+@dataclass
+class ColumnRewriter(RelAlgRewriter):
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelAlg.Column, rewriter: PatternRewriter):
+    res_type = self.lookup_type_in_parent_op(op.col_name.data, op.parent_op())
```
ingomueller-net, line 87: You can actually call this function on `op` instead of `op.parent_op()` (and call the parameter of that function `op` instead of `parent_op`). It's a bit shorter, and the function name already implies that it will look at the parent...

experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,151 @@
...
+  def match_and_rewrite(self, op: RelAlg.PandasTable,
+                        rewriter: PatternRewriter):
+    schema_names = [s.elt_name.data for s in op.schema.ops]
+    schema_types = [self.convert_datatype(s.elt_type) for s in op.schema.ops]
+    result_type = RelSSA.Bag.get(schema_types, schema_names)
+    new_op = RelSSA.PandasTable.get(op.table_name.data, result_type)
+    rewriter.insert_op_before_matched_op(new_op)
+    rewriter.erase_matched_op()
+
+
```
ingomueller-net, line 139: Maybe a section heading for "conversion setup" or similar?

ingomueller-net (approved):
Looks good! The only major comment I have is whether the schema shouldn't be part of the `Operator` interface. This may be such a major change that you may want to do that in a next PR (if you agree that it's a good idea).

experimental/sql/src/alg_to_ssa.py
webmiche: I imagined that there might be reasons why we want to get a None from the first function later on, whereas I was sure what the exact behavior of the second function should be. I guess the return type hints at what will happen, but I agree this should be more expicit.

I renamed the second to `find_...` s.t. the naming does not suggest the same behavior as much as before.

experimental/sql/src/alg_to_ssa.py
ingomueller-net: Ha, I didn't see this comment during my review because you had clean-up this line already. I think the clean-up was a good step in the right direction, but see my comments there.

experimental/sql/src/alg_to_ssa.py
webmiche: Yes you are right, as long as we do not have operators with nested expressions that access fields that are not part of the output anymore. I am unsure whether there can be operators like that in relational algebra, I can think of a natural join that, in some sense, removes a column (compared to the cross-product), but obviously that column is part of the result anyway...

My plan was to add matching for more operators, when they are introduced and then look at their inputs, because the name certainly has to be part of one of the input schemas.

ingomueller-net: We haven't really fleshed that out yet but I think that making sure that these names exist is the job of some earlier conversions. In the relational algebra dialects, I think that that should be an invariant: only names can be accessed that are made available by the encompassing operator (with matching types).

Of course, you can spell out what all types do (then add a comment that that's how you plan to extend it) but I think it would be a good thing to factor out that functionality. *Every* operator produces a relation for which we have to be able to say what the schema is and -- as I argue -- the column expression should by design refer to that schema in every operator that is parametrized with an expression...

webmiche: I think the thing I am uncertain about atm is that to me the `column()` expression takes a field of the *input* tuple. AFAIU, what you are suggesting is looking up the type in the *output* relation. Is it really the case that this always works? If so, I think your approach is absolutely valid and makes sense.

webmiche: I implemented your suggestion :)

ingomueller-net: > I think the thing I am uncertain about atm is that to me the `column()` expression takes a field of the _input_ tuple. AFAIU, what you are suggesting is looking up the type in the _output_ relation. Is it really the case that this always works? If so, I think your approach is absolutely valid and makes sense.

I hadn't though about input vs output names. I think I only though about input names as well. One operator that changes the set of output names is the "rename" operator (or similar), which would have to handle the new names with a different mechanism than `column()`.

> I implemented your suggestion :)

Cool, I'll take a look.

webmiche: The other operator that changes the set I can think of is a projection. But both don't actually need expressions to parametrize, so for those two, your approach should work.

experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,163 @@
...
+      return RelSSA.Int32()
+    raise Exception(
+        f"datatype conversion not yet implemented for {type(type_)}")
+
+  def lookup_type_in_schema(self, name: str,
+                            bag: RelSSA.Bag) -> Optional[RelSSA.DataType]:
+    """
+    Looks up the type of name in the schema of bag.
+    """
+    for s in bag.schema.data:
```
nicolasvasilache, line 40: Seems like some verification on the type would be welcome?
Could you add a TODO to the definition of Bag ?

experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    parent_op = op.parent_op()
+    if isinstance(parent_op, ModuleOp):
+      raise Exception(f"element not found in parent schema: {name}")
+    if isinstance(parent_op, RelSSA.Operator):
+      type_ = self.lookup_type_in_schema(name, parent_op.results[0].typ)
+      if type_:
+        return type_
+      raise Exception(f"element not found in parent schema: {name}")
+    return self.find_type_in_parent_operator(name, parent_op)
```
nicolasvasilache, line 59: MLIR actively discourages recursion. This ones seems easy to preemptively convert to a while loop.

experimental/sql/src/alg_to_ssa.py
```diff
@@ -0,0 +1,163 @@
...
+# All other `Rewriter`s inherit from that class. To introduce SSA, the passes
+# insert the children (and the their whole block), that should become operands,
+# before the matched op, then use the `added_operations_before` functionality of
+# the rewriter to link the operands to the new operation.
+
+
+@dataclass
+class RelAlgRewriter(RewritePattern):
+
+  def convert_datatype(self, type_: RelAlg.DataType) -> RelSSA.DataType:
```
nicolasvasilache, line 27: This is quite strange to me that a RewritePattern allows type conversion.
There are non-obvious caveats on the MLIR side re conversion infra, type converter, rollbacks, implicit type conversions introduced by the infra, verification and how all these mix with block arguments and returns.

Is there a good doc / rationale for this? Maybe loop in Tobias and/or Matthieu ?

experimental/sql/test/alg_to_ssa/selection_op.xdsl
```diff
@@ -0,0 +1,23 @@
...
+            rel_alg.column() ["col_name" = "id"]
+        } {
+            rel_alg.literal() ["val" = 5 : !i64, "type" = !rel_alg.int32]
+        }
+    }
+}
+
+//      CHECK:    %0 : !rel_ssa.bag<[!rel_ssa.schema_element<"id", !rel_ssa.int32>]> = rel_ssa.pandas_table() ["table_name" = "t"]
+// CHECK-NEXT:    %1 : !rel_ssa.bag<[!rel_ssa.schema_element<"id", !rel_ssa.int32>]> = rel_ssa.select(%0 : !rel_ssa.bag<[!rel_ssa.schema_element<"id", !rel_ssa.int32>]>) {
+// CHECK-NEXT:        %2 : !rel_ssa.int32 = rel_ssa.column() ["col_name" = "id"]
```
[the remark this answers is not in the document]
webmiche: In a sense yes. I think this gets clear in the rel_impl dialect. where there is also a tuple type, that is the operand of the column operation.

experimental/sql/src/alg_to_ssa.py
webmiche: This actually goes beyond my understanding of MLIR. Maybe @math-fehr or @tobiasgrosser can help?

math-fehr: So here, this is not really a type conversion like in MLIR AFAIU.
This is just a method that will convert a type to another, and this method is used in the rewrite pattern.
Like in here:
```
  def match_and_rewrite(self, op: RelAlg.Literal, rewriter: PatternRewriter):
    new_op = RelSSA.Literal.get(op.val, self.convert_datatype(op.type))
    rewriter.replace_matched_op([new_op, RelSSA.Yield.get([new_op])])
```

xDSL does not have support for type conversion AFAIK. Currently, we are just writing rewrite patterns that will happen to convert types in blocks contained in the operations.

PR #462, opened 2022-04-26
SQL: rel_impl first prototype

webmiche:
This patch adds a first prototype of the rel_impl dialect, a dialect that should be closer to the implementation than the rel_ssa dialect. In particular, there are block arguments introduced for selections and there should only be operations for join implementations in the future.

### DIFF
--- a/experimental/sql/dialects/rel_impl.py
+++ b/experimental/sql/dialects/rel_impl.py
@@ -12,8 +12,8 @@ from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBloc
 # expresses relational queries using low-level concepts like tuples as block
 # arguments and specific join implementations. The dialect consists of two types
 # of operations: Expressions and Operators. Operators work on Bags whereas
-# expressions work on single values and are used to specify the exact behavior
-# of operators. The IndexByName and Yield Expressions bridge the gap to convert
+# expressions work on scalars and are used to specify the exact behavior of
+# operators. The IndexByName and Yield Expressions bridge the gap to convert
 # from one to the other.
 
 #===------------------------------------------------------------------------===#
@@ -171,7 +171,7 @@ class Expression(Operation):
 @irdl_op_definition
 class IndexByName(Expression):
   """
-  Indexes into `tuple` at the position of the filed `col_name` and returns its
+  Indexes into `tuple` at the position of the field `col_name` and returns its
   content.
 
   Example:
--- a/experimental/sql/dialects/rel_impl.py
+++ b/experimental/sql/dialects/rel_impl.py
@@ -1,0 +1,354 @@
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+from dataclasses import dataclass
+from typing import Any, Union, List
+from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
+from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr
+from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder
+
+# This file contains the relational implementation dialect, a dialect that
+# expresses relational queries using low-level concepts like tuples as block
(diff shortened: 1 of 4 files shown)


experimental/sql/dialects/rel_impl.py
```diff
@@ -0,0 +1,354 @@
...
+from typing import Any, Union, List
+from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
+from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr
+from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder
+
+# This file contains the relational implementation dialect, a dialect that
+# expresses relational queries using low-level concepts like tuples as block
+# arguments and specific join implementations. The dialect consists of two types
+# of operations: Expressions and Operators. Operators work on Bags whereas
+# expressions work on single values and are used to specify the exact behavior
```
ingomueller-net, line 15: Are these "values" always tuples or can they be something else? If they can only be tuples, consider using that word here.

experimental/sql/dialects/rel_impl.py
```diff
@@ -0,0 +1,354 @@
...
+  Interface class for all Expressions, i.e., operations that work on single
+  values.
+  """
+  ...
+
+
+@irdl_op_definition
+class IndexByName(Expression):
+  """
+  Indexes into `tuple` at the position of the filed `col_name` and returns its
```
ingomueller-net, line 174:
```suggestion
  Indexes into `tuple` at the position of the field `col_name` and returns its
```

experimental/sql/dialects/rel_impl.py
```diff
@@ -0,0 +1,354 @@
...
+
+  @staticmethod
+  @builder
+  def get(types: list[DataType], names: list[str]) -> 'Bag':
+    schema_elts = [SchemaElement.get(n, t) for n, t in zip(names, types)]
+    return Bag([ArrayAttr.from_list(schema_elts)])
+
+
+@irdl_attr_definition
+class Tuple(ParametrizedAttribute):
```
ingomueller-net, line 136: As far as I can tell, this is the only difference from the previous dialect, right?

I totally support the strategy to start with the "maximum" number of dialects and then reduce but this seems like a good candidate to merge... 😉

ingomueller-net (approved):

experimental/sql/dialects/rel_impl.py
webmiche: At the moment, there are literals as well, so they are not only tuples. I guess using the word `scalar` would be the right word here.

experimental/sql/dialects/rel_impl.py
webmiche: I 100% agree here. I am not quite sure which version we would take in the merger, probably this lower-level one because it is probably simpler to lower.

The main thing I can think of that will be added as well is the implementations of joins, which is not a huge change either :)

experimental/sql/dialects/rel_impl.py
ingomueller-net: Is this the same `Scalar` that is used in the `Sum`? See my discussion there. In plain SQL and in traditional relational algebra, relations *always* consist of tuples and these "schemaless" top-level operations like Ibis' `Sum` don't exist. Of course, we may decide to support them anyways but my feeling is that our lives may be simple if we initially consider them out of scope.

webmiche: What I understand as a scalar is similar to how in linear algebra matrix elements are scalar. Say a table is basically a 2D matrix, a scalar is then an entry with a fixed column and fixed row (so a single integer for example). From my understanding, these things do exist in a sense in relational algebra when writing things like: `t.a = 5`. The thing that is compared as part of `t.a` has a fixed column ('a') and, when comparing, has a fixed row, the current row.

So to me scalars are the things a table consists of and I would argue that this is the distinction between operators and expressions, one works on scalars, the other one on bags (possibly parametrized by some scalar expression similar to scalar times vector multiplication).

ingomueller-net: OK, makes sense. I think my comment boils down to the question whether we want relations to be restricted "bags of tuples of scalars" or also allow "bags of scalars". SQL does not allow the the latter, but your first implementation of `Sum` required it. If we do not allow "bags of scalars", then the "values" in the next above are always tuples...

webmiche: Are they? What about literals? I guess we can look at literals as "constant tuples" :thinking:

nicolasvasilache:
LG!

PR #466, opened 2022-04-28
SQL: Ibis frontend refactor

webmiche:
This patch refactors the ibis frontend to use more of multipledispatch.

### DIFF
--- a/experimental/sql/src/ibis_frontend.py
+++ b/experimental/sql/src/ibis_frontend.py
@@ -30,7 +30,7 @@ def convert_datatype(type_: ibis.expr.datatypes) -> id.DataType:
   raise KeyError(f"Unknown datatype: {type(type_)}")
 
 
-def visit_schema(schema: ibis.expr.schema.Schema) -> Region:  #type: ignore
+def visit_schema(schema: ibis.expr.schema.Schema) -> Region:
   ops = []
   for n, t in zip(schema.names, schema.types):
     ops.append(id.SchemaElement.get(n, convert_datatype(t)))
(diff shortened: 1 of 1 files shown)


experimental/sql/src/ibis_frontend.py
```diff
@@ -36,7 +30,7 @@ def convert_datatype(type_: ibis.expr.datatypes) -> id.DataType:
   raise KeyError(f"Unknown datatype: {type(type_)}")
 
 
-def visit_schema(schema: ibis.expr.schema.Schema) -> Region:  #type: ignore
+def visit_schema(schema: ibis.expr.schema.Schema) -> Region:
```
ingomueller-net, line 33: Why does this function and the next not use multipledispatch? Add a small comment?

experimental/sql/src/ibis_frontend.py
```diff
@@ -50,92 +44,63 @@ def visit_ibis_expr_list(l: List[ibis.expr.types.Expr]) -> Region:
...
-    table = Region.from_operation_list([visit(op.table)])
-    metrics = visit_ibis_expr_list(op.metrics)
-    new_op = id.Aggregation.get(table, metrics)
-    return new_op
-  raise KeyError(f"Unknown tableExpr: {type(op)}")
-
-
-@dispatch(ibis.expr.types.StringColumn)
+    op: ibis.expr.types.Expr) -> Operation:
+  return visit(op.op())
```
ingomueller-net, line 50: This is nice!

ingomueller-net (approved):
Cool, looks much better, IMO!

experimental/sql/src/ibis_frontend.py
webmiche: They return Regions as they work on multiple nodes in the ibis tree whereas the other functions return Operations for the single tree nodes they work. Added a comment :)

PR #468, opened 2022-05-02
SQL: ssa_to_impl conversion

webmiche:
This patch adds a first version of the conversion from ssa to impl. There are several not nice parts here, some should be made part of xDSL at some point, others require more thought process for future operations.


experimental/sql/src/ssa_to_impl.py
```diff
@@ -0,0 +1,137 @@
...
+        [RelImpl.Literal.get(op.value, self.convert_datatype(op.result.typ))])
+
+
+@dataclass
+class ColumnRewriter(RelSSARewriter):
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelSSA.Column, rewriter: PatternRewriter):
+    # TODO: In this current version, there is only one block_arg, so it is clear
+    # what to access here. What if this is not the case?
```
webmiche, line 63: I think this will becomes especially interesting when we have joins in our IR, considering that the join region will have multiple tuple arguments, so we will have to figure out to which tuple a given name belongs. I think this can be written by iterating over all tuples and all names, which I will implement once we actually have a usecase for it.

ingomueller-net (approved):

PR #469, opened 2022-05-02
SQL: sum in the rel_ssa dialect

webmiche:
This patch adds support for the sum operation in the rel_ssa dialect.

### DIFF
--- a/experimental/sql/dialects/rel_ssa.py
+++ b/experimental/sql/dialects/rel_ssa.py
@@ -304,6 +304,51 @@ class Select(Operator):
                         result_types=[input.results[0].typ])
 
 
+@irdl_op_definition
+class Aggregate(Operator):
+  """
+  Applies the ith function of `functions` to the ith column name fo `col_names`
+  of `input`.
+
+
+  Example:
+
+  '''
+  %0 : !rel_ssa.bag<[!rel_ssa.schema_element<"a0", !rel_ssa.int32>]> = rel_ssa.aggregate(%0 : !rel_ssa.bag<[!rel_ssa.schema_element<"id", !rel_ssa.int32>]>) ["col_names" = ["id"], "functions" = ["sum"]]
+  '''
+  """
+  name = "rel_ssa.aggregate"
+
+  input = OperandDef(Bag)
+  col_names = AttributeDef(ArrayOfConstraint(StringAttr))
+  functions = AttributeDef(ArrayOfConstraint(StringAttr))
+  result = ResultDef(Bag)
+
+  def verify_(self) -> None:
+    # TODO: Check that the same number of col_names and functions are given?
+    for f in self.functions.data:
+      if not f.data in ["sum"]:
+        raise Exception(f"function {f.data} is not a supported function")
+
+  @builder
+  @staticmethod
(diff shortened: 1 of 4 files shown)


experimental/sql/dialects/rel_ssa.py
```diff
@@ -304,6 +304,51 @@ def get(input: Operation, predicates: Region) -> 'Select':
...
+        attributes={
+            "col_names":
+                ArrayAttr.from_list([StringAttr.from_str(c) for c in col_names]
+                                   ),
+            "functions":
+                ArrayAttr.from_list([StringAttr.from_str(f) for f in functions])
+        },
+        result_types=[
+            Bag.get([Int32()] * len(functions),
+                    ["a" + str(i) for i in range(len(functions))])
```
webmiche, line 348: Currently, this names the new columns `a0, a1, a2, ...`. What is the standard name of such columns? @ingomueller-net 
Should the name of these columns be part of the attributes of this operation? Or will they nested into renames anyways?

ingomueller-net: I *think* that there is no standard way. In SQL, the user can give a new name to the resulting column, which I bet Ibis does as well. If it is omitted, I remember some systems doing what you did and others trying to turn the computation into some readable name (`sum(a)` --> `sum_a` or even `sum(a)`).

What is super important, though, is that you catch name clashes with names that already exist. In your IR, are the column names of the *input* unique? If yes, you can just reuse them as result names, no?

ingomueller-net (approved):

experimental/sql/dialects/rel_ssa.py
webmiche: Ah yes, that's the best way to do it, thanks for the suggestion!

experimental/sql/dialects/rel_ssa.py
```diff
@@ -304,6 +304,51 @@ def get(input: Operation, predicates: Region) -> 'Select':
...
+class Aggregate(Operator):
+  """
+  Applies the ith function of `functions` to the ith column name fo `col_names`
+  of `input`.
+
+
+  Example:
+
+  '''
+  %0 : !rel_ssa.bag<[!rel_ssa.schema_element<"a0", !rel_ssa.int32>]> = rel_ssa.aggregate(%0 : !rel_ssa.bag<[!rel_ssa.schema_element<"id", !rel_ssa.int32>]>) ["col_names" = ["id"], "functions" = ["sum"]]
```
nicolasvasilache, line 317: I do not understand this modeling. 
Where is `a0` coming from in the result type `!rel_ssa.bag<[!rel_ssa.schema_element<"a0", !rel_ssa.int32>]>` ?

Also, this performs a reduction but the types for the input and output seem to be the same which feels fishy.

webmiche: Yes the types do seem fishy in this case, but it actually happens to be correct here since the sum reduction results in another integer.

The a0 hints at a problem that we have with reductions: Since they create a column, we have to name that column. But the name should be unique, so we need to be careful. The current version just gives the same name as the input column since that column won't be part of the result anymore. It seems that I forgot to update this in the documentation.

PR #470, opened 2022-05-03
Purge vector - warp related code that is being upstreamed and creates…

nicolasvasilache:
… build conflicts.

This will allow bumping LLVM more easily.


nicolasvasilache:
FYI @ThomasRaoux @MaheshRavishankar @giuseros in case this is relevant to you.

@giuseros, if you expect to continue relying on pass-like CLI strings, you may want to copy the code that this revision deletes somewhere in your directory.

PR #474, opened 2022-05-04
SQL: codegen pipeline

webmiche:
This patch adds a first version of a mirror of the dataflow dialect introduced in #471. Currently, this just mirrors the exact same dialect and does not have any special behavior. However, as soon as we have both PRs merged, we should be able to connect the xdsl and the mlir world within this project.


experimental/sql/dialects/dataflow.py
```diff
@@ -0,0 +1,113 @@
...
+# Data types
+#===------------------------------------------------------------------------===#
+
+#===------------------------------------------------------------------------===#
+# Query types
+#===------------------------------------------------------------------------===#
+
+
+@irdl_attr_definition
+class Tuple(ParametrizedAttribute):
```
webmiche, line 21: AFAIK, you did not specify the tuple type in this dialect. Is the tuple you are using some builtin thing from LLVM, @ingomueller-net?

In that case, I would probably have t add it to xdsl too, s.t. we can actually lower.

ingomueller-net: There is actually a [`TupleType`](https://mlir.llvm.org/docs/Dialects/Builtin/#tupletype) in the "built-in" dialect (defined [here](https://github.com/llvm/llvm-project/blob/3bacd94c312c35684ac5985d37c594acb62b2aa6/mlir/include/mlir/IR/BuiltinTypes.td#L739)).

ingomueller-net (approved):
LGTM!

In the end, I am not splitting the dialects for now. If you replace "dataflow" with "iterators", all should work the same way, though.

webmiche:
This PR currently depends on a new release of xdsl (v.0.4.2). I am working on releasing that version asap.

webmiche:
In order to try using the Python bindings, this current contains ingos commits to expose them. Once this is merged, I will rebase to remove them.

ingomueller-net (approved):
LGTM!

experimental/sql/tools/IteratorsMlirConverter.py
```diff
@@ -0,0 +1,24 @@
+import mlir_iterators.ir as ir
+from mlir_iterators.dialects import iterators as it
+from xdsl.mlir_converter import MLIRConverter
+from xdsl.ir import Attribute
+from dialects.iterators import Stream
+from importlib import import_module
+
+# PYTHONPATH:
+# /home/michel/opencompl/MLIR-lite/ChocoPyCompiler/:/home/michel/MasterThesis/iree-llvm-sandbox/build/tools/sandbox/python_packages/:/home/michel/opencompl/MLIR-lite/xdsl/src:/home/michel/MasterThesis/iree-llvm-sandbox/experimental/sql:/home/michel/opencompl/llvm-project/build/tools/mlir/python_packages/mlir_core
```
webmiche, line 9: Notice that I have both python_bindings versions in my path.

experimental/sql/tools/rel_opt.py
```diff
@@ -3,17 +3,26 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+try:
+  from tools.IteratorsMlirConverter import IteratorsMlirConverter
+  mlir_loaded = True
+except ImportError:
+  mlir_loaded = False
+
 from xdsl.xdsl_opt_main import xDSLOptMain
```
webmiche, line 12: Notice this not obivous dependency: I have to import the `IteratorsMlirConverter` before import `xDSLOptMain`, since that would load the xDSL MlirConverter, registering the normal `mlir` module.

ingomueller-net: Oh, this seems brittle. How about adding the following lines to a (to be created) [`__init__.py`](experimental/sql/src/ibis_mlir/__init__.py)?

```python
import mlir_iterators.ir
from xdsl import load_mlir_module
load_mlir_module(mlir_iterators)
```

This has to be done before anything that could potentially transitively call `ensure_mlir_module_loaded`, and having it in `__init__.py` seems like an easy and sure way to ensure that, no?

webmiche: That seems to have worked, the ordering does not matter anymore :+1:

webmiche:
This work has been merged mainly. All the other things an be found here: [codegen branch](https://github.com/webmiche/iree-llvm-sandbox/tree/codegen_prototype)

PR #477, opened 2022-05-05
Try to bundle iree-dialects under iree.compiler so that sandbox pytho…

nicolasvasilache:
…n can also run with a real iree installation

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,10 +1,12 @@
 include(AddMLIRPython)
 
-set(_addl_sources)
-list(APPEND _addl_sources
-  IREEDialectsPythonSources
-  IREEDialectsPythonExtensions
-)
+################################################################################
+################################################################################
+################################################################################
+# Wrapping sandbox python. 
+################################################################################
+################################################################################
+################################################################################
 
 ################################################################################
 # Sources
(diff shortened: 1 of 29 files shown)


python/CMakeLists.txt
```diff
@@ -23,15 +25,13 @@ declare_mlir_python_extension(IREELLVMSandboxSources.API
...
 ################################################################################
-# Common CAPI
+# CAPI
 ################################################################################
 
-# Bundle our own, self-contained CAPI library with all of our deps.
+# Bundle our own, self-contained CAPI library.
 add_mlir_python_common_capi_library(IREELLVMSandboxCAPI
   INSTALL_COMPONENT IREELLVMSandboxPythonModules
   INSTALL_DESTINATION python_packages/iree_llvm_sandbox/mlir/_mlir_libs
```
stellaraccident, line 37: Needs to be .../iree/compiler/_mlir_libs

python/CMakeLists.txt
```diff
@@ -1,10 +1,12 @@
 include(AddMLIRPython)
```
stellaraccident, line 2: This is load bearing: https://github.com/google/iree/blob/f4c4cced32aa6d3db17f85cf86fc3b43836f22ad/llvm-external-projects/iree-dialects/python/CMakeLists.txt#L5

PR #482, opened 2022-05-11
Iterators: Lower iterators to LLVM+func

ingomueller-net:
This PR replaces the lowering from iterators to the placeholder C++ runtime with a proper lowering through LLVM. The lowering is implemented with a custom walker, which provides the state type of the upstream iterators to the lowering of each iterator op such that these state types can be inlined.

At least for the moment, this PR also uses `llvm.struct`s for the values that are passed from one iterator to the next with `Next` in order to avoid the need for further lowering passes from `tuple`. I plan to revise this decision in one of the very few next PRs.


experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -196,6 +229,530 @@ struct PrintOpLowering : public ConversionPattern {
   }
 };
 
+//===----------------------------------------------------------------------===//
+// Helpers for creating Open/Next/Close functions
+//===----------------------------------------------------------------------===//
+
+/// Creates a new function at the parent module of originalOp with the given
+/// types and name and initializes the function body with a first block.
+FuncOp createOpenNextCloseFunc(Operation *originalOp, RewriterBase &rewriter,
```
nicolasvasilache, line 238: `static` qualifier, here and everywhere we don't want the function to escape.

nicolasvasilache, line 238: The name of this helper can be improved to something more descriptive, e.g. `createPrivateFuncInParentModule`.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -196,6 +229,530 @@ struct PrintOpLowering : public ConversionPattern {
...
+    Value three =
+        rewriter.create<arith::ConstantIntOp>(op->getLoc(), /*value=*/3,
+                                              /*width=*/32);
+    Value hasNext = rewriter.create<arith::CmpIOp>(
+        op->getLoc(), arith::CmpIPredicate::sle, currentIndex, three);
+    scf::IfOp ifOp =
+        rewriter.create<scf::IfOp>(op->getLoc(), opInfo.stateType, hasNext,
+                                   /*withElseRegion=*/true);
+
+    // Fill "then" region of ifOp.
```
nicolasvasilache, line 350: Can we use a much more compact IR building form ?
E.g. mlir/lib/Conversion/VectorToSCF/VectorToSCF.cpp has 
```
    auto check = lb.create<scf::IfOp>(
        resultTypes, cond,
        /*thenBuilder=*/
        [&](OpBuilder &b, Location loc) {
          maybeYieldValue(b, loc, hasRetVal, inBoundsCase(b, loc));
        },
        /*elseBuilder=*/
        [&](OpBuilder &b, Location loc) {
          if (outOfBoundsCase) {
            maybeYieldValue(b, loc, hasRetVal, outOfBoundsCase(b, loc));
          } else {
            b.create<scf::YieldOp>(loc);
          }
        });
```

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -196,6 +229,530 @@ struct PrintOpLowering : public ConversionPattern {
...
+    {
+      OpBuilder::InsertionGuard guard(rewriter);
+      rewriter.setInsertionPointToStart(&ifOp.getThenRegion().front());
+
+      Value one =
+          rewriter.create<arith::ConstantIntOp>(op->getLoc(), /*value=*/1,
+                                                /*width=*/32);
+      Value updatedCurrentIndex = rewriter.create<arith::AddIOp>(
```
nicolasvasilache, line 358: Same here, you may want to use `ArithBuilder` from mlir/include/mlir/Dialect/Arithmetic/Utils/Utils.h

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -196,6 +229,530 @@ struct PrintOpLowering : public ConversionPattern {
...
+  assert(maybeOpInfo.hasValue());
+  auto opInfo = maybeOpInfo.getValue();
+
+  // Create converted state.
+  Value convertedState =
+      rewriter.create<UndefOp>(op->getLoc(), opInfo.stateType);
+
+  // Create Open function. --------------------------------------------------
+  {
+    FuncOp funcOp = createOpenFunc(op, rewriter, opInfo);
```
nicolasvasilache, line 310: It's nice that the function declaration is separate, I would add another level of helpers where the creation of the body of the function is also separate so that the convertIteratorOp only has:
```
createOpenFunc(...);
buildOpenFuncBody(SampleInputOp op);
createNext
etc.
```
In the current form, it is hard to distinguish simple boilerplate from more load-bearing logic.

nicolasvasilache: Each of the "buildxxxBody" should also have some pseudo-IR snippet comment to show what is produced

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -196,6 +229,530 @@ struct PrintOpLowering : public ConversionPattern {
...
+      SmallVector<Location> whileResultLocs = {op->getLoc(), op->getLoc(),
+                                               op->getLoc()};
+      scf::WhileOp whileOp = rewriter.create<scf::WhileOp>(
+          op->getLoc(), whileResultTypes, whileInputs);
+
+      // Fill the "before" region of whileOp.
+      {
+        OpBuilder::InsertionGuard guard(rewriter);
+        Block *before =
+            rewriter.createBlock(&whileOp.getBefore(), {}, whileInputTypes,
```
nicolasvasilache, line 502: We should have (or propose upstream) similar idiomatic builders for this.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -196,6 +229,530 @@ struct PrintOpLowering : public ConversionPattern {
...
+                                                     currentAggregate});
+      } // "before" region
+
+      // Fill the "after" region of whileOp.
+      {
+        OpBuilder::InsertionGuard guard(rewriter);
+        Block *after = rewriter.createBlock(&whileOp.getAfter(), {},
+                                            whileResultTypes, whileResultLocs);
+
+        Value currentState = after->getArgument(0);
```
nicolasvasilache, line 520: Let's do a similar refactoring here first if you don't mind, this will allow seeing what can truly compose at the type level vs what needs Operation* and is limited in its ability to look back into the IR.

nicolasvasilache (changes requested):
Sent a first batch of suggestions to allow refactoring and improving legibility to better decouple impl details from substance.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
```diff
@@ -0,0 +1,94 @@
...
+llvm::Optional<IteratorAnalysis::IteratorInfo>
+IteratorAnalysis::getIteratorInfo(Operation *op) const {
+  auto it = opMap.find(op);
+  if (it == opMap.end())
+    return llvm::None;
+  return it->getSecond();
+}
+
+llvm::SmallVector<SymbolRefAttr, 3>
+IteratorAnalysis::createFunctionNames(Operation *op) {
```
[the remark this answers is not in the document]
ingomueller-net: This and the next function need the member variable `uniqueNumber` in order to produce unique names. I guess that that wasn't clear enough. Maybe because of the name? How about renaming to `findUniqueFunctionNames` and `uniqueFunctionName` (where `unique` in the second name is a verb)?

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
```diff
@@ -0,0 +1,94 @@
...
+    symbols.push_back(symbol);
+  }
+
+  // Increment such that subsequent calls get a different value.
+  uniqueNumber++;
+
+  return symbols;
+}
+
+StringAttr IteratorAnalysis::createUniqueFunctionName(StringRef prefix,
```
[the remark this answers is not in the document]
ingomueller-net: (See discussion above.)

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.h
```diff
@@ -0,0 +1,93 @@
...
+  /// Returns the operation this analysis was constructed from.
+  Operation *getOperation() const { return parentOp; }
+
+  /// Returns the information for the given op.
+  llvm::Optional<IteratorInfo> getIteratorInfo(Operation *op) const;
+
+private:
+  /// Assembles all required information of a given iterator op.
+  template <typename OperatorType>
+  void buildIteratorInfo(OperatorType op);
```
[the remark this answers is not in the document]
ingomueller-net: Yep, good point! Fixed in next commit.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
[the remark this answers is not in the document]
ingomueller-net: Also note that the collision resolution *has* to be done with member variable -- it is *not* sufficient to look up symbols in the module because collisions can occur with Open/Next/Close functions that are only added in the future based on previously pre-assigned names from this analysis.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
```diff
@@ -0,0 +1,94 @@
...
+      break;
+    }
+    uniqueNumber++;
+  }
+  return StringAttr::get(module.getContext(), candidateName);
+}
+
+/// The state of SampleInputOp consists of a single number that corresponds
+/// to the next number returned by the iterator.
+LLVM::LLVMStructType IteratorAnalysis::computeStateType(SampleInputOp op) {
```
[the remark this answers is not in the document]
ingomueller-net: This particular function could be static, but in general no. The iterator state *includes* the state of potential upstream iterators (i.e., this analysis builds the *nested* state types), whose types are looked up in the `DenseMap` of the analysis (through `getIteratorInfo`). See `computeStateType(ReduceOp op)` for an example.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
```diff
@@ -0,0 +1,94 @@
...
+  auto elementType =
+      resultType.getElementType().dyn_cast<LLVM::LLVMStructType>();
+  assert(elementType && elementType.getBody().size() == 1);
+  auto counterType = elementType.getBody().front();
+  return LLVM::LLVMStructType::getNewIdentified(
+      op->getContext(), "iterators.sampleInputState", {counterType});
+}
+
+/// The state of ReduceOp only consists of the state of its upstream.
+LLVM::LLVMStructType IteratorAnalysis::computeStateType(ReduceOp op) {
```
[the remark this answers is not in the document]
ingomueller-net: After looking at this function again, what do you think was missing to understand what the function does on first sight? Maybe a more detailed description of the intent of the analysis at the class documentation of `IteratorAnalysis`?

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.h
```diff
@@ -0,0 +1,93 @@
...
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/Optional.h"
+#include "llvm/ADT/SmallVector.h"
+
+namespace mlir {
+namespace iterators {
+
+/// Constructs information about the state type and the Open/Next/Close
+/// functions of all iterator ops nested inside the given parent op.
+class IteratorAnalysis {
```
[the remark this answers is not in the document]
ingomueller-net: Yes, it *does* create the nested types. I guess a more detailed explanation of this intention is missing here? (See two lines below: `State of the iterator including the state of its potential upstreams.`)

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
ingomueller-net: Thanks! Done in next commit.

ingomueller-net: Sure! Done in next commit. I didn't make `private` part of the name but clarified that all `Open`/`Next`/`Close` functions are private in the function documentation.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
ingomueller-net: Ah, nice pointer! I didn't know about this! Done in next commit.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
ingomueller-net: OK, done in quite a big refactor among the latest commits. This was a good investment! Check out whether you like it :)

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
ingomueller-net: OK, I implemented a helper for myself for the time being in one of the latest commits.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
ingomueller-net: Not sure I fully understand. I have done refactorings to (1) have a generic part that works on `Operation *` that takes care of a lot of plumbing and (2) extract the op-specific conversion logic into `build*Body` functions, which work on CRTP `Op` types. I guess you are referring to this `Operation *` vs `Op` question, right?

ingomueller-net:
@nicolasvasilache: I believe I have addressed all of your comments, here in the review and what we discussed offline. The code has changed significantly -- I recommend that you briefly skim the commit history since your first review but then do a review of the full PR from scratch.

nicolasvasilache (changes requested):
Added a second batch of comments, much better already, thanks!

I still need to look deeper into the reduce.next and the sink cases.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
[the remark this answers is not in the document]
ingomueller-net: OK, I the last commit adds some high-level documentation. I decided to add most of it to the dialect and the conversion pass; `IteratorAnalysis` is only a means to implement `IteratorstoLLVM` and most of the background missing here also required there. Let me know if you disagree or have other ideas to improve the documentation further.

ingomueller-net (commented):
Cool, thanks for the additional input! I think I addressed all points -- let me know what you think. 

(I also finally found how to *answer* a batch of comments, so this should reduce the number of messages for such large PRs... 😜 )

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -296,26 +233,756 @@ struct PrintOpLowering : public ConversionPattern {
...
 
+//===----------------------------------------------------------------------===//
+// SampleInputOp.
+//===----------------------------------------------------------------------===//
+
+/// Builds IR that resets the current index to 0. Possible result:
+///
+/// %0 = llvm.mlir.constant(0 : i32) : i32
+/// %1 = llvm.insertvalue %0, %arg0[0 : index] :
+///          !llvm.struct<"iterators.sampleInputState", (i32)>
```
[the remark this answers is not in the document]
ingomueller-net: I just realize that I *have* introduced snakeCase for SSA value names all over the place. Let me still fix that in this PR.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -9,34 +9,65 @@
...
-#include "mlir/Dialect/Utils/StructuredOpsUtils.h"
+#include "mlir/Dialect/SCF/SCF.h"
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/Block.h"
+#include "mlir/IR/BlockAndValueMapping.h"
+#include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Location.h"
```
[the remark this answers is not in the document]
nicolasvasilache: https://llvm.org/docs/CodingStandards.html#include-as-little-as-possible 

```
[#include as Little as Possible](https://llvm.org/docs/CodingStandards.html#id36)
#include hurts compile time performance. Don’t do it unless you have to, especially in header files.
```

plus caveats follow.

It does boil down to trial, error and building experience with the codebase.

nicolasvasilache:
landing this and iterating on followups in-tree

PR #487, opened 2022-05-12

ingomueller-net:
Replaces #485, which I pushed to the wrong repository.

This now includes #488, which should be merged first once ready.


nicolasvasilache:
Hmm .. IWYU is a google thing and actually goes against the LLVM coding standards which we use: https://llvm.org/docs/CodingStandards.html#include-style

ingomueller-net:
That's a good pointer -- I didn't know about it!

I am not sure, though, whether what I did is discouraged by that rule, though: when it says that the [list of includes should be minimal](https://llvm.org/docs/CodingStandards.html#minimal-list-of-includes), it mainly refers to not including a file if a forward declaration is sufficient. The iwyu tool does that automatically in some cases (and my commit contains a few instances where it did so). The explanation also says:

> You **must** include all of the header files that you are using — you can include them either directly or indirectly through another header file.

The first part of that sentence pretty much paraphrases "include what you use", and, as I understand it, the second part of the sentence only relativizes that as far as to say you *may* include things directly or indirectly. Am I missing something?

ingomueller-net:
Two more thoughts:

1. If I can choose, I prefer to follow iwyu. The latest breakage after updating IREE/LLVM was due to not doing so.
2. What do you think of using [`IncludeBlocks: Merge`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) in `experimental/iterators/.clang-format` (or `.clang-format`) in combination with `IncludeCategories` that correspond to the [recommended order](https://llvm.org/docs/CodingStandards.html#include-style)?

nicolasvasilache:
My recollection is that we generally strongly prefer transitive includes in MLIR.
See for example how many times "mlir/IR/Builders.h" is included, although it is absolutely ubiquitous ..

Adding @ftynse for a second set of eyes.

Generally I would not change .clang-format to differ form anything in core MLIR.

nicolasvasilache:
Landed a bunch of other cleanups in #507 closing this one that will be hard to rebase.

PR #488, opened 2022-05-12
Iterators: Dedicated CI target + fix build

ingomueller-net:


ingomueller-net:
The first commit fails "on purpose". On the current main, building the iterators sub-project would break *if it were executed* but it is not. The first commit just enables that sub-project, the second one fixes the build.

nicolasvasilache (approved):

PR #489, opened 2022-05-12
Fix IREE test.

nicolasvasilache:
This requires IREE using an LLVM version that does not crash when registering ops multiple times in the transform dialect (incoming).
Additionally, the roundtripping through a file is currently broken in core and requires manual intervention (i.e. manually prefixing transform op names with `transform.`).
    
Once both are fixed, the test passes.


ftynse (approved):

ftynse:
I can't reproduce the issue, you may be running some version of IREE that I don't have. As far as I can see, the registration shouldn't be a problem since we do have the op here https://github.com/google/iree/blame/main/compiler/src/iree/compiler/Codegen/TransformDialectExtensions/TransformDialectExtensions.cpp#L189 and the overall setup is to run the `iree-compiler` binary that calls that function.

PR #493, opened 2022-05-14
SQL: impl to iterators conversion

webmiche:
This patch adds a first version of the conversion from rel_impl to the iterators dialect.

### DIFF
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -50,6 +50,7 @@ class PandasTableRewriter(RelImplRewriter):
   @op_type_rewrite_pattern
   def match_and_rewrite(self, op: RelImpl.PandasTable,
                         rewriter: PatternRewriter):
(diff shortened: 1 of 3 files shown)


ingomueller-net (approved):
LGTM!

experimental/sql/src/impl_to_iterators.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+@dataclass
+class PandasTableRewriter(RelImplRewriter):
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelImpl.PandasTable,
+                        rewriter: PatternRewriter):
+    rewriter.replace_matched_op(
+        it.SampleInputOp.get(self.convert_bag(op.result.typ)))
```
ingomueller-net, line 54: We could be a bit conservative here and only replace a pandas table if its name is "sampleInput". If that's too big a hassle, I'd still add a TODO here...

PR #494, opened 2022-05-16
SQL: MLIRConverter

webmiche:
This patch adds the IteratorsMlirConverter, a child of the standard [MLIRConverter](https://github.com/xdslproject/xdsl/blob/main/src/xdsl/mlir_converter.py) of xDSL. The tests are currently not run on the CI due to an ongoing discussion on where to run them exactly.


experimental/sql/test_mlir/lit.cfg
```diff
@@ -0,0 +1,27 @@
...
+else:
+  config.environment["PYTHONPATH"] = config.test_source_root + "/../"
+
+config.environment[
+    "PATH"] = config.test_source_root + "/../tools/:" + os.environ["PATH"]
+
+config.available_features = []
+
+if sys.version_info[0] == 3 and sys.version_info[1] == 8:
+  config.available_features.append('python38')
```
ingomueller-net, line 27: What does this do? Is this used in this PR? What about other Python versions?

experimental/sql/tools/rel_opt.py
```diff
@@ -3,8 +3,15 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+try:
+  from tools.IteratorsMlirConverter import IteratorsMLIRConverter
```
ingomueller-net, line 7: What should the capitalization of the file name be? This line shows that the capitalization is not consistent...

ingomueller-net (approved):
LGTM!

experimental/sql/test_mlir/lit.cfg
webmiche: This is a fragment of a functionality we needed in a different repo. Basically, this checks that the python version is 3.8 (`sys.version_info[0] == 3 and sys.version_info[1] == 8`) and if so adds the `python38` feature. Then, in the tests one can use `REQUIRES python38` and the test will not be executed if the feature is not set.

I removed it, since it is not needed for this project.

PR #495, opened 2022-05-16
(WIP) Iterators: Split CI workflow into independent test and build jobs.

ingomueller-net:


ingomueller-net:
The required packing and uploading of the build result takes longer than a cached build! This doesn't make sense...

PR #496, opened 2022-05-16
Consume IREE as submodule by default.

ingomueller-net:
This PR changes the default build process such that IREE (and thus LLVM) is consumed as a git submodule instead of using a specific branch. This freezes the version of these dependencies such that unrelated updates of that branch do not affect them.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "third_party/iree"]
+	path = third_party/iree
+	url = https://github.com/google/iree.git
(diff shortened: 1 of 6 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/iree"]
+	path = third_party/iree
+	url = https://github.com/nicolasvasilache/iree.git
```
nicolasvasilache, line 3: you could take the real iree directly, I was just using my fork because I was maintaining the proper hash there.

nicolasvasilache (approved):
Looks great, thanks!

ingomueller-net:
@webmiche: Note that the build process has changed with the landing of this PR. See the instructions in the README.

PR #501, opened 2022-05-25
Iterators: Replace SampleInputOp by more generic ConstantStreamOp.

ingomueller-net:
This PR depends on and therefore includes #482 and #500. It should be rebased once those PRs are landed.

This replaces `SampleInputOp` with a similar but more general op called `ConstantStreamOp`, which allows to define constant SSA values of type `Stream`. Instead of providing always the same stream, the new op allows to define the values of the stream as array attribute. Furthermore, LLVM structs of many numeric types are supported (instead of only those with a single `i32`). This new op will continue to be mostly useful for testing, but allows to test much more generally than the previous version.


nicolasvasilache (approved):
Landing as is but this is most likely overkill, just taking constant stream directly from a global rather than an attribute would simplify impl a lot.

PR #502, opened 2022-05-26
Iterators: Parametrize ReduceOp with reduce function.

ingomueller-net:
This PR depends on and therefore includes #482, #500, and #501. It should be rebased once those PRs are landed.

This PR extends the `ReduceOp` with a parameter for the reduction logic, i.e., the logic that combines two elements iteratively. This also removes the restriction on element types: only `!llvm.struct<(i32)>` was supported, now all LLVM struct types supported by the other operators are supported as well.


nicolasvasilache (approved):
Nice!

PR #503, opened 2022-05-30
SQL: add support for naming to ibis dialect

webmiche:
When working on projections, I realized that I had not represented an obscure, but crucial component of the selection node. The selection node has a tuple field called `inputs`. In this field the name of a projected column can be found, which is crucial to get the schema for the result right.

This patch adds a list of names to the selection node in the ibis_dialect. I feel that `inputs` represents things that are already represented by other parts of the selection node (like the table), so I decided not to mirror the field itself for the moment.

### DIFF
--- a/experimental/sql/test/frontend/naming.ibis
+++ b/experimental/sql/test/frontend/naming.ibis
@@ -1,21 +1,14 @@
 # RUN: rel_opt.py -f ibis %s | filecheck %s
 
-table['a', 'b']
+table[table['b'].name('d')]
 
-#      CHECK: ibis.selection() ["names" = ["a", "b"]] {
+#      CHECK: ibis.selection() ["names" = ["d"]] {
 # CHECK-NEXT:     ibis.pandas_table() ["table_name" = "t"] {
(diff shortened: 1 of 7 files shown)


experimental/sql/test/frontend/naming.ibis
```diff
@@ -0,0 +1,19 @@
+# RUN: rel_opt.py -f ibis %s | filecheck %s
```
nicolasvasilache, line 1: Can we rename this .py ? 
It wasn't clear to me this actually creates data out of thin air and injects it.

A typical mlir "opt" binary does not do that.

webmiche: Are you referring to the definition of the tables in rel_opt.py?

I think this is a discussion better placed in this [PR](https://github.com/google/iree-llvm-sandbox/pull/505), as this PR shows another drawback of our current filecheck approach.

webmiche: I changed parts of rel_opt in the [PR](https://github.com/google/iree-llvm-sandbox/pull/509). I think this addressed your comment here.

nicolasvasilache: Looks better but I would even go one step further in followup commits.
Let's land and iterate.

PR #504, opened 2022-05-30
SQL: add more comparators to ibis_dialect

webmiche:
This patch adds the greaterEqual comparator to the ibis dialect. The functionality is basically 1-1 with the equals comparator. I am unsure whether functionality could be reused by adding a `Logical` operation and making these operations children of it. I currently only see the two regions definitions that could be moved to a parent class, but I would be happy about suggestions.

### DIFF
--- a/experimental/sql/tools/rel_opt.py
+++ b/experimental/sql/tools/rel_opt.py
@@ -39,6 +39,7 @@ class RelOptMain(xDSLOptMain):
     def parse_ibis(f: IOBase):
       import ibis
       import pandas as pd
+      import numpy as np
 
       connection = ibis.pandas.connect({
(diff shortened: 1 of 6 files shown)


experimental/sql/tools/rel_opt.py
```diff
@@ -39,6 +39,7 @@ def register_all_frontends(self):
     def parse_ibis(f: IOBase):
       import ibis
       import pandas as pd
+      import numpy as np
```
webmiche, line 42: This is needed for the `np.int64(0)` in the queries, since o/w, ibis seems to parse integers as i8.

nicolasvasilache (approved):

PR #507, opened 2022-06-14
Apply cleanups and refactorings for #482.

nicolasvasilache:
This simplifies the overall implementation, reduces API footprints and better exposes dependences coming from def-use traversal requirements directly in the API.


nicolasvasilache:
Landing now, will address comments in followups

PR #537, opened 2022-07-08
Iterators-filter-op

ingomueller-net:
This op structurally almost identical with ReduceOp (expect of course what it lowers to).


ingomueller-net:
Closing in favor of #539. (This PR was done from a branch in the wrong repo, plus #539 contains several related ops that can be reviewed more conveniently together.)

PR #539, opened 2022-07-11
Iterators: Create FilterOp and MapOp.

ingomueller-net:
This PR implements to new iterators (`map` and `filter`) with an almost identical structure as the existing `reduce` iterator.

### DIFF
--- a/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
+++ b/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
@@ -105,7 +105,7 @@ class LookupFuncType<string opName, string symbolName>
             "       .getFunctionType()">;
 
 /// A predicate, that is, a function with a signature of the form (T) -> i1.
-class Iterators_Predicate
+class Iterators_PredicateFunctionType
   : Type<And<[FunctionType.predicate,
               CPred<"$_self.dyn_cast<FunctionType>().getInputs().size() == 1">,
               CPred<"$_self.dyn_cast<FunctionType>().getResults().size() == 1">,
@@ -117,14 +117,14 @@ class Iterators_Predicate
          "FunctionType">;
 
 /// A FlatSymbolRef referring to a predicate.
-def Iterators_PredicateRefAttr
+def Iterators_PredicateFunctionSymbol
   : Confined<FlatSymbolRefAttr, [
         ReferToOp<"func::FuncOp">,
         AttrConstraint<
           SubstLeaves<"$_self",
                       LookupFuncType<"_op", "$_self.dyn_cast<FlatSymbolRefAttr>()">.result,
-                      Iterators_Predicate<>.predicate>,
-          "referring to a " # Iterators_Predicate<>.summary>
+                      Iterators_PredicateFunctionType<>.predicate>,
+          "referring to a " # Iterators_PredicateFunctionType<>.summary>
       ]>;
 
 def Iterators_FilterOp : Iterators_Op<"filter",
@@ -145,7 +145,7 @@ def Iterators_FilterOp : Iterators_Op<"filter",
   }];
   let arguments = (ins
       Iterators_StreamOfLLVMStructOfNumerics:$input,
-      Iterators_PredicateRefAttr:$predicateRef
+      Iterators_PredicateFunctionSymbol:$predicateRef
     );
   let results = (outs Iterators_StreamOfLLVMStructOfNumerics:$result);
   let extraClassDeclaration = [{
@@ -157,7 +157,7 @@ def Iterators_FilterOp : Iterators_Op<"filter",
 }
 
 /// A predicate, that is, a function with a signature of the form (T) -> i1.
-class Iterators_MapFunc
+class Iterators_MapFunctionType
   : Type<And<[FunctionType.predicate,
               CPred<"$_self.dyn_cast<FunctionType>().getInputs().size() == 1">,
               CPred<"$_self.dyn_cast<FunctionType>().getResults().size() == 1">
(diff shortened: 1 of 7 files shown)


experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -104,6 +104,58 @@ class LookupFuncType<string opName, string symbolName>
             "    " # symbolName # ".dyn_cast<FlatSymbolRefAttr>())"
             "       .getFunctionType()">;
 
+/// A predicate, that is, a function with a signature of the form (T) -> i1.
+class Iterators_Predicate
```
nicolasvasilache, line 108: I'd rename these 2 PredicateFunctionType + PredicateFunctionSymbol to be more idiomatic.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
```diff
@@ -93,11 +101,11 @@ mlir::iterators::IteratorAnalysis::IteratorAnalysis(Operation *rootOp)
           auto stateType = computeStateType(op);
           setIteratorInfo(op, IteratorInfo(op, nameAssigner, stateType));
         })
-        /// The state of ReduceOp only consists of the state of its upstream
-        /// iterator, i.e., the state of the iterator that produces its input
-        /// stream.
-        // TODO: ReduceOp verifier that op.input does not come from a bbArg.
-        .Case<ReduceOp>([&](auto op) {
+        /// The respective state of FilterOp and ReduceOp only consist of the
```
nicolasvasilache: I would make this comment generic in the op name now, to avoid updating each and every time a new op gets added with this property.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -435,6 +435,191 @@ static Value buildStateCreation(ConstantStreamOp op, RewriterBase &rewriter,
...
+
+  // Extract upstream state.
+  Type upstreamStateType = upstreamInfos[0].stateType;
+  Value initialUpstreamState =
+      createExtractValueOp(rewriter, loc, upstreamStateType, initialState, {0});
+
+  // Main while loop.
+  Type i1 = rewriter.getI1Type();
+  SmallVector<Type> nextResultTypes = {upstreamStateType, i1, elementType};
+  scf::WhileOp whileOp = scf::createWhileOp(
```
nicolasvasilache, line 518: As a followup you could start using ImplicitLocOpBuilder and call it just `b`.
Some if these would become 1-liners and generally improve legibility I think.

nicolasvasilache (commented):
Looks FilterOp looks good, looking at MapOp now.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -156,6 +156,57 @@ def Iterators_FilterOp : Iterators_Op<"filter",
   }];
 }
 
+/// A predicate, that is, a function with a signature of the form (T) -> i1.
+class Iterators_MapFunc
```
nicolasvasilache, line 160: Same comment as before re. idiomatic name of these classes.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
```diff
@@ -101,11 +109,11 @@ mlir::iterators::IteratorAnalysis::IteratorAnalysis(Operation *rootOp)
           auto stateType = computeStateType(op);
           setIteratorInfo(op, IteratorInfo(op, nameAssigner, stateType));
         })
-        /// The respective state of FilterOp and ReduceOp only consist of the
-        /// state of its upstream iterator, i.e., the state of the iterator that
-        /// produces its input stream.
+        /// The respective state of FilterOp, MapOp, and ReduceOp only consist
```
nicolasvasilache, line 112: yeah let's avoid that going forward plz

nicolasvasilache (approved):

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
ingomueller-net: The commit *"Make signature and usage of computeStateType independent of Op type."* in #554 fixes this, which you approved in the meantime, so I'll consider this solved.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
ingomueller-net: Dito: solved by #554.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
ingomueller-net: Aha, I see, yes, that would avoid having to carry `loc` around everywhere and shorten a few lines.

Before I can do that, I'd have to change the `build{Open,Next,Close}Body` function signatures, because they currently carry a `RewriterBase`, which `ImplicitLocOpBuilder` doesn't derive from. But I think that that's a good idea anyways: I *think* I don't need the rewriter functionally (and thus the functions are called `build*` and not `rewrite*`).

How would you handle the `scf::createWhileOp` helper and similar that we created? They (1) take a `Location` argument and (2) take function arguments with a `Location` in their signature. Add a variant without `Location` argument in both places?

ingomueller-net (commented):
Awesome, thanks for the review! Addressed one issue with an additional commit, another one is addressed in #554 already, and I'll address the final one in a subsequent PR.

nicolasvasilache:
You should be able to construct an ImplicitLocOpBuilder from OpBuilder + loc wherever you need it

PR #540, opened 2022-07-13
SQL: update reduce of iterators

webmiche:
This PR updates the reduceOp of the iterators dialect mirror. Currently, this does not pass the CI as [the LLVM dialect PR](https://github.com/xdslproject/xdsl/pull/133) in xdsl needs to be merged (and released) first. Feel free to already comment on the code though.


webmiche:
This currently depends on and include #541.

experimental/sql/src/impl_to_iterators.py
```diff
@@ -50,24 +53,84 @@ class FullTableScanRewriter(RelImplRewriter):
...
+        it.ReduceOp.get(op.input.op, StringAttr.from_str("sum_struct"),
+                        self.convert_bag(op.result.typ)))
 
 
 #===------------------------------------------------------------------------===#
 # Conversion setup
 #===------------------------------------------------------------------------===#
 
 
+def add_reduce_functions(ctx: MLContext, mod: ModuleOp):
```
ingomueller-net, line 80: It looks like you intend to predefine all possible reduce functions. However, I think that that approach doesn't scale: there is an infinite number of possible schemas we'd want to be able to run on and each of them would result in a different function (per SQL function per field). I was imagining your lowering to produce a reduce function specifically for each reduce op that it lowers (with local visibility and with a unique name that may or may not reflect what it does, as we discussed).

If this is what you'll want to do eventually and this is just a shortcut, I suggest you document this (with a TODO).

experimental/sql/src/impl_to_iterators.py
```diff
@@ -50,24 +53,84 @@ class FullTableScanRewriter(RelImplRewriter):
...
+                          [IntegerType.from_width(32)])
+                  ],
+                  lambda ba1, ba2: [
+                      l := LLVMExtractValue.build(
+                          result_types=[IntegerType.from_width(32)],
+                          attributes={
+                              "position":
+                                  ArrayAttr.from_list(
+                                      [IntegerAttr.from_index_int_value(0)])
+                          },
```
ingomueller-net, line 104: Maybe you can bind the integer type and the index attribute to a local variable and factor those expressions out to avoid repetition.

ingomueller-net (approved):
LGTM modulo two small comments.

experimental/sql/dialects/iterators.py
```diff
@@ -5,7 +5,7 @@
 from dataclasses import dataclass
 from typing import Any, Union, List
 from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
-from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr, IntegerType, TupleType
+from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr, IntegerType, TupleType, FlatSymbolRefAttr
```
stellaraccident, line 8: Nit, I usually find long imports like this easier to read if put in parens, one per line. My 2 cents.

ingomueller-net: @stellaraccident: Thanks for the suggestion! It has actually sparked a [discussion](https://github.com/xdslproject/xdsl/issues/147) on what convention to use in `xdsl`. Can you recommend a tool that applies the formatting you suggest automatically?

PR #541, opened 2022-07-13
SQL: update to xdsl v0.5

webmiche:
This PR updates the project to work with xdsl v0.5, which gives on one hand more IDE support and on the other hand the crucial LLVMStructType.

Currently, xdsl v0.5 is not yet released, so the CI currently breaks.

### DIFF
--- a/.github/workflows/testSQLDialect.yml
+++ b/.github/workflows/testSQLDialect.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: 3.10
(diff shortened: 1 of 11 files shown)


ingomueller-net (approved):
LGTM!

webmiche:
In order to make xdsl v0.5.1 work, I had to go to python v 3.10. It might make sense to move the other CI scripts to that version in order to prevent confusion later on @nicolasvasilache @ingomueller-net

ingomueller-net:
SGTM! I'll wait for Nicolas' opinion before doing the change...

PR #542, opened 2022-07-17
SQL: add filter to iterators

webmiche:
This PR adds the filter op and its lowering from impl. It also adds a lot of expression rewriters, like Compare, Literal and Yield.

### DIFF
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -8,7 +8,7 @@ from typing import List, Type, Optional
 from xdsl.dialects.builtin import ArrayAttr, StringAttr, ModuleOp, IntegerAttr, IntegerType, TupleType
 from xdsl.dialects.llvm import LLVMStructType, LLVMExtractValue, LLVMInsertValue
 from xdsl.dialects.func import FuncOp, Return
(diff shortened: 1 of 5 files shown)


experimental/sql/src/impl_to_iterators.py
```diff
@@ -125,8 +220,21 @@ def impl_to_iterators(ctx: MLContext, query: ModuleOp):
       it.SinkOp.get(query.body.blocks[0].ops[0].body.blocks[0].ops[-1]))
   # Adding the return
   query.body.blocks[0].ops[0].body.blocks[0].add_op(Return.get())
-  walker = PatternRewriteWalker(GreedyRewritePatternApplier(
-      [FullTableScanRewriter(), AggregateRewriter()]),
+  index_walker = PatternRewriteWalker(GreedyRewritePatternApplier([
```
webmiche, line 223: I need to rewrite all the `IndexByName`s first, since otherwise, their respective operand type does not contain the schema anymore.

ingomueller-net: Makes sense. Add a comment?

ingomueller-net (approved):
LGTM!

PR #545, opened 2022-07-21
SQL: add yield_value and yield_tuple 

webmiche:
This PR diversifies the semantics of the current yield. It splits it into `yield_value`, to yield values such as booleans for filters, and `yield_tuple` to yield results for maps, e.g.

### DIFF
--- a/experimental/sql/dialects/rel_impl.py
+++ b/experimental/sql/dialects/rel_impl.py
@@ -295,8 +295,17 @@ class Compare(Expression):
                           result_types=[Boolean()])
 
 
-@irdl_op_definition
 class Yield(Expression):
+  """
+  Parent class of operations that gridge the gap from expressions back to
+  operators by yielding the result of an expression to the encompassing
+  operator.
+  """
+  ...
+
+
+@irdl_op_definition
+class YieldValue(Expression):
   """
   Bridges the gap from expressions back to operators by yielding the result of
   an expression to the encompassing operator.
@@ -304,17 +313,39 @@ class Yield(Expression):
   Example:
 
   '''
-  rel_impl.yield(%0 : !rel_impl.bool)
+  rel_impl.yield_value(%0 : !rel_impl.bool)
   '''
   """
-  name = "rel_impl.yield"
+  name = "rel_impl.yield_value"
+
+  op = OperandDef(AnyAttr())
+
+  @staticmethod
+  @builder
+  def get(op: Operation) -> 'YieldValue':
+    return YieldValue.create(operands=[op.result])
+
+
+@irdl_op_definition
+class YieldTuple(Expression):
+  """
+  Bridges the gap from expressions back to operators by yielding the result of
+  an expression to the encompassing operator.
+
+  Example:
+
+  '''
+  rel_impl.yield_tuple(%0 : !rel_impl.int32)
+  '''
+  """
+  name = "rel_impl.yield_tuple"
 
   ops = VarOperandDef(AnyAttr())
 
   @staticmethod
   @builder
-  def get(ops: list[Operation]) -> 'Yield':
-    return Yield.create(operands=[o.result for o in ops])
+  def get(ops: list[Operation]) -> 'YieldTuple':
+    return YieldTuple.create(operands=[o.result for o in ops])
 
 
 @irdl_op_definition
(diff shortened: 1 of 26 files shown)


experimental/sql/test/impl_to_iterators/map.xdsl
```diff
@@ -20,6 +20,7 @@ module() {
 // CHECK-NEXT:     ^0(%{{.*}} : !llvm.struct<"", [!i32]>):
 // CHECK-NEXT:       %{{.*}} : !i32 = llvm.extractvalue(%{{.*}} : !llvm.struct<"", [!i32]>) ["position" = [0 : !index]]
 // CHECK-NEXT:       %{{.*}} : !i32 = arith.addi(%{{.*}} : !i32, %{{.*}} : !i32)
+// CHECK-NEXT:       %{{.*}} : !llvm.struct<"", [!i32]> = llvm.mlir.undef()
 // CHECK-NEXT:       %{{.*}} : !llvm.struct<"", [!i32]> = llvm.insertvalue(%{{.*}} : !llvm.struct<"", [!i32]>, %{{.*}} : !i32) ["position" = [0 : !index]]
```
ingomueller-net, line 24: I am not an expert with `FileCheck` and much less one on best practices with it but it seems odd like the line with `insertvalue` has not changed. I think that this is because you only use wildcards (`{{.*}}`) and not variable binding. With the latter, you'd have `%[[V1.*]] : ... llvm.mlir.undef()` to bind the name of the SSA value to `V1` and then `... llvm.insertvalue(%[[V1] : ...) ...` to match that name in subsequent lines. This way you can check whether `insertvalue` uses the correct operands. The downside, of course, is that it takes (even) longer to write these tests...

If you do want to change your tests to the stricter form, don't do it all at once. Just apply the new style gradually over time...

ingomueller-net (approved):
LGTM!

experimental/sql/dialects/rel_impl.py
```diff
@@ -295,26 +295,57 @@ def get(left: Operation, right: Operation,
...
+  @staticmethod
+  @builder
+  def get(op: Operation) -> 'YieldValue':
+    return YieldValue.create(operands=[op.result])
+
+
+@irdl_op_definition
+class YieldTuple(Yield):
   """
   Bridges the gap from expressions back to operators by yielding the result of
```
ingomueller-net, line 332: Explain the concept of a tuple and how the variadic results implicitly form one?

experimental/sql/test/impl_to_iterators/map.xdsl
webmiche: I think you are right, I should probably restructure this, since it does matter here, whether we return the result of the `llvm.mlir.undef` or the result of the `insertvalue`.

ingomueller-net (approved):
LGTM! I think that this is a pretty clean solution; I am happy with it 🙂

experimental/sql/src/alg_to_ssa.py
```diff
@@ -7,7 +7,9 @@
 from typing import List, Type, Optional
 from xdsl.dialects.builtin import ArrayAttr, StringAttr, ModuleOp, IntegerAttr
 
-from xdsl.pattern_rewriter import RewritePattern, GreedyRewritePatternApplier, PatternRewriteWalker, PatternRewriter, op_type_rewrite_pattern
+from xdsl.pattern_rewriter import (RewritePattern, GreedyRewritePatternApplier,
```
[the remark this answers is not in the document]
webmiche: I don't quite understand. Do you mean one pair of brackets per line? :thinking:

experimental/sql/test/impl_to_iterators/map.xdsl
webmiche: It seems that filecheck (a python reimplementation of FileCheck, that does not require compiling llvm) cannot handle these kinds of matchings. I added this style of testing to the `test_mlir` folder, that contains other tests that would require building LLVM, but are not yet supported by our CI.

I also found a small issue by testing like this, so that was worth it :)

experimental/sql/src/alg_to_ssa.py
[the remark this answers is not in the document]
ingomueller-net: No, I think the idea is to have it in this form:

```python
from xdsl.pattern_rewriter import (
    RewritePattern,
    GreedyRewritePatternApplier,
    # ...
)
```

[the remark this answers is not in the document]
webmiche: I will keep the current style with brackets for long imports relying on yapf to do some formatting. I will probably adhere to what we decide for xdsl, once we actually decide something :)

PR #546, opened 2022-07-21
SQL: datatype conversion on the iterators level

webmiche:
This PR adds the conversion of all `decimal`, `timestamp` and `string` to the conversion to iterators. It also adds lowering for decimal and timestamp literals.

### DIFF
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -37,7 +37,10 @@ class RelImplRewriter(RewritePattern):
     if isinstance(type_, RelImpl.Timestamp):
       return IntegerType.from_width(32)
     if isinstance(type_, RelImpl.String):
-      return LLVMStructType(["", [IntegerType.from_width(8)] * 8])
+      return LLVMStructType([
+          StringAttr.from_str(""),
+          ArrayAttr.from_list([IntegerType.from_width(8)] * 8)
+      ])
     raise Exception(f"type conversion not yet implemented for {type(type_)}")
 
   def convert_bag(self, bag: RelImpl.Bag) -> it.Stream:
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -37,6 +37,8 @@ class RelImplRewriter(RewritePattern):
     if isinstance(type_, RelImpl.Timestamp):
       return IntegerType.from_width(32)
     if isinstance(type_, RelImpl.String):
+      # TODO: This is a shortcut to represent strings in some way. Adjust this
+      # to a) non-fixed length strings or b) dynamically fixed size strings.
       return LLVMStructType([
           StringAttr.from_str(""),
           ArrayAttr.from_list([IntegerType.from_width(8)] * 8)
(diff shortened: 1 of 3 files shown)


experimental/sql/src/impl_to_iterators.py
```diff
@@ -30,6 +32,15 @@ def convert_datatype(self, type_: RelImpl.DataType) -> Attribute:
...
       # TODO: this is obviously broken, but the backend currently only support i32
       return IntegerType.from_width(32)
+    if isinstance(type_, RelImpl.Decimal):
+      return IntegerType.from_width(32)
+    if isinstance(type_, RelImpl.Timestamp):
+      return IntegerType.from_width(32)
+    if isinstance(type_, RelImpl.String):
+      return LLVMStructType([
+          StringAttr.from_str(""),
+          ArrayAttr.from_list([IntegerType.from_width(8)] * 8)
```
ingomueller-net, line 42: I think that this deserves a comment (being a shortcut, limitations, etc.). Maybe also a runtime error if longer strings are encountered...

experimental/sql/src/impl_to_iterators.py
```diff
@@ -30,6 +32,15 @@ def convert_datatype(self, type_: RelImpl.DataType) -> Attribute:
     if isinstance(type_, RelImpl.Int64):
       # TODO: this is obviously broken, but the backend currently only support i32
       return IntegerType.from_width(32)
+    if isinstance(type_, RelImpl.Decimal):
```
ingomueller-net, line 35: I guess that the following comment is really about some earlier PR: You currently only support a static number of digits before and after the point, right? How many? You need to be careful about doing math with those: some operations, such as multiplication, change the number of required digits (e.g., `0.01 * 0.01 == 0.0001`). If you parametrized the type with the number of digits after the point, you could do "math on types" in order to reflect that. How do you plan to deal with it currently?

Depending on what you do, I suggest you add comments about shortcuts being taken (which is absolutely fine!) and/or rename the type to `DecimalP8S2` (for "decimal with a *precision* of 8 decimal digits and a *scale* of two decimal digits") or similar. (See also the [T-SQL decimal type](https://docs.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql?view=sql-server-ver16) for example.)

ingomueller-net (approved):
LGTM! I guess my comment about decimals is best addressed in some subsequent PR... Do add it to some list of TODOs, though...

experimental/sql/src/impl_to_iterators.py
webmiche: Currently, there is an error if any string is encountered in the LiteralRewriter. It only matches on Ints, Decimals and Timestamp, but not on Strings. 

I will add the comment :)

experimental/sql/src/impl_to_iterators.py
webmiche: I plan to add two parameters to decimal in a subsequent PR. The thing about arithmetic on them is interesting tough. I will need to think about it and investigate what ibis does exactly. I might push a version of that with the parametrization. Added a comment in the BinOp rewrite.

PR #547, opened 2022-07-22
SQL: add mul and sub to rel_impl conversion

webmiche:
This PR adds mul and sub to the conversion from rel_impl to iterators, since Q6 needs mul.


ingomueller-net (approved):
LGTM!

PR #548, opened 2022-07-22
SQL: add and to iterators, fix int64 lowering

webmiche:
This PR adds the lowering of and to the conversion from rel_impl to iterators and fixes the lowering of int64.

### DIFF
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -119,7 +119,6 @@ class LiteralRewriter(RelImplRewriter):
       rewriter.replace_matched_op(
           Constant.from_int_constant(op.value.value, op.value.typ))
     elif isinstance(op.result.typ, RelImpl.Decimal):
-      print(op.value.data)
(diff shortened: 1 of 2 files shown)


experimental/sql/src/impl_to_iterators.py
```diff
@@ -149,7 +157,7 @@ def match_and_rewrite(self, op: RelImpl.IndexByName,
 class CompareRewriter(RelImplRewriter):
 
   def convert_comparator(self, comparator: str) -> int:
-    if comparator == "==":
+    if comparator == "=":
```
ingomueller-net, line 160: Naughty SQL only uses one `=` ;)

ingomueller-net (approved):
Nice! LGTM!

experimental/sql/src/impl_to_iterators.py
webmiche: Yep, this was code I had previously copied from ChocoPy, since we used a similar approach there :sweat_smile:

PR #549, opened 2022-07-22
SQL: fix aggregation result name bug in ssa to impl

webmiche:
This PR fixes a bug in ssa_to_impl, where the result name would always be the same as the input name, even with other names specified for the result.


ingomueller-net (approved):
Good catch! LGTM!

PR #550, opened 2022-07-22
SQL: fix bug in project result type

webmiche:
This PR fixes a nasty problem. So, the rewrite of expression needs the parent operation to be rewritten already, s.t. it knows what types the column operations have. However, the Project operation only knows its result type by introspecting its region, where, well there are no types, since this is still in rel_alg during this stage. This fix crawls over all expression nodes in rel_alg to discover the types if the result columns.

### DIFF
--- a/experimental/sql/src/alg_to_ssa.py
+++ b/experimental/sql/src/alg_to_ssa.py
@@ -141,6 +141,7 @@ class BinOpRewriter(RelAlgRewriter):
 @dataclass
 class ProjectRewriter(RelAlgRewriter):
 
+  # TODO: This could be implemented in a more natural way using Analysis Passes in MLIR.
   def find_type_of_expression(self, op: RelAlg.Expression,
                               input_bag: RelSSA.Bag) -> RelSSA.DataType:
     if isinstance(op, RelAlg.Column):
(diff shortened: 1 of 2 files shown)


experimental/sql/src/alg_to_ssa.py
```diff
@@ -133,6 +133,16 @@ def match_and_rewrite(self, op: RelAlg.BinOp, rewriter: PatternRewriter):
...
 class ProjectRewriter(RelAlgRewriter):
 
+  def find_type_of_expression(self, op: RelAlg.Expression,
+                              input_bag: RelSSA.Bag) -> RelSSA.DataType:
+    if isinstance(op, RelAlg.Column):
+      return input_bag.lookup_type_in_schema(op.col_name.data)
+    if isinstance(op, RelAlg.Literal):
+      return op.type
+    if isinstance(op, RelAlg.BinOp):
+      return self.find_type_of_expression(op.lhs.op, input_bag)
```
ingomueller-net, line 143: I suppose that this works because all `BinOp`s only accept two operands of the same type? Add a comment about this assumption?

ingomueller-net:
Dialect conversion is a complex process and I can't talk with confidence about how things work in MLIR but I *think* that the idea is that types should be converted in a context-free manner. In your situation, I think that a textbook design would allow to inspect the result type of the original project op, use a type converter to convert it to the target dialect, and that would (have to) be the result type of the converted project op. Similarly, the arguments of the inner region would be converted by a type converter (without further context) before the region itself is converted. It seems to me like op and its region could be converted in any order in that case...

Why does such an approach not work here? Maybe your `find_type_of_expression` is actually a degenerated version of that...

ingomueller-net (commented):
(Waiting for your reply on my general comment...)

PR #551, opened 2022-07-22
SQL: parametrize decimals

webmiche:
This PR adds the parametrization of decimals as requested in #546. Currently, there is no change in parameters when operations are performed. I decided to leave this out since, there isn't something like this done for other datatypes aswell. Consider int32. Adding two int32s should give an int33, but MLIR restricts results of addi to be the same as the input type ([here](https://mlir.llvm.org/docs/Dialects/ArithmeticOps/#arithaddi-mlirarithaddiop)).

I am still thinking about how to do this nicely, as I feel like I am in a weird mix of "the top-level should know" vs "aren't these implementation details?". I will probably push a version of something later to this PR.

### DIFF
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -45,6 +45,9 @@ class Decimal(DataType):
   """
   name = "ibis.decimal"
 
+  prec: ParameterDef[IntegerAttr]
+  scale: ParameterDef[IntegerAttr]
+
 
 @irdl_attr_definition
 class Timestamp(DataType):
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -43,7 +43,7 @@ class Decimal(DataType):
 
   https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/datatypes.py#L344
   """
-  name = "ibis.decimal"
+  name = "ibis.decimal<4 : !i32, 2 : !i32>"
 
   prec: ParameterDef[IntegerAttr]
   scale: ParameterDef[IntegerAttr]
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -43,7 +43,7 @@ class Decimal(DataType):
 
   https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/datatypes.py#L344
   """
-  name = "ibis.decimal<4 : !i32, 2 : !i32>"
+  name = "ibis.decimal"
 
   prec: ParameterDef[IntegerAttr]
   scale: ParameterDef[IntegerAttr]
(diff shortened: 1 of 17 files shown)


webmiche:
BTW, ibis just ignores this issue, i.e., multiply two columns of "decimal(4, 2)" stays a "decimal(4, 2)".

ingomueller-net:
> BTW, ibis just ignores this issue, i.e., multiplying two columns of "decimal(4, 2)" stays a "decimal(4, 2)".

I'd say Ibis is wrong ;) See [here](https://www.ibm.com/docs/en/i/7.1?topic=operators-decimal-arithmetic-in-sql) the documentation of DB2 on the topic, and [here](http://sqlfiddle.com/#!9/ad3148a/1) some example on MySQL I quickly hacked together.

> Consider int32. Adding two int32s should give an int33, but MLIR restricts results of addi to be the same as the input type ([here](https://mlir.llvm.org/docs/Dialects/ArithmeticOps/#arithaddi-mlirarithaddiop)).

You are bringing up another interesting point. SQL's semantic aren't trivial here. If I am not mistaken, the SQL standard actually asks to throw an error *at runtime* if/when an overflow occurs, so we'd have to do *runtime checks* in the general case! I haven't managed to find a good reference within a couple of minutes search but the [Postgres](https://www.postgresql.org/docs/current/datatype-numeric.html) documentation says it throws an error if you ask it to "*store* values outside of the allowed range". [This example](http://sqlfiddle.com/#!17/9eecb/94385) corroborates this behavior.

It would, however, by interesting, to do exactly the type of static analysis you say in order to *avoid* these runtime checks. If we know that the maximum of some input columns `a` and `b` fit into `i17` and `i4`, respectively, we can be sure that `a+b` fits into `i22` and thus lower the operation to machine instructions on `i32`s *without runtime checks*. Knowing the exact bitwidths may also be interesting for other optimizations: if we join on multiple keys, say `a : i11` and `b : i3`, we may be able to pack them into fewer machine words using bit fiddling and treat them as one during much of the execution, in this case, a single `i16`.

Long story short: I continue to believe that decimals should be parametrized and that arithmetics on them have non-trivial and hence interesting effects on types. We can, however and as always, decide to take shortcuts if things get too complex.

ingomueller-net:
> BTW, ibis just ignores this issue, i.e., multiplying two columns of "decimal(4, 2)" stays a "decimal(4, 2)".

I have looked for some documentation on arithmetics in Ibis but didn't find any. Maybe it's worth raising an issue asking for the place of that documentation (unless you find it) and point to the difference in behavior with MySQL, DB2, and probably other systems...

webmiche:
I feel like we are going a bit too far for this PR rn, so I would suggest we add a todo in the code and an entry on my todo list to look into decimal arithmetic and keep the discussion on the pure, "static" representation side for this PR :)

I think a fundamental problem of your suggested analysis is that you probably have to inspect all of the data to make sure it fits into small integers, which will cost you so much time that you could just have run the query instead :sweat_smile: 
These are certainly interesting discussions tough and if we were to expose them on the frontend (say we addd functionality that some domain expert can tell us that there are only 128 values an integer column can take) we can certainly store this in a datatype, that is an i7 (backed by the smallest integer our hardware allows usually i16 or i8) and then optimize the runtime checks away by looking at the types, while still doing the arithmetic on i16 or i8 respectively...

This reminds me a lot of the work we did for FPL with Tobias :sweat_smile:

PR #552, opened 2022-07-22
SQL: nullable handling

webmiche:
This PR rewrites nullable handling. For the ibis dialect, it introduces a parameter that shows whether a given datatype is nullable. For all other dialects it introduces a new datatype `nullable` that is parametrized by another datatype.

### DIFF
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -37,6 +37,16 @@ class DataType(ParametrizedAttribute):
 
 
 @irdl_attr_definition
+class Nullable(DataType):
+  """
+  Models the fact that an ibis datatype is nullable.
+  """
+  name = "ibis.nullable"
+
+  datatype: ParameterDef[DataType]
+
+
+@irdl_attr_definition
 class Decimal(DataType):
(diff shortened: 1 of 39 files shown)


ingomueller-net (commented):
What's the rationale for using each of the two approaches, respectively? Why not use the same approach for all dialects?

webmiche:
For ibis I wanted to stay close to how the actual framework works. The framework has types, which are classes with boolean fields, that tell you whether they are nullable or not. I felt that this encoding is the closest to that.

However, I personally prefer representation since I feel it is more intuitive to read and think about. IMO, nullable basically gives "one more value" to a general datatype, the null value. I feel that this can be more easily lowered in the backend, for example some sort of bitmask that tells you which entries are null or something.

I must admit tough that I do not have enough knowledge and experience with how null values are generally handled in databases/on hardware. I think there is also an argument to be had that there should be a `non_nullable` type parametrized with a datatype instead of a `nullable` type since that seems to be the default when just specifying tables, where as `non_nullable` usually requires a specific kind of constaint (I think).

You probably have better intuition on the second part of my answer and I am certainly happy to follow your call!

PR #553, opened 2022-08-08
Iterators: Implement scan op that supports scanning pandas DataFrames

ingomueller-net:


nicolasvasilache (commented):
nice.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -613,10 +613,10 @@ static Value buildCloseBody(FilterOp op, RewriterBase &rewriter,
 ///          !llvm.struct<"iterators.filter_state", (!nested_state)>
 static Value buildStateCreation(FilterOp op, RewriterBase &rewriter,
                                 LLVM::LLVMStructType stateType,
-                                ValueRange upstreamStates) {
+                                ValueRange operands) {
   Location loc = op.getLoc();
   Value undefState = rewriter.create<UndefOp>(loc, stateType);
-  Value upstreamState = upstreamStates[0];
+  Value upstreamState = operands[0];
```
nicolasvasilache, line 707: Can you use `FilterOpOperandAdapter(operands)` here and everywhere to avoid magic constants?
You may need to introduce the helper functions in the extraClass declaration in the op.
It was part of the LLVM cleanups / refactorings I wanted to do while you were away but never got to ..

ingomueller-net: Done as part of #554.

experimental/iterators/test/python/dialects/iterators/dialect.py
```diff
@@ -1,11 +1,16 @@
 # RUN: %PYTHON %s | FileCheck %s
 
+import ctypes
 import os
 
+import pandas as pd
+import numpy as np
+
+from mlir_iterators.runtime.pandas_to_iterators import to_columnar_batch_descriptor
```
webmiche, line 9: On my machine, these kinds of imports seem to crash as it tries to find `runtime` in `mlir_iterators` of the python_bindings. Maybe rename?

ingomueller-net:
@nicolasvasilache: I have cleaned this up enough for a first review from you.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -122,6 +148,66 @@ static Value getOrCreateGlobalString(OpBuilder &builder, Twine name,
...
+  ColumnarBatchFromMemrefsLowering(TypeConverter &typeConverter,
+                                   MLIRContext *context,
+                                   PatternBenefit benefit = 1)
+      : OpConversionPattern(typeConverter, context, benefit) {}
+
+  LogicalResult match(ColumnarBatchFromMemrefsOp op) const override {
+    return success();
+  }
+
+  void rewrite(ColumnarBatchFromMemrefsOp op, OpAdaptor adaptor,
```
nicolasvasilache, line 180: You want to check the memref is 1-D with a canonical layout; if not bail out.
Otherwise, surprises will occur if you ignore the non-0 offset or non-1 strides.

nicolasvasilache: ok I see it is in the verifier so please ignore then.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -50,6 +50,33 @@ def Iterators_PrintOp : Iterators_Base_Op<"print"> {
   let arguments = (ins Iterators_LLVMStructOfNumerics:$element);
 }
 
+def Iterators_ColumnarBatchFromMemrefsOp
+    : Iterators_Base_Op<"columnar_batch_from_memrefs"> {
```
nicolasvasilache, line 54: How about "view_as_columnar_batch" ?
This is more consistent with the fact that there is no underlying alloc.
The from_memref is visible in the type so is redundant.
Additionally, in the future, you may also want to allow tensors.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -292,6 +292,30 @@ def Iterators_ReduceOp : Iterators_Op<"reduce",
   }];
 }
 
+def Iterators_ScanColumnarBatchOp : Iterators_Op<"scan_columnar_batch",
```
nicolasvasilache, line 295: This seems dangerous as it seems to go into an N^2 design, I also see a FullTableScanOp and a ColumnarScanOp.

Why not just a single ScanOp that adapts to the type of its argument?

experimental/iterators/test/Conversion/IteratorsToLLVM/scan-columnar-batch.mlir
```diff
@@ -0,0 +1,51 @@
+// RUN: mlir-proto-opt %s \
+// RUN:   -convert-iterators-to-llvm -reconcile-unrealized-casts \
+// RUN: | FileCheck --enable-var-scope %s
+
+!element_type = !llvm.struct<(i32)>
```
nicolasvasilache, line 5: seems unused ?

experimental/iterators/test/python/dialects/iterators/dialect.py
```diff
@@ -91,3 +96,34 @@ def testEndToEnd():
...
+
+  data = np.array([(0, 3), (1, 4), (2, 5)], dtype=[('a', 'i4'), ('b', 'i8')])
+  df = pd.DataFrame.from_records(data)
+  arg = ctypes.pointer(to_columnar_batch_descriptor(df))
+
+  # CHECK:      (0, 3)
+  # CHECK-NEXT: (1, 4)
+  # CHECK-NEXT: (2, 5)
+  engine = ExecutionEngine(mod)
+  engine.invoke('main', arg)
```
nicolasvasilache, line 129: Nice!

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
ingomueller-net: I agree with the `from_memrefs` part.

I am not sure about `view`. It's not a view in the sense of `ViewLikeOpInterface`. Also, other ops also don't make their view intensions explicit (such as [Vector_TypeCast](https://github.com/llvm/llvm-project/blob/b83d0d/mlir/include/mlir/Dialect/Vector/IR/VectorOps.td#L2098)).

So maybe just `to_columnar_batch` (like `to_tensor` and `to_memref` in bufferization)?

I thought about supporting tensors as well (and had some hacky version implemented at some point) but then didn't see the benefit compared to converting manually with `bufferization.to_memref`. Do you see one? (And how'd you support them? By allowing either tensors or memrefs in the same op?)

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
ingomueller-net: What are the two dimensions? For each memory format, we need a way to iterator over the elements (i.e., "scan" it); that's only one dimension.

I don't have a strong opinion about overloading argument types, yet, but my initial reflex is to have different ops for different input types. I expect them to require relatively different lowerings, so a single op would essentially consists of a large switch and then different logic. (For similar enough formats, where the lowering is essentially the same, or once we have an abstraction that makes them the same, my arguments don't apply, so I'd agree with having a single op.)

experimental/iterators/test/Conversion/IteratorsToLLVM/scan-columnar-batch.mlir
ingomueller-net: Indeed. Thanks! Fixed.

experimental/iterators/test/python/dialects/iterators/dialect.py
ingomueller-net: 🥳

ingomueller-net (commented):
@nicolasvasilache: Thanks for the review! I have fixed all small issues; one question about the naming of `columnar_batch_from_tensors` before I fix that as well plus a question about the N^2 design.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
nicolasvasilache: I wouldn't worry about ViewLikeOpInterface, that is slated to evolve but we only have a very primitive buffer type at the moment.

We can use another name than `view` if you prefer but the key distinction that needs to be visible is whether this copies underlying data or just shuffles pointers around.

Another design possibility if to just call it `columnar_batch` and have a view or copy UnitAttr.

I would avoid cargo culting naming from the bufferization ops as these are exclusively used as "fake casts" during the conversion process.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
nicolasvasilache: `that's only one dimension` yes that's precisely why I am suggesting a ScanOp: we shouldn't add a second dimension that is "just the type": a scan can be polymorphic.

As to the big switch aspect: experience shows that this is a symptom of a missing interface.
Once we start seeing swtiches grow, it makes sense to extract out the interfaces we want to see.
If you already know what the interface looks like even better but that is generally hard to foresee.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
ingomueller-net: I went with your suggestion in the end. I am thinking of `view` as a verb now: view the operand as a columnar batch, which is a bit different that `ViewLikeOpInterface`, I guess. I was thinking of "reinterpret" but that is longer and more complex to say, so ...

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
ingomueller-net: I see that this is an interesting part of the design space. I don't see things clearly enough, yet, though, to foresee what possible design make sense. Can we defer the discussion until we introduce scans for other formats?

ingomueller-net (commented):
@ntv: I have fixed all small issues and renamed `columnar_batch_from_memrefs` as per your suggestion. I am unsure about the "n^2 issue," for which I can't currently foresee the design space really. Can be defer that discussion?

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
ingomueller-net: Notes from an offline meeting with @nicolasvasilache:

* There is a terminology issue with the word `scan`. In functional programming, a scan is a generalized prefix sum; in databases, a scan is the operator that reads the data from a base table. We thus need a better term and document the difference in terminology.
* It is still an open question whether we should have one polymorphic "read base table op" or a dedicated one per data format. The lowering for each format may be different, and there may be different optimizations that apply depending on the format, but maybe some or all of that can be done with interfaces.

webmiche: Connecting this to discussions we currently have in a different PR:
For columnar formats, there is also the dimension of does the scan access all columns of a given table or only parts. So essentially, should we hand the pointers to all the columns and tell the operation which ones to read from, or should we only hand it the ones that it should actually read from.

ingomueller-net: Some thoughts about naming:

* A DB "scan" operator takes a single value, for example a columnar batch, and helps us *iterate* over them. So names like `iterate_columnar_batch` could be appropriate. Python has functions as `iter_values`, `iter_keys`, and `iter_items` on its dictionaries, for example, and I think they serve the same purpose.
* In a similar spirit, these "scan" operators turn a single value into an `iterator.stream`, so `columnar_batch_to_stream` could express that.
* On a different note, there may be different ways to iterate over the same thing: for a columnar batch, one might want to iterate over slices of the batch (`iter_slices`?) in order to do something similar to tiling, or over individual tuples (`iter_tuples`?). In a tensor or memref, one might want to do the same thing: iterate over smaller tensors/memrefs (in different orders) or over individual numbers.

ingomueller-net:
FYI, I have factored out #563, #564, #565, and #566 out of the original PR in order to be able to concentrate on the design aspect here. (#566 still needs to be merged.) I haven't addressed any of the outstanding design issues yet, so nobody other than me needs to take action at this point.

ingomueller-net:
Superseded by #590, #591, #592, #593, #594, #595, #579, and #581.

PR #554, opened 2022-08-08
Iterators: Improve analysis

ingomueller-net:
This PR is based on and therefore includes #539. It extends and cleans up the iterator analysis in preparation for #553. In that PR, we introduce an iterator op (a scan) than has a *non-iterator* operand -- something that doesn't occur with the ops we currently have. This PR makes the analysis correct for ops with non-iterator operands and cleans up a few other things.

### DIFF
--- a/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
+++ b/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
@@ -1168,9 +1168,15 @@ static Value convert(IteratorOpInterface op, ValueRange operands,
   // Assemble IteratorInfo for all the upstream iterators (i.e. all the defs).
   llvm::SmallVector<IteratorInfo, 8> upstreamInfos;
   for (Value operand : op->getOperands()) {
-    auto definingOp = cast<IteratorOpInterface>(operand.getDefiningOp());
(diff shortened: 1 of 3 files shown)


experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
```diff
@@ -613,10 +613,10 @@ static Value buildCloseBody(FilterOp op, RewriterBase &rewriter,
 ///          !llvm.struct<"iterators.filter_state", (!nested_state)>
 static Value buildStateCreation(FilterOp op, RewriterBase &rewriter,
                                 LLVM::LLVMStructType stateType,
-                                ValueRange upstreamStates) {
+                                ValueRange operands) {
   Location loc = op.getLoc();
   Value undefState = rewriter.create<UndefOp>(loc, stateType);
-  Value upstreamState = upstreamStates[0];
+  Value upstreamState = operands[0];
```
nicolasvasilache, line 619: mentioned somehwere else but please try to use OpOperandAdapter if appropriate.

nicolasvasilache:
Skimmed through the code in the WIP progress that includes this and it looked good modulo minor comments.
I can't see differences offhand in this PR so this LGTM.
If there are material differences, please ask me to review again.

nicolasvasilache (approved):

PR #555, opened 2022-08-12
SQL: add iterators scan op

webmiche:
This PR adds the iterators scan op, the way to actually lower `FullTableScan`s to the iterators dialect. The PR relies on the version of the python bindings from #553. With this version, we can now go all the way through in terms of building the IR. Executing the code is still WIP.

### DIFF
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -41,6 +41,8 @@ def convert_datatype(type_: RelImpl.DataType) -> Attribute:
         StringAttr.from_str(""),
         ArrayAttr.from_list([IntegerType.from_width(8)] * 8)
     ])
+  if isinstance(type_, RelImpl.Nullable):
+    return convert_datatype(type_.type)
   raise Exception(f"type conversion not yet implemented for {type(type_)}")
 
 
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -4,7 +4,7 @@
 
 from dataclasses import dataclass
 from xdsl.ir import Operation, MLContext, Region, Block, Attribute
-from typing import List, Type, Optional
+from typing import List, Type, Optional, Tuple
 from xdsl.dialects.builtin import ArrayAttr, StringAttr, ModuleOp, IntegerAttr, IntegerType, TupleType
 from xdsl.dialects.llvm import LLVMStructType, LLVMExtractValue, LLVMInsertValue, LLVMMLIRUndef
 from xdsl.dialects.func import FuncOp, Return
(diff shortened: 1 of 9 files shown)


experimental/sql/src/impl_to_iterators.py
```diff
@@ -23,43 +23,49 @@
 # `Rewriter`s. All other `Rewriter`s inherit from that class.
 
 
+def convert_datatype(type_: RelImpl.DataType) -> Attribute:
```
webmiche, line 26: I decided to move these functions out in order to use them in `get_batch_and_name_list`. While I had an earlier version where this was done using a Rewriter, this version was not very clean IMO, as it matched on function ops, so it would also match on the functions added for map and filter.

experimental/sql/src/impl_to_iterators.py
```diff
@@ -308,12 +312,41 @@ def match_and_rewrite(self, op: RelImpl.Project, rewriter: PatternRewriter):
 #===------------------------------------------------------------------------===#
 
 
+def get_batch_and_name_list(
```
ingomueller-net, line 315: Add a short documentation for this function?

experimental/sql/src/impl_to_iterators.py
```diff
@@ -308,12 +312,41 @@ def match_and_rewrite(self, op: RelImpl.Project, rewriter: PatternRewriter):
...
+      curr_batch = it.ColumnarBatch.get(
+          TupleType([
+              ArrayAttr.from_list([
+                  convert_datatype(e.elt_type) for e in o.result.typ.schema.data
+              ])
+          ]))
+      batches.append(curr_batch)
+      names.append(o.table_name.data)
+
+  return names, batches
```
ingomueller-net, line 330: Return those in the same order as they appear in the function name?

experimental/sql/src/impl_to_iterators.py
```diff
@@ -308,12 +312,41 @@ def match_and_rewrite(self, op: RelImpl.Project, rewriter: PatternRewriter):
...
+  body_block = Block.from_arg_types(batches)
+  body_block.add_ops(
+      [query.body.blocks[0].detach_op(o) for o in query.body.blocks[0].ops])
+  query.body.detach_block(0)
+  f = FuncOp.from_region("main", batches, [],
+                         Region.from_block_list([body_block]))
   query.body.add_block(Block.from_ops([f]))
+  # Populating a mapping from table names to BlockArguments
+  for n, b in zip(names, f.body.blocks[0].args):
+    table_mapping[n] = b
```
ingomueller-net, line 349: Does this work if the same table is scanned twice? It'll appear twice in `names` and `batches` but only once in `table_mapping`. Maybe it *does* but it's a bit obscure...

How about having `get_batch_and_name_list` return an (ordered?) dictionary instead?

ingomueller-net (approved):
LGTM modulo minor comments.

experimental/sql/src/impl_to_iterators.py
webmiche: I think it makes most sense to only load each table once, so only have it once in the returned lists. I added a check that the given table has not been loaded previously.

PR #557, opened 2022-08-16
SQL: projection pushdown

webmiche:
This PR adds a very simple projection pushdown step. Currently, it handles two cases, reflected in the test cases. I think what we will need (and what I need to think about) for Q6 is how we do this for aggregate, since aggregate is similar to project in what it does to schemas and we will need something like this for Q6.

### DIFF
--- a/experimental/sql/src/projection_optimizations.py
+++ b/experimental/sql/src/projection_optimizations.py
@@ -179,7 +179,7 @@ def optimize_projections(ctx: MLContext, query: ModuleOp):
   projection_fuser_walker = PatternRewriteWalker(GreedyRewritePatternApplier(
       [ProjectionFuser()]),
                                                  walk_regions_first=False,
-                                                 apply_recursively=False,
+                                                 apply_recursively=True,
                                                  walk_reverse=False)
   projection_fuser_walker.rewrite_module(query)
 
--- a/experimental/sql/src/projection_optimizations.py
+++ b/experimental/sql/src/projection_optimizations.py
@@ -103,6 +103,47 @@ class ProjectionInference(ProjectionOptimizer):
 
 
 @dataclass
+class ProjectionFuser(ProjectionOptimizer):
+
+  def rename_expr(self, op: RelAlg.Expression, map: dict[str,
+                                                         str]) -> Operation:
+    if isinstance(op, RelAlg.Column):
+      return RelAlg.Column.get(map[op.col_name.data])
+    if isinstance(op, RelAlg.Multiply):
+      return RelAlg.Multiply.get(
+          Region.from_operation_list([self.rename_expr(op.lhs, map)]),
+          Region.from_operation_list([self.rename_expr(op.rhs, map)]))
+    if isinstance(op, RelAlg.Literal):
+      return RelAlg.Literal.get(op.val, op.type)
+    if isinstance(op, RelAlg.Compare):
+      return RelAlg.Multiply.get(
+          op.comparator.data,
+          Region.from_operation_list([self.rename_expr(op.left, map)]),
+          Region.from_operation_list([self.rename_expr(op.right, map)]))
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelAlg.Project, rewriter: PatternRewriter):
+    if not isinstance(op.input.op, RelAlg.Project):
+      return
+
+    child_cols = self.flatten(
+        [self.find_cols_in_expr(o) for o in op.input.op.projections.ops])
+    child_res = [s.data for s in op.input.op.names.data]
+    # If the child projection simply projects onto a subset, we can just infer
+    # its column mapping and apply it on the parents projection.
+    if len(child_cols) == len(child_res) and all(
+        [isinstance(o, RelAlg.Column) for o in op.input.op.projections.ops]):
+      child_dict = dict(zip(child_cols, child_res))
+
+      rewriter.replace_matched_op(
+          RelAlg.Project.get(
+              rewriter.move_region_contents_to_new_regions(op.input.op.input),
+              Region.from_operation_list([
+                  self.rename_expr(o, child_dict) for o in op.projections.ops
+              ]), op.names))
+
+
+@dataclass
 class IdentityProjectionRemover(ProjectionOptimizer):
 
   @op_type_rewrite_pattern
@@ -132,9 +173,16 @@ def optimize_projections(ctx: MLContext, query: ModuleOp):
                                                   walk_reverse=False)
   infer_projections_walker.rewrite_module(query)
 
+  projection_fuser_walker = PatternRewriteWalker(GreedyRewritePatternApplier(
+      [ProjectionFuser()]),
+                                                 walk_regions_first=False,
+                                                 apply_recursively=False,
+                                                 walk_reverse=False)
+  projection_fuser_walker.rewrite_module(query)
+
   identity_projection_walker = PatternRewriteWalker(GreedyRewritePatternApplier(
       [IdentityProjectionRemover()]),
-                                                    walk_regions_first=False,
+                                                    walk_regions_first=True,
                                                     apply_recursively=False,
                                                     walk_reverse=False)
   identity_projection_walker.rewrite_module(query)
--- a/experimental/sql/src/projection_optimizations.py
+++ b/experimental/sql/src/projection_optimizations.py
@@ -124,7 +124,7 @@ class IdentityProjectionRemover(ProjectionOptimizer):
       rewriter.replace_matched_op(new_op)
 
 
-def projection_pushdown(ctx: MLContext, query: ModuleOp):
+def optimize_projections(ctx: MLContext, query: ModuleOp):
   infer_projections_walker = PatternRewriteWalker(GreedyRewritePatternApplier(
       [ProjectionInference()]),
                                                   walk_regions_first=False,
--- a/experimental/sql/src/projection_optimizations.py
+++ b/experimental/sql/src/projection_optimizations.py
@@ -14,9 +14,9 @@ from xdsl.pattern_rewriter import (RewritePattern, GreedyRewritePatternApplier,
 import dialects.rel_alg as RelAlg
 
 # This file contains several rewrites on the rel_alg level, all of which concern
-# projections. Currently, there are two rewriters. The first one introduces a
-# projection before any operator, the second one removes projetions that are
-# just identity mappings.
+# projections. Currently, there are three rewriters. The first one introduces a
+# projection before any operator, the second one fuses subsequent projections,
+# and the third one removes projetions that are just identity mappings.
 
 
 @dataclass
@@ -104,6 +104,9 @@ class ProjectionInference(ProjectionOptimizer):
 
 @dataclass
 class ProjectionFuser(ProjectionOptimizer):
+
+  # This rewriter fuses two subsequent projections. Currently, it only works on
+  # renaming style projections, but it will be extended in the future.
 
   def rename_expr(self, op: RelAlg.Expression, map: dict[str,
                                                          str]) -> Operation:
--- a/experimental/sql/src/projection_pushdown.py
+++ b/experimental/sql/src/projection_pushdown.py
@@ -16,7 +16,7 @@ import dialects.rel_alg as RelAlg
 # This file contains several rewrites that all compose to classic projection
 # pushdown. Currently, this works in three steps and, hence, has three
 # rewriters. The first one introduces a projection before any operator, the
-# second one fuses subsequent projections, and the third one removes projetions
+# second one fuses subsequent projections, and the third one removes projections
 # that are just identity mappings.
 
 
--- a/experimental/sql/src/projection_pushdown.py
+++ b/experimental/sql/src/projection_pushdown.py
@@ -50,6 +50,19 @@ class ProjectionOptimizer(RewritePattern):
     if isinstance(op, RelAlg.Aggregate):
       return [s.data for s in op.col_names.data]
 
+  def find_schema(self, op: RelAlg.Operator) -> list[str]:
+    """
+    Returns the list of cols that are the input schema of this operation.
+    """
+    if isinstance(op, RelAlg.Table):
+      return [o.elt_name.data for o in op.schema.ops]
+    if isinstance(op, RelAlg.Select):
+      return self.find_schema(op.input.op)
+    if isinstance(op, RelAlg.Project):
+      return [s.data for s in op.names.data]
+    if isinstance(op, RelAlg.Aggregate):
+      return [s.data for s in op.col_names.data]
+
 
 @dataclass
 class ProjectionInference(ProjectionOptimizer):
@@ -97,9 +110,12 @@ class IdentityProjectionRemover(ProjectionOptimizer):
     if not all([isinstance(o, RelAlg.Column) for o in op.projections.ops]):
       return
 
-    cols = self.flatten([self.find_cols_in_expr(o) for o in op.projections.ops])
+    input_schema = self.find_schema(op.input.op)
+    output_schema = self.flatten(
+        [self.find_cols_in_expr(o) for o in op.projections.ops])
 
-    if cols == [s.data for s in op.names.data]:
+    if output_schema == [s.data for s in op.names.data
+                        ] and input_schema == output_schema:
       new_op = op.input.blocks[0].detach_op(op.input.op)
       rewriter.replace_matched_op(new_op)
 
--- a/experimental/sql/src/projection_pushdown.py
+++ b/experimental/sql/src/projection_pushdown.py
@@ -13,23 +13,99 @@ from xdsl.pattern_rewriter import (RewritePattern, GreedyRewritePatternApplier,
 
 import dialects.rel_alg as RelAlg
 
+# This file contains several rewrites on the rel_alg level, all of which concern
+# projections.
+
 
 @dataclass
-class ProjectionPushdownRewriter(RewritePattern):
+class ProjectionOptimizer(RewritePattern):
+
+  def flatten(self, l: list[list[str]]) -> list[str]:
+    return [elem for sublist in l for elem in sublist]
 
   def find_cols_in_expr(self, op: RelAlg.Expression) -> list[str]:
     """
-    Returns the list of cols that are used in this operation or its children.
+    Returns the list of cols that are used in this expression or its children.
     """
-    if isinstance(op, RelAlg.BinOp):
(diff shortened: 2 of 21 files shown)


experimental/sql/src/projection_pushdown.py
```diff
@@ -0,0 +1,86 @@
...
+    ]
+
+    project_cols = [
+        elem for sublist in (self.find_cols_in_expr(o)
+                             for o in op.projections.ops) for elem in sublist
+    ]
+
+    result_cols = [s.data for s in op.names.data]
+
+    # TODO: handle renames of columns
```
ingomueller-net, line 52: In the spirit of my comment above, I think pushing a projection through a rename consists of (1) removing all renames where the new name does not occur in the projection, (2) apply the reverse rename to the projection (what happens if the rename is not injective, e.g., `a->b, a->c`?), and (3) swap the projection with the rename.

ingomueller-net (commented):
I have some comments about making the pass more general. If you don't want to do it in this PR, you can name things a bit differently and leave comments about how you plan to extend things...

webmiche:
Thanks a lot for your comments!

I think I am currently generally a bit confused how to do all of these things since this is basically the first time that I implement something like this and don't just do it conceptually. I will think about your feedback and evolve this further. 

One conceptual question is what we actually mean with projection pushdown. Do we mean the thing where we have a project and another operator and we swap their positions? Or do we need something more like "projection inference" where we basically infer a projection in front of any operator, that projects out all the columns that are not needed further up in the tree? Notice that the second thing is quite involved since something are not needed in an operator, but are actually part of the desired output so that needs to be kept in mind as well.

I feel like we need the second thing for Q6, so I will try to go more for that. However, I feel that the name "projection inference" is more descriptive than "projection pushdown". What do you think?

experimental/sql/src/projection_pushdown.py
```diff
@@ -0,0 +1,86 @@
...
+@dataclass
+class ProjectionPushdownRewriter(RewritePattern):
+
+  def find_cols_in_expr(self, op: RelAlg.Expression) -> list[str]:
+    """
+    Returns the list of cols that are used in this operation or its children.
+    """
+    if isinstance(op, RelAlg.BinOp):
+      return self.find_cols_in_expr(op.lhs.op) + self.find_cols_in_expr(
+          op.rhs.op)
```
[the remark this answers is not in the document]
webmiche: You are right. i initially implemented it like this since the different child regions are named differently, but I can just go over the `op.regions` field.

experimental/sql/src/projection_pushdown.py
```diff
@@ -0,0 +1,86 @@
...
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelAlg.Project, rewriter: PatternRewriter):
+    if not isinstance(op.input.op, RelAlg.Select):
+      return
+
+    select_cols = [
+        elem for sublist in (
+            self.find_cols_in_expr(o) for o in op.input.op.predicates.ops)
+        for elem in sublist
+    ]
```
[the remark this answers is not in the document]
webmiche: I think I want to split this into several steps with the idea of peephole rewrites in mind.

First, all kinds of operators (but projections) get a projection inserted just before them mapping away all columns that won't be needed later on.

Second, the exact projections, so the expressions that define the projections, are pushed down as far as possible.

Third, projections, that follow are followed by other projections, are fused into single projections.

Lastly, all projections that are just the identity are removed.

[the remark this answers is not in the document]
webmiche: I am not quite sure that this is the way we really want to have this in the end, and I am in particular uncertain about what to do with joins, but I think that this could be a nice first step in the right direction. And if these things really work the way I want them to I feel that this might give us a lot of what we need.

experimental/sql/src/projection_pushdown.py
```diff
@@ -0,0 +1,138 @@
...
+    elif isinstance(op, RelAlg.Aggregate):
+      rewriter.replace_matched_op(
+          RelAlg.Aggregate.get(Region.from_operation_list([new_proj]),
+                               [s.data for s in op.col_names.data],
+                               [s.data for s in op.functions.data],
+                               [s.data for s in op.res_names.data]))
+
+
+@dataclass
+class IdentityProjectionRemover(ProjectionOptimizer):
```
webmiche, line 104: If I didn't forget about a corner case, then this will certainly go into the thesis. I think this rewriter is a good example of something that is very easy on this abstraction level, but close to impossible on e.g. the iterators level.

webmiche:
While this current version certainly doesn't do all the smart things one could do, it is enough to optimize Q6 to the point that we want atm.

I suggest merging this (after you approved the code obvly) and reiterating on it, once we actually need to.

experimental/sql/test/projection_pushdown_tests/extended_case.xdsl
```diff
@@ -0,0 +1,70 @@
+// RUN: rel_opt.py -p projection-pushdown %s | filecheck %s
+
+module() {
```
webmiche, line 3: This query has basically the structure of Q6, showing that we can reach the optimization level we want atm.

Notice that the "d" of the input table gets projected out properly, as we want it to.

experimental/sql/src/projection_pushdown.py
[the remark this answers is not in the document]
ingomueller-net: I have the feeling that what you describe should work. It may not be the most efficient way to do things if you don't pay attention to the order in which patterns apply. (Projections only affect downstream operators, so you can do the rewrite in one pass from root to the leaves but patterns might be applied in the opposite direction, so you'd get quadratic runtime.) But it's definitely OK as a first step and probably fixable by controlling the application order...

ingomueller-net:
> One conceptual question is what we actually mean with projection pushdown. Do we mean the thing where we have a project and another operator and we swap their positions? Or do we need something more like "projection inference" where we basically infer a projection in front of any operator, that projects out all the columns that are not needed further up in the tree? Notice that the second thing is quite involved since something are not needed in an operator, but are actually part of the desired output so that needs to be kept in mind as well.
> 
> I feel like we need the second thing for Q6, so I will try to go more for that. However, I feel that the name "projection inference" is more descriptive than "projection pushdown". What do you think?

I think "projection push-down" generally refers to something like the second option. If you have access to the full text of the [Ullman book](http://infolab.stanford.edu/~ullman/dscb.html), you can have a look at Chapter 16.1.4. It says that (unlike for selections) "pushing" projections usually consists in keeping the original projection also where it was and adding some new projection before the upstream operator. I guess that adding a projection in front of any operator and cleaning up identities at the end as a somewhat over-eager version of that but essentially with the same effect.

webmiche:
Thanks a lot for the reference. I have read Chapter 16.2.4 and it seems to come to similar conclusions as I did. I agree that it feels like my peephole approach probably has the same effect. I don't quite understand your argument about quadratic time, since I was thinking of applying these patterns on each Operator exactly once. However, I guess to really get something optimal, we might want to apply patterns to the resulting Operators as well (in particular, projection fusing). I have to think about that a bit :)

ingomueller-net:
> I don't quite understand your argument about quadratic time, since I was thinking of applying these patterns on each Operator exactly once. However, I guess to really get something optimal, we might want to apply patterns to the resulting Operators as well (in particular, projection fusing). I have to think about that a bit :)

You are right: the way you have implemented things, you will only apply each pattern once and I *think* it will come up with the optimal result. *But* there is a quadratic runtime hidden in `find_cols_upstream`: it has a worst case runtime of O(n) (it traverses the whole tree when applied to the root op) and is applied to O(n) ops. This is because it's implemented recursively.

It may be possible to do it without recursion and materialize the required columns of each op somehow. But then, these materializations may change. Say you rewrite the IR from leaves to root. Every time you rewrite the IR, all ops from the rewritten op to the leaves may have their column sets change again, leading again to quadratic runtime. If you drive the rewriting as "push down", i.e., from root to leaves, that wouldn't happen.

Anyways, all of this is "just" about compilation performance and probably secondary at this point...

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+import dialects.rel_alg as RelAlg
+
+# This file contains several rewrites on the rel_alg level, all of which concern
+# projections. Currently, there are three rewriters. The first one introduces a
+# projection before any operator, the second one fuses subsequent projections,
+# and the third one removes projetions that are just identity mappings.
+
+
+@dataclass
+class ProjectionOptimizer(RewritePattern):
```
ingomueller-net, line 23: I think that all of the member functions here could be static. Maybe the should rather be top-level functions (outside of any class)?

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+  # tables) that map away all columns that aren't needed by this Operator or
+  # further up the tree.
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelAlg.Operator, rewriter: PatternRewriter):
+    if isinstance(op, RelAlg.Project) or isinstance(
+        op, RelAlg.Table) or isinstance(op, RelAlg.SchemaElement):
+      return
+    cols = list(dict.fromkeys(self.find_cols_upstream(op)))
+    if cols.__contains__(None):
```
ingomueller-net, line 82:
```suggestion
    if None in cols:
```

Why not simply this?

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+      return
+
+    new_proj = RelAlg.Project.get(
+        rewriter.move_region_contents_to_new_regions(op.input),
+        Region.from_operation_list([RelAlg.Column.get(s) for s in cols]),
+        ArrayAttr.from_list([StringAttr.from_str(s) for s in cols]))
+
+    # TODO: I mainly need this pattern match to call the proper builder. This
+    # could maybe be done nicer.
+    if isinstance(op, RelAlg.Select):
```
ingomueller-net, line 92: I wish that this wouldn't need a type switch here. Is there no generic way to get all operator inputs here and replace them with the new input?

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+
+@dataclass
+class ProjectionFuser(ProjectionOptimizer):
+
+  # This rewriter fuses two subsequent projections. Currently, it only works on
+  # renaming style projections, but it will be extended in the future.
+
+  def rename_expr(self, op: RelAlg.Expression, map: dict[str,
+                                                         str]) -> Operation:
+    if isinstance(op, RelAlg.Column):
```
ingomueller-net, line 113: Again, is there no way to do this without type switch? Can you recursively traverse the expression until you find the `Column`s in the leaves?

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+  def match_and_rewrite(self, op: RelAlg.Project, rewriter: PatternRewriter):
+    if not isinstance(op.input.op, RelAlg.Project):
+      return
+
+    child_cols = self.flatten(
+        [self.find_cols_in_expr(o) for o in op.input.op.projections.ops])
+    child_res = [s.data for s in op.input.op.names.data]
+    # If the child projection simply projects onto a subset, we can just infer
+    # its column mapping and apply it on the parents projection.
+    if len(child_cols) == len(child_res) and all(
```
ingomueller-net, line 137: The code rather seems to test for a rename but the comment talks about "subset". Is the comment out of date?

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # If any of the expressions in the projections list is not a Column
+    # expression, this is not an IdentityProjection.
+    # TODO: Maybe optimize simple expressions paterns (like add 0 or mult 1)
+    if not all([isinstance(o, RelAlg.Column) for o in op.projections.ops]):
+      return
+
+    input_schema = self.find_schema(op.input.op)
+    output_schema = self.flatten(
+        [self.find_cols_in_expr(o) for o in op.projections.ops])
```
ingomueller-net, line 163: It should be possible to inspect get the column names directly, since we know that all `projections.ops` are `Column` ops, no?

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+      return self.flatten([
+          self.find_cols_in_expr(o) for o in op.predicates.ops
+      ]) + self.find_cols_upstream(op.parent_op())
+    if isinstance(op, RelAlg.Project):
+      return self.flatten(
+          [self.find_cols_in_expr(o) for o in op.projections.ops])
+    if isinstance(op, RelAlg.Aggregate):
+      return [s.data for s in op.col_names.data]
+
+  def find_schema(self, op: RelAlg.Operator) -> list[str]:
```
ingomueller-net, line 55: I am wondering whether this isn't so fundamental that it could be a property/member function of `class Operator`.

webmiche:
Yes, I agree that this is probably the one to think about. However, I feel that one could make an argument that the recursion does not go up the tree too far, since it is actually only called on `Select` and not on any other operator. But I guess the worst-case scenario really is quadratic.

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+
+  def find_cols_upstream(self, op: RelAlg.Operator) -> list[str]:
+    """
+    Returns the list of cols that are used in this operation or its parents. If
+    the list contains None, a ModuleOp (so the output) was hit, showing that all
+    columns need to be kept.
+    """
+    if isinstance(op, ModuleOp):
+      return [None]
+    if isinstance(op, RelAlg.Select):
```
ingomueller-net, line 45: It may not be possible here but I'd like to be sure: can we remove the type-switch here and use a more generic traversal?

experimental/sql/test/projection_optimization_tests/rename_select.xdsl
```diff
@@ -0,0 +1,39 @@
+// RUN: rel_opt.py -p optimize-projections %s
+// This test currently fails, but shows a potential improvement.
```
ingomueller-net, line 2: Can you say why it fails?

experimental/sql/tools/rel_opt.py
```diff
@@ -32,6 +33,7 @@ def register_all_passes(self):
     self.available_passes['alg-to-ssa'] = alg_to_ssa
     self.available_passes['ssa-to-impl'] = ssa_to_impl
     self.available_passes['impl-to-iterators'] = impl_to_iterators
+    self.available_passes['optimize-projections'] = optimize_projections
```
ingomueller-net, line 36: I *think* this is effectively the textbook projection push-down. Shouldn't we call it that way?

ingomueller-net (approved):
I think that this is OK as a first stab. It's probably not as efficient as it could be due to the quadratic cost for inferring the referenced columns and its overly eager approach (adding stuff everywhere, then clean it up again) but I guess that that's secondary. I have several comments about type switches, which I think would be good you address where possible.

ingomueller-net:
> Yes, I agree that this is probably the one to think about. However, I feel that one could make an argument that the recursion does not go up the tree too far, since it is actually only called on `Select` and not on any other operator. But I guess the worst-case scenario really is quadratic.

Aha, good point, I didn't think about that. Yes, quite a few of the ops we currently have specify their list of columns explicitly. A few of the missing ones don't, though: joins, cross product, union all from the top of my mind.

experimental/sql/src/projection_optimizations.py
```diff
@@ -0,0 +1,191 @@
...
+
+@dataclass
+class IdentityProjectionRemover(ProjectionOptimizer):
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelAlg.Project, rewriter: PatternRewriter):
+
+    # If any of the expressions in the projections list is not a Column
+    # expression, this is not an IdentityProjection.
+    # TODO: Maybe optimize simple expressions paterns (like add 0 or mult 1)
```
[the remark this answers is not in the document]
webmiche: Yes indeed, but I wanted it written down somewhere and this was the moment I was thinking about it :)

experimental/sql/src/projection_optimizations.py
webmiche: I agree. I feel like there might be a step later on, where I add a `rel_alg_utils` file that contains these kinds of analysis functions.

experimental/sql/src/projection_optimizations.py
webmiche: I think it might be possible to write this with less cases, but I feel that would make a lot of assumptions on the general structure of rel_alg, so I would prefer to not do that.

Notice that the recursive call is only called in some cases, sometimes we need a call to `find_cols_in_expr`, some times we don't...

webmiche:
I have implemented to mapping from columns to expressions for projection fusion and I think I got rid of the pattern matches that can be removed in a sane way.

experimental/sql/src/projection_optimizations.py
ingomueller-net: I am not so much worried about the recursion than I am about splitting the knowledge about each individual op across many different places in the code base. With such a design, you need to touch many many places every time you add a new op and you can easily forgot some. (On the other hand, maybe this won't happen too often for relational algebra ;-) And of course, it is only possible if you can guarantee correctness based on common properties/assumptions...)

experimental/sql/src/projection_pushdown.py
```diff
@@ -13,15 +13,17 @@
...
 import dialects.rel_alg as RelAlg
 
-# This file contains several rewrites on the rel_alg level, all of which concern
-# projections. Currently, there are three rewriters. The first one introduces a
-# projection before any operator, the second one fuses subsequent projections,
-# and the third one removes projetions that are just identity mappings.
+# This file contains several rewrites that all compose to classic projection
+# pushdown. Currently, this works in three steps and, hence, has three
+# rewriters. The first one introduces a projection before any operator, the
+# second one fuses subsequent projections, and the third one removes projetions
```
ingomueller-net, line 19:
```suggestion
# second one fuses subsequent projections, and the third one removes projections
```

ingomueller-net (approved):
LGTM!

Nice generalization in the fusion logic of projections. Does this fix the previously broken test case?

webmiche:
No it does not since that test case needs to push a renaming projection trough a Select, something that I have not yet implemented. I would leave this open for a next version of the projection pushdown.

PR #558, opened 2022-08-19
SQL: fuse projections and tables

webmiche:
This PR adds the fusion of tables with subsequent (simple) projections. In order t do that, it introduces the `partial_table_scan` operation, an operation similar to the `full_table_scan`, but with col_names to define which columns should be loaded.

### DIFF
--- a/experimental/sql/dialects/rel_impl.py
+++ b/experimental/sql/dialects/rel_impl.py
@@ -6,7 +6,7 @@ from dataclasses import dataclass
 from typing import Any, Union, List, Optional
 from xdsl.ir import Block, Region, Operation, SSAValue, ParametrizedAttribute, Data, MLContext, Attribute
 from xdsl.dialects.builtin import StringAttr, ArrayAttr, ArrayOfConstraint, IntegerAttr
-from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder
+from xdsl.irdl import AttributeDef, OperandDef, ResultDef, RegionDef, SingleBlockRegionDef, irdl_attr_definition, irdl_op_definition, ParameterDef, AnyAttr, VarOperandDef, builder, OptAttributeDef
 
 # This file contains the relational implementation dialect, a dialect that
 # expresses relational queries using low-level concepts like tuples as block
@@ -432,22 +432,34 @@ class Operator(Operation):
 @irdl_op_definition
 class FullTableScanOp(Operator):
   """
-  Performs a full table scan of the table `table_name` and produces a bag with the given schema.
+  Performs a full table scan of the table `table_name` and produces a bag with
+  the given schema. If the optional attribute cols is set, only the given
+  columns are loaded. Otherwise, all columns are loaded.
 
   Example:
 
   '''
-  %0 : rel_impl.bag<[!rel_impl.int32]> = rel_impl.full_table_scan() ["table_name" = "t"]
+  %0 : rel_impl.bag<[!rel_impl.schema_element<"a", !rel_impl.int32>, !rel_impl.schema_element<"b", !rel_impl.int32>]> = rel_impl.full_table_scan() ["table_name" = "t"]
+  %0 : rel_impl.bag<[!rel_impl.schema_element<"a", !rel_impl.int32>]> = rel_impl.full_table_scan() ["table_name" = "t", cols = ["a"]]
   '''
   """
   name = "rel_impl.full_table_scan"
 
   table_name = AttributeDef(StringAttr)
+  cols = OptAttributeDef(ArrayAttr)
   result = ResultDef(Bag)
 
   @staticmethod
   @builder
-  def get(name: str, result_type: Bag) -> 'FullTableScanOp':
+  def get(name: str,
+          result_type: Bag,
+          cols: None | List[str] = None) -> 'FullTableScanOp':
+    if cols:
+      return FullTableScanOp.create(attributes={
+          "table_name": StringAttr.from_str(name),
+          "cols": ArrayAttr.from_list([StringAttr.from_str(s) for s in cols])
+      },
+                                    result_types=[result_type])
     return FullTableScanOp.create(
         attributes={"table_name": StringAttr.from_str(name)},
         result_types=[result_type])
--- a/experimental/sql/src/impl_to_iterators.py
+++ b/experimental/sql/src/impl_to_iterators.py
@@ -248,13 +248,27 @@ class BinOpRewriter(RelImplRewriter):
 @dataclass
 class FullTableScanRewriter(RelImplRewriter):
 
+  # This table_mapping is assumed to be populated already. In this version, this
+  # is done through the use of get_batch_and_name_list. The batch types are then
+  # used to build BlockArguments and this dictionary is populated by mapping
+  # strings to the BlockArguments built from the corresponding types. Notice
+  # that we assume that the name of partial loads are in the form
+  # "table_name","col1","col2"....
+  #
+  # In cases where only parts of a table is needed, the ColumnarBatch is assumed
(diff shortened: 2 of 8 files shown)


experimental/sql/tools/rel_opt.py
```diff
@@ -34,6 +35,7 @@ def register_all_passes(self):
     self.available_passes['ssa-to-impl'] = ssa_to_impl
     self.available_passes['impl-to-iterators'] = impl_to_iterators
     self.available_passes['projection-pushdown'] = projection_pushdown
+    self.available_passes['fuse-proj-table'] = fuse_proj_table
```
webmiche, line 38: Looking for a better name for this :)

experimental/sql/src/impl_to_iterators.py
```diff
@@ -258,6 +258,21 @@ def match_and_rewrite(self, op: RelImpl.FullTableScanOp,
...
+@dataclass
+class PartialTableScanRewriter(RelImplRewriter):
+
+  table_mapping: dict[str, it.ColumnarBatch]
+
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelImpl.PartialTableScanOp,
+                        rewriter: PatternRewriter):
+    rewriter.replace_matched_op(
+        it.ScanColumnarBatch.get(
```
ingomueller-net, line 270: I suppose that this assumes that the table mapping has been populated with column batch arguments that have the corresponding projection applied, right? Just to be sure: the logic that applies these projections isn't part of this commit, yet, right? Can you explain this mechanism in comments?

ingomueller-net: Oh, I see, `get_batch_and_name_list` applies the projections. Can you explain that here, where get_batch_and_name_list` does the projection, and in the documentation of that function?

experimental/sql/tools/rel_opt.py
ingomueller-net: `-fuse-proj-into-scan`?

ingomueller-net (approved):
Does this work end-to-end already? (If yes, shouldn't there be an end-to-end test?)

Otherwise, LGTM!

webmiche:
This works end-to-end and it is what I am currently using on my evaluation branch. I was not sure about an end-to-end test as we did not introduce any tests covering multiple layers so far, so I did not want to do this with this PR. However, I feel as tough it might make sense to introduce a full vertical test (with execution) at some point (maybe directly after this PR). For that, however, we will need things that are in #553 plus some more stuff on my evaluation branch, so I want to wait for that to be merged. What do you think?

experimental/sql/dialects/rel_impl.py
```diff
@@ -453,6 +453,33 @@ def get(name: str, result_type: Bag) -> 'FullTableScanOp':
         result_types=[result_type])
 
 
+@irdl_op_definition
+class PartialTableScanOp(Operator):
```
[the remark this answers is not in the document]
webmiche: I agree and I implemented the approach with an optional attribute.

experimental/sql/src/impl_to_iterators.py
webmiche: Actually, it's not really get_batch_and_name_list that does the projection since we want to project on to SSAValues (in this case BlockArguments matching the right table). And these block arguments can only be built later on.

So this function really only returns batchtypes and names and does not do any mapping.

I will add some comments to make this clearer.

webmiche:
Not sure why these CI jobs are failing. Maybe @ingomueller-net or @nicolasvasilache have an idea?

experimental/sql/src/impl_to_iterators.py
```diff
@@ -248,29 +248,27 @@ def match_and_rewrite(self, op: RelImpl.BinOp, rewriter: PatternRewriter):
...
+    if op.cols:
+      rewriter.replace_matched_op(
+          it.ScanColumnarBatch.get(
+              self.table_mapping[op.table_name.data + "," +
+                                 ",".join([s.data for s in op.cols.data])],
+              convert_bag(op.result.typ)))
+    else:
+      rewriter.replace_matched_op(
+          it.ScanColumnarBatch.get(self.table_mapping[op.table_name.data],
+                                   convert_bag(op.result.typ)))
```
ingomueller-net, line 271: Maybe you can factor out `table_name` (or similar), which is the only difference between the two cases.

experimental/sql/dialects/rel_impl.py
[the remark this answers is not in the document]
ingomueller-net: OK, cools, looks good!

experimental/sql/src/impl_to_iterators.py
ingomueller-net: Aha, right, the actual projection is done even earlier, completely outside of the plan. `get_batch_and_name_list` populates the column batches that are accessed by some scan op, some of which may be projected before the call into the plan. The new comments are better, but it doesn't mention yet that there is this convention between the call site and the plan. Can you elaborate on that?

ingomueller-net (commented):
Cool, thanks for the update! Only very minor comments, now. LGTM!

Aha, of course, the target dialect isn't ready in main yet 😛 Sure, let's add end-to-end tests once that is done.

ingomueller-net:
> Not sure why these CI jobs are failing. Maybe @ingomueller-net or @nicolasvasilache have an idea?

The tests failed with some connection error to Github. My best guess is that this is transient. I restarted the job -- let's see if this solves the problem...

experimental/sql/src/impl_to_iterators.py
webmiche: Are you referring to the projection of the columnar batch itself? I am not sure whether this is actually the right approach. We might have to think about changing your scan_columnar_batch at some point to also take col names and only builds tuples from these columns.

I changed the comment again :)

ingomueller-net: You haven't pushed that comment yet, have you?

webmiche: Yes, indeed. It's there now :)

experimental/sql/src/impl_to_iterators.py
```diff
@@ -258,6 +258,21 @@ def match_and_rewrite(self, op: RelImpl.FullTableScanOp,
                                  convert_bag(op.result.typ)))
 
 
+@dataclass
+class PartialTableScanRewriter(RelImplRewriter):
```
ingomueller-net: Only realizing this now: Is there any reason not to make "partial" scans an option of the full scan in this dialect as well?

experimental/sql/src/impl_to_iterators.py
```diff
@@ -338,6 +353,16 @@ def get_batch_and_name_list(
           ]))
       batches.append(curr_batch)
       names.append(o.table_name.data)
```
ingomueller-net: I think the documentation of this function (`Scans over all the operations ...`) should be updated to explain what happens with partial scans/projections.

experimental/sql/src/impl_to_iterators.py
ingomueller-net: Thanks for the update! I still don't understand where exactly the projection happens, though. As I understand it, the lowered code can use some SSA values with tables where the projection has already been applied (i.e., a columnar batch with only a subset of the original columns) and the table mapping contains the names of those SSA values. But who creates these values?

experimental/sql/src/impl_to_iterators.py
webmiche: This comment is out of date, right?

experimental/sql/src/impl_to_iterators.py
webmiche: This comment is out of date, right?

webmiche:
I feel that this can be merged, if the two comments I just addressed are actually out of date (@ingomueller-net ?).

ingomueller-net:
Yes, sorry, I must have looked at some earlier commit when I made these comments. Good to go!

PR #559, opened 2022-08-24
SQL: model joins

webmiche:
This PR adds a simple modeling for inner joins to all dialects. Currently, this works with a region similar to select and project. For the impl dialect, I decided to have two input tuples to that region. That's a design decision we might want to revisit.

Also, notice that I assumed that the names of columns are unique in joins. I feel that if that isn't the case, we might have to add a renaming projection when encountering the same names.


experimental/sql/test/ssa_to_impl/join.xdsl
```diff
@@ -0,0 +1,22 @@
...
+  }
+}
+
+//      CHECK:  %{{.*}} : !rel_impl.bag<[!rel_impl.schema_element<"a", !rel_impl.int64>, !rel_impl.schema_element<"b", !rel_impl.int64>]> = rel_impl.full_table_scan() ["table_name" = "t"]
+// CHECK-NEXT:  %{{.*}} : !rel_impl.bag<[!rel_impl.schema_element<"c", !rel_impl.int64>, !rel_impl.schema_element<"d", !rel_impl.int64>]> = rel_impl.full_table_scan() ["table_name" = "u"]
+// CHECK-NEXT:  %{{.*}} : !rel_impl.bag<[!rel_impl.schema_element<"a", !rel_impl.int64>, !rel_impl.schema_element<"b", !rel_impl.int64>, !rel_impl.schema_element<"c", !rel_impl.int64>, !rel_impl.schema_element<"d", !rel_impl.int64>]> = rel_impl.nested_loop_join(%{{.*}} : !rel_impl.bag<[!rel_impl.schema_element<"a", !rel_impl.int64>, !rel_impl.schema_element<"b", !rel_impl.int64>]>, %{{.*}} : !rel_impl.bag<[!rel_impl.schema_element<"c", !rel_impl.int64>, !rel_impl.schema_element<"d", !rel_impl.int64>]>) {
+// CHECK-NEXT:  ^{{.*}}(%{{.*}} : !rel_impl.tuple<[!rel_impl.schema_element<"a", !rel_impl.int64>, !rel_impl.schema_element<"b", !rel_impl.int64>]>, %{{.*}} : !rel_impl.tuple<[!rel_impl.schema_element<"c", !rel_impl.int64>, !rel_impl.schema_element<"d", !rel_impl.int64>]>):
```
ingomueller-net, line 17: At some point, we probably want to detect equi-joins and represent them with a dedicated op instead of a generic "theta" join. I guess an equi-join would have a list of column names for the left and right side, which need to be equal pairwise. Not that only equi-joins have O(n + m) or O(log n + log m) running time; theta joins need to build the Cartesian product and test each combination of records in O(n*m) time. Detecting them is thus required before we can decide to use a hash or sort join.

Have you thought about this? Where do you think is the best place to do that? One could carry the theta joins all to way to impl, detect them in that dialect, and thus only require non-theta joins from that point on. In relational algebra, one often has a natural join, which is an equi-join with assumptions on the names, plus a theta join, but strictly speaking, only the latter is required and the former is syntactic sugar.

ingomueller-net: On a related note, it would also be interesting to detect joins from Cartesian products + filters. That, however, I think is best done in the alg dialect.

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -275,6 +275,37 @@ def get(table: Region, metrics: Region, names: list[str]) -> 'Aggregation':
         })
 
 
+@irdl_op_definition
+class InnerJoin(Operation):
+  """
+  Models an ibis inner join of table `left` with table `right` and the condition
+  `predicates`. An empty region for `predicates` means that this operation is
+  just a cartesian product.
```
ingomueller-net, line 283:
```suggestion
  just a Cartesian product.
```

experimental/sql/dialects/rel_alg.py
```diff
@@ -296,6 +296,36 @@ def get(input: Region, predicates: Region) -> 'Select':
     return Select.create(regions=[input, predicates])
 
 
+@irdl_op_definition
+class InnerJoin(Operator):
+  """
+  Joins table `left` with table `right` under condition `predicates`. An empty
+  region for `predicates` means that this operation is just a cartesian product.
```
ingomueller-net, line 303:
```suggestion
  region for `predicates` means that this operation is just a Cartesian product.
```

experimental/sql/dialects/rel_impl.py
```diff
@@ -482,6 +482,45 @@ def get(input: Operation, predicates: Region) -> 'Select':
                          result_types=[input.results[0].typ])
 
 
+@irdl_op_definition
+class NestedLoopJoin(Operator):
+  """
+  Performs a nested loop join of `left` with `right` under condition
+  `predicates`. An empty region for `predicates` means that this operation is
+  just a cartesian product.
```
ingomueller-net, line 490:
```suggestion
  just a Cartesian product.
```

experimental/sql/test/ssa_to_impl/join.xdsl
ingomueller-net: And the opposite: an inner join without predicates (or even with non-equi predicates) could be canonicalized to a Cartesian product (plus filters).

experimental/sql/dialects/rel_ssa.py
```diff
@@ -494,6 +494,45 @@ def get(input: Operation, predicates: Region) -> 'Select':
                          result_types=[input.result.typ])
 
 
+@irdl_op_definition
+class InnerJoin(Operator):
+  """
+  Joins table `left` with table `right` under condition `predicates`. An empty
+  region for `predicates` means that this operation is just a cartesian product.
```
ingomueller-net, line 501:
```suggestion
  region for `predicates` means that this operation is just a Cartesian product.
```

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -275,6 +275,37 @@ def get(table: Region, metrics: Region, names: list[str]) -> 'Aggregation':
...
+  """
+  name = "ibis.inner_join"
+
+  left = SingleBlockRegionDef()
+  right = SingleBlockRegionDef()
+  predicates = SingleBlockRegionDef()
+
+  @staticmethod
+  @builder
+  def get(left: Region, right: Region, predicates: Region) -> 'InnerJoin':
```
ingomueller-net, line 305: If the names from left and right need to be distinct, wouldn't it be a good idea to assert that here?

experimental/sql/dialects/rel_alg.py
```diff
@@ -296,6 +296,36 @@ def get(input: Region, predicates: Region) -> 'Select':
...
+  """
+  name = "rel_alg.inner_join"
+
+  left = SingleBlockRegionDef()
+  right = SingleBlockRegionDef()
+  predicates = SingleBlockRegionDef()
+
+  @staticmethod
+  @builder
+  def get(left: Region, right: Region, predicates: Region) -> 'InnerJoin':
```
ingomueller-net, line 325: Dito: assert distinctness?

experimental/sql/dialects/rel_impl.py
```diff
@@ -482,6 +482,45 @@ def get(input: Operation, predicates: Region) -> 'Select':
...
+  left = OperandDef(Bag)
+  right = OperandDef(Bag)
+  predicates = SingleBlockRegionDef()
+
+  result = ResultDef(Bag)
+
+  @staticmethod
+  @builder
+  def get(left: Operation, right: Operation,
+          predicates: Region) -> 'NestedLoopJoin':
```
ingomueller-net, line 512: Dito: assert distinctness?

experimental/sql/dialects/rel_ssa.py
```diff
@@ -494,6 +494,45 @@ def get(input: Operation, predicates: Region) -> 'Select':
...
+
+  left = OperandDef(Bag)
+  right = OperandDef(Bag)
+  predicates = SingleBlockRegionDef()
+
+  result = ResultDef(Bag)
+
+  @staticmethod
+  @builder
+  def get(left: Operation, right: Operation, predicates: Region) -> 'InnerJoin':
```
ingomueller-net, line 522: Dito: assert distinctness?

experimental/sql/src/alg_to_ssa.py
```diff
@@ -57,9 +57,10 @@ def find_type_in_parent_operator(self, name: str,
     parent_op = op.parent_op()
     while (parent_op and not isinstance(parent_op, ModuleOp)):
       if isinstance(parent_op, RelSSA.Operator):
-        type_ = parent_op.operands[0].typ.lookup_type_in_schema(name)
-        if type_:
-          return type_
+        for o in parent_op.operands:
```
ingomueller-net, line 60: Update function documentation accordingly.

ingomueller-net: I am realizing that this function makes the assumption that the names it should look up (1) come from the inputs and (2) are distinct across the inputs. Is this always the case? In the natural join, it's not.

In other words, I think the function doesn't actually look up the names in the parent because which names exist in the parent may be operator-specific; it rather looks them up in the inputs of the parent.

I don't remember exactly what we need this function for so I don't have a good suggestion at the moment but I have the feeling that the current code isn't particularly elegant...

experimental/sql/src/ibis_frontend.py
```diff
@@ -103,6 +103,15 @@ def visit(  #type: ignore
   return id.UnboundTable.get(op.name, schema)
 
 
+@dispatch(ibis.expr.operations.relations.InnerJoin)
+def visit(op: ibis.expr.operations.relations.InnerJoin) -> Operation:
```
ingomueller-net, line 107: Dito: assert distinctness?

experimental/sql/src/ssa_to_impl.py
```diff
@@ -72,10 +72,15 @@ class ColumnRewriter(RelSSARewriter):
 
   @op_type_rewrite_pattern
   def match_and_rewrite(self, op: RelSSA.Column, rewriter: PatternRewriter):
-    # TODO: In this current version, there is only one block_arg, so it is clear
-    # what to access here. What if this is not the case?
+    # NOTE: This relies on the fact that all input columns have different names.
```
ingomueller-net, line 75: Assert that this is the case?

experimental/sql/src/ssa_to_impl.py
```diff
@@ -148,6 +153,26 @@ def match_and_rewrite(self, op: RelSSA.Project, rewriter: PatternRewriter):
...
+  @op_type_rewrite_pattern
+  def match_and_rewrite(self, op: RelSSA.InnerJoin, rewriter: PatternRewriter):
+    predicates = Region.from_block_list([
+        Block.from_arg_types([
+            self.create_tuple_of_bag(op.left.typ),
+            self.create_tuple_of_bag(op.right.typ)
+        ])
+    ])
+    # The following loop moves the operations of (the old) op.predicates to the
+    # predicates region of the new operation.
```
ingomueller-net, line 168: Isn't there a function in xdsl that does that? Can't there be more than one block?

ingomueller-net (commented):
LGTM! Most of my comments are minor or forward-looking. What may require some discussion is my comment on `find_type_in_parent_operator`. Let me know if you want to have a chat...

experimental/sql/dialects/ibis_dialect.py
webmiche: I am not sure we want to infer all the schemas here. This would require going over the entire tree. For the SSA-shaped dialect, this certainly makes sense tough!

experimental/sql/src/ssa_to_impl.py
webmiche: There isn't yet. There is only the option to move to a new region without block arguments. However, I feel that this is something we should add at some point.

webmiche: There cannot be more blocks in our current design. Clear this is an assumption on the IR, on which there is no verifier atm.

experimental/sql/dialects/ibis_dialect.py
ingomueller-net: Ah, I tend to forget that such information isn't available in the tree dialect. OK, do it only in the SSA-shaped one, then.

Am I looking at this too quickly or is this really a downside of the tree-shaped dialect?

ingomueller-net (approved):
LGTM!

experimental/sql/src/alg_to_ssa.py
webmiche: This function is mainly used to rewrite columns since we need to infer types here. The distinctness is something I am currently assuming. I have added several checks that this is the case, but I don't see how to nicely check this on this level atm. I feel our handling of columns will change at some point. Maybe removing the table form the column node was a bad idea after all...

Anywho, renamed and updated the docu.

experimental/sql/test/ssa_to_impl/join.xdsl
webmiche: At this point, I feel that it might even make sense to not have theta joins at all, but only equi-joins and Cartesian products. Maybe also natural joins, but AFAIU, these are basically inferred by their schemas and just a flavor of equi-joins (so maybe natural joins on rel_alg level lowered to equi-joins on this level?).

experimental/sql/dialects/ibis_dialect.py
webmiche: It's certainly a downside of being more declarative (leave out details -> details might not check out later on). In our current pipeline, we can argue that ibis does these checks, so we are "fine", but obviously we don't want to rely on that. 

Notice also that if we had at the very beginning decided to model trees with yields, we could find the types there now potentially. Something like:
```
%0 : bag = rel_alg.project() {
  %1 : bag = rel_alg.table()...
  rel_alg.yield(%1)
}
```

That's certainly an interesting insight for the thesis :)

ingomueller-net: Yep, indeed!

experimental/sql/test/ssa_to_impl/join.xdsl
webmiche: So, I guess there are two ways here: a) only model Cartesian products or b) model Cartesian products and equi-joins. As both require Cartesian products, I will implement that as a first version. After that, we can still decide whether we want equi-joins for this prototype. I must admit that I don't quite understand how naming is/should be handled in Cartesian products yet. I guess I can implement another version that assumes  distinct names, but this is clearly not an end solution...

One thing that I feel a bit uneasy about is that I feel like we are losing some semantics by deciding to model joins as Cartesian product + filter. In particular, semantics that were there before and might have to be rediscovered later on. That's clearly not what out multi-level IR approach wants to do, I think it is quite literally the opposite of what we actually want to do (preserve just enough semantics through the layers to apply optimizations).

Notice that I assume that left/right outer joins are basically just other "types" of Cartesian products, so in a finalized pipeline, these would have to be represented as well.

ingomueller-net: Yes, I agree, only having Cartesian product is quite limited for what we want to do. I didn't want to argue in favor of doing that. I do think that we should have equi-joins, i.e., go with option b).

For the name, the text book way is to say that Cartesian products of inputs with overlapping attribute names is undefined (or, as syntactic sugar, the attribute names are prefixed with the relation name on both sides, assume those are different). So I guess requiring distinctness is a sensible choice.

Outer joins are an interesting point! Note that "pure" relational algebra does not define NULL values, so it also does not have outer joins. Many extensions define them, though, but get into all problems you get into if you have NULLs. One possibility to model that are two Boolean attributes `outer_left` and `outer_right` (or similar), which is maybe more concise than having four different ops.

webmiche:
I rewrote this for Cartesian products. I guess equi-joins can be added in a lter PR.

experimental/sql/dialects/rel_impl.py
```diff
@@ -482,6 +482,39 @@ def get(input: Operation, predicates: Region) -> 'Select':
                          result_types=[input.results[0].typ])
 
 
+@irdl_op_definition
+class NestedLoopCartesianProduct(Operator):
```
webmiche, line 486: I am not happy with this name. i wanted to express, that this is a lower lveel than rel_ssa, but I feel that nested loops is probably the most sensible way to implement Cartesian products anyhow, so defining it here feels out of place. I think this could just (again) be call `cartesian_product`. What do you think, @ingomueller-net?

experimental/sql/dialects/rel_impl.py
```diff
@@ -483,38 +483,30 @@ def get(input: Operation, predicates: Region) -> 'Select':
 
 
 @irdl_op_definition
-class NestedLoopJoin(Operator):
+class NestedLoopCartesianProduct(Operator):
```
ingomueller-net, line 486: Why not call this simply `CartesianProduct` here as well? There is no other (typical) way to compute the Cartesian product than a nested loop, right? That distinction makes sense for the join because there are many other variants...

experimental/sql/dialects/rel_impl.py
```diff
@@ -483,38 +483,30 @@ def get(input: Operation, predicates: Region) -> 'Select':
...
   """
-  Performs a nested loop join of `left` with `right` under condition
-  `predicates`. An empty region for `predicates` means that this operation is
-  just a Cartesian product.
+  Computes a nested loop cartesian Product of `left` with `right`.
```
ingomueller-net, line 488:
```suggestion
  Computes the Cartesian product of `left` with `right`.
```

experimental/sql/src/ibis_frontend.py
```diff
@@ -105,11 +105,10 @@ def visit(  #type: ignore
 
 @dispatch(ibis.expr.operations.relations.InnerJoin)
 def visit(op: ibis.expr.operations.relations.InnerJoin) -> Operation:
-  return id.InnerJoin.get(
-      Region.from_operation_list([visit(op.left)]),
-      Region.from_operation_list([visit(op.right)]),
-      Region.from_operation_list(
-          [visit(op.predicates[0])] if op.predicates else []))
+  if op.predicates:
+    raise Exception(f"theta joins can not yet be lowered")
```
ingomueller-net, line 109: Can you not lower the `InnerJoin` to *two* ops, a `CartesianProduct` followed by a `Filter`?

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -276,34 +276,29 @@ def get(table: Region, metrics: Region, names: list[str]) -> 'Aggregation':
 
 
 @irdl_op_definition
-class InnerJoin(Operation):
+class CartesianProduct(Operation):
   """
-  Models an ibis inner join of table `left` with table `right` and the condition
-  `predicates`. An empty region for `predicates` means that this operation is
-  just a Cartesian product.
+  Models an ibis Cartesian product of tables `left` and `right`.
```
ingomueller-net, line 281: Maybe it makes sense to add a short definition à la ", i.e., the combination of all elements from `left` and `right`"?

ingomueller-net (approved):
Yep, I think that makes sense, thanks! LGTM!

PR #560, opened 2022-08-24
Add job scheduler

matthias-springer:
A simple scheduler that dispatches jobs to worker processes.


nicolasvasilache (approved):
Cool!

I imagine you want to use this with the NG search ?

One think that I haven't looked at in a long time is where we are on the state needs to be serialized and passed through the MP boundary.

Please make sure that this impl. works with where we were when we left off before committing too deeply.

Besides that, LGTM!

PR #563, opened 2022-09-16
Convert function signatures and calls in IteratorsToLLVM.

ingomueller-net:
This PR used to be one commit of #553 and should be non-controversial.

It enables type conversion on function signatures and function calls. Eventually, this is is required to pass in pandas DataFrames (in some to-be-defined type) into compiled code.

### DIFF
--- a/experimental/iterators/test/Integration/Dialect/Iterators/CPU/print.mlir
+++ b/experimental/iterators/test/Integration/Dialect/Iterators/CPU/print.mlir
@@ -2,9 +2,17 @@
 // RUN: mlir-cpu-runner -e main -entry-point-result=void \
 // RUN: | FileCheck %s
 
+func.func @print_empty_tuple(%tuple : tuple<>) -> () {
+  "iterators.printtuple"(%tuple) : (tuple<>) -> ()
+  return
+}
+
 func.func @main() {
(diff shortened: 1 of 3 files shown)


experimental/iterators/test/Integration/Dialect/Iterators/CPU/print.mlir
```diff
@@ -2,11 +2,19 @@
 // RUN: mlir-cpu-runner -e main -entry-point-result=void \
 // RUN: | FileCheck %s
 
+func.func @print_empty_tuple(%tuple : tuple<>) -> () {
+  "iterators.printtuple"(%tuple) : (tuple<>) -> ()
```
nicolasvasilache, line 6: note: printtuple reads quite poorly -> "print_tuple".
Also, print is one of those ops that is probably nicer to have as polymorphic.

ingomueller-net: Fair point. Changing that wouldn't be part of this PR, though, right?

nicolasvasilache: that's def a separate PR

PR #564, opened 2022-09-16
Iterators: [NFC?] Pass TypeConverter through IteratorAnalysis.

ingomueller-net:
This PR used to be part of #553 and should be non-controversial. It extends the `IteratorAnalysis` class and related functions with a `TypeConverter` member/argument such that that type converter can be used when computing the type of an iterator state (which is one of the main tasks of the analysis).

### DIFF
--- a/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
+++ b/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
@@ -4,6 +4,7 @@
 #include "iterators/Utils/NameAssigner.h"
 #include "mlir/Dialect/LLVMIR/LLVMTypes.h"
 #include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/Transforms/DialectConversion.h"
 #include "llvm/ADT/TypeSwitch.h"
 
 using namespace mlir;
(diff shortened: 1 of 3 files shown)


nicolasvasilache (changes requested):
I don't see much value in expanding all these APIs, seems like you want a more specialized type converter class that either "isa" typeConverter or "hasa" typeConverted.

ingomueller-net:
In the PR of the "read from memrefs" op (#553), I currently need a `TypeConverter` [here](https://github.com/iree-org/iree-llvm-sandbox/blob/b82cb1cdf125d4c411d6ebac6b64fd1d41ec5eb1/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp#L92) in order to determine the LLVM type of one field of the LLVM struct that represents the state of the iterator, namely that of the input "columnar batch"/"table"/...

I guess that the fact that I need this is a consequence of the fact that the lowering from iterators to LLVM is rather large step. But if I want to lower the iterator state to an LLVM struct, all of its contents have to be lowered to LLVM in the same step, so any state that would otherwise contain higher-level values, needs to do such a lowering. I understand that this is not elegant at all, but improving this fundamentally seems to require a redesign of the whole conversion. Is that something we want to do *before* merging in the "read from memrefs" op?

Let's suppose we don't and try to find a fix. Indeed, `IteratorAnalysis` now "has a" `TypeConverter`. But that doesn't have a formal function like `mlir::isa()`, right? And I would say that `IteratorAnalysis` "is *not* a" `TypeConverter` -- it computes types for the ops of a given module (in a way that is sensitive to their use-def chains) so it can't be used as a `TypeConverter`. Plus it does a bunch of other things (namely determining the names of the open/next/close functions). In short, I don't see how either of the two help?

Also note that we have to pass the `TypeConverter` to all `computeStateType` functions because they aren't member functions of the `IteratorAnalysis` (by design) -- unless we want to have have different calls to these functions depending on the function signatures (which I think isn't elegant and which I removed not too long ago in #554).

nicolasvasilache:
The redesign of the analysis, type vs SSA etc is a much longer topic.

Here I am just suggesting creating a new struct from the typeconverter and moving all the static funcs into it: low tech refactoring that avoids changing all the signatures everywhere.

ingomueller-net:
> Here I am just suggesting creating a new struct from the typeconverter and moving all the static funcs into it: low tech refactoring that avoids changing all the signatures everywhere.

Ah, OK, I see, thanks! Is the updated PR what you had in mind?

nicolasvasilache (approved):

PR #565, opened 2022-09-16
NFC - Use matchAndRewrite instead of match and rewrite.

ingomueller-net:
This PR used to be part of #553 and shouldn't be controversial. It uses `OpConversionPattern::matchAndRewrite` instead of separated `match` and `rewrite`, which I had previously overlooked.


nicolasvasilache:
Please clearly mark NFC PRs as "NFC - xxx"

nicolasvasilache (approved):

PR #569, opened 2022-09-19
[doc] Iterators: Create new top-level README with motivation and background.

ingomueller-net:
This PR creates a new top-level README file for the Iterators project. It currently provides a high-level motivation as well as some background on database-style iterators and an analogy of those to Python's iterators.

This is certainly not complete but maybe it is worth making a first reviewing pass to make sure it's going into the right direction.

PR #570, opened 2022-09-20
[doc] Iterators: Add defintions for currently-used iterator concepts.

ingomueller-net:
~~This PR depends on and currently contains #569.~~ I expect this PR to entail more discussion than the previous one since it defines some concepts with an intention of implementing them in some way and introduces terms that may not be generally accepted.

### DIFF
--- a/experimental/iterators/README.md
+++ b/experimental/iterators/README.md
@@ -273,3 +273,55 @@ the sequence-based relational model), neither the order of the rows nor that of
 the columns matter, (3) the elements of relations may be strings or similar
 variable-length data types, and (4) nested relational algebra also allows for
 structured element types (which may have variable length).
+
+## Basic Concepts
+
+This section defines the central concepts of this dialect and relates them to
+those introduced in the [Background](#background) section.
+
+* **Stream**: A collection of elements of a particular type that (1) is ordered
+  and (2) can only be iterated over in that order one element at the time.
+
+  Stream is the main data type that iterators consume and produce.
+
+  Ideally, there should be no restriction on the type of the elements and it
+  seems like it is possible to achieve that ideal (since Python achieves it).
+
+  We currently assume that streams are finite but most concepts will probably
+  remain unchanged if that restriction were lifted (see Python supporting
+  infinite iterators).
+
+* **Iterator Op**: An operation producing a stream and consuming zero or more
+  other streams (plus zero or more non-stream operands) identified by a name.
+
+  An iterator op only specifies *that* streams are handled in a particular way
+  but *how* that is happening is only specified in its documentation, not in the
+  form of code.
+
+  The equivalent to iterator ops in Python are functions returning iterators
+  (like `map` and similar built-in functions as well as those from `itertools`),
+  which are really "iterator factories" -- not iterators. We choose, say, `map`
+  by name, which produces an iterator that behaves according to the
+  documentation of `map`. We do *not* write down ourselves the imperative or
+  comprehension-based logic of the iterator, *nor* do we see or need to
+  understand the `iterator` interface.
+
+  Iterator ops can be lowered to an implementation based on the Volcano iterator
+  model but other lowerings are conceivable. (For example,
+  [this project](https://github.com/Dinistro/circt-stream) lowers ops that are
+  approximately equivalent to iterator ops to FPGAs.) Lowering materializes the
+  semantics specified by the name of the iterator op in a lower-level form;
+  similarly, the call to `map` materializes its semantics as an `iterator`
+  object that behaves according its specification.
+
+* **Iterator Body**: The function bodies of the functions `open`, `next`, and
+  `close`.
+
+  This is the equivalent of Python's `iterator`s, though with a different
+  protocol. It contains the implementation of the logic of the iterator exposed
+  through the open/next/close iterator interface.
+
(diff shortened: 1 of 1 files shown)


nicolasvasilache (approved):
LGTM!

experimental/iterators/README.md
```diff
@@ -273,3 +273,55 @@ the sequence-based relational model), neither the order of the rows nor that of
...
+This section defines the central concepts of this dialect and relates them to
+those introduced in the [Background](#background) section.
+
+* **Stream**: A collection of elements of a particular type that (1) is ordered
+  and (2) can only be iterated over in that order one element at the time.
+
+  Stream is the main data type that iterators consume and produce.
+
+  Ideally, there should be no restriction on the type of the elements and it
+  seems like it is possible to achieve that ideal (since Python achieves it).
```
webmiche, line 288: I am not sure I would claim that Python achieves this really. AFAIK, Python just tries to do it and if it works cool, otherwise fail. A aot compiled project like this might need stronger assumptions/regulations than that.

experimental/iterators/README.md
```diff
@@ -273,3 +273,55 @@ the sequence-based relational model), neither the order of the rows nor that of
...
+
+  We currently assume that streams are finite but most concepts will probably
+  remain unchanged if that restriction were lifted (see Python supporting
+  infinite iterators).
```
webmiche, line 292: Same here, are Python iterators really infinite or are they just not statically bound? I guess this comes down to a discussion about how to define the infiniteness of streams, which I don't feel we should have here. I think I am trying to say be careful around these kinds of comparisons :)

webmiche (commented):
Sorry for jumping in late, I don't think these comments need any changes, they are just pointing out some things to keep in mind :)

experimental/iterators/README.md
ingomueller-net: Well, they are not *really' infinite because you can't run a truly infinite amount of computations on current hardware :wink: But they are (claim to be?) infinite *conceptually*. The [`itertools`](https://docs.python.org/3/library/itertools.html) library has literally a section of "infinite iterators" such as `cycle`. I guess what "infinite" means here is "until you stop it" (but it won't stop by itself). I guess this definition would work for us, no?

Do you have any concrete failures in mind that could happen in Python? Or issues in AOT compilation?

One "issue" I see is that you can't do the same things with infinite streams as what you can do with finite ones. For example, you can't get the max of an infinite stream (in general) and expect the computation to finish. However, I'd argue that this is an issue of the ops, not of the type.

Finally, I have to say that, since I haven't actually implemented any of this, I wouldn't put my hand to cut for any of it either :wink: I hope that the word "probably" reflects this enough...

webmiche: Yes, I see I wasn't being very clear either what infinite means. As I said, we probably first have to define what an infinite stream really is and I actually don't know the standard definition or the most popular one or whatever.

What I tried to point out is just a feeling I have that aot compilation usually uses/needs stronger assumptions/guarantees than interpretation and that this might come back to bite us at some point. "until you stop it" is conceptually probably closer to interpretation ("just run until it finishes") compared to compilation ("analyse, translate to binary, then execute"). 

I think this sounds very interesting as something to think about. I will have lunch with Christian (the stream dialect guy) today, so I will try to keep in mind to bring this up. Maybe he has some interesting inputs considering that he has been working on related concepts for the last half a year :)

I have btw. discussed things like the max of infinite streams with him before and he basically redefined it to "the max until now" or, if you have some sort of windows, "the max over a window". I guess the latter sidesteps the issue in a sense by defining an infinte stream of elements as an infinite stream of finite streams of elements, which is an interesting idea in itself...

webmiche: Thinking a bit more about it, I feel that isn't really the fact that we have compilation vs interpretation but rather the kinds of type systems and language concepts these are usually mixed with. In python, everything is boxed types, so we essentially have a layer of abstraction and types dont really matter in some sense or are implicit. In compiled languages we usually have static, explicit types, so an int64 will be a 64bit thing and this has to be known statically (unless you want to implement pythons boxing with some array or pointer type). I feel that the former style mixes better with until you stop it than the latter :)

ingomueller-net: I agree with the need to define things clearly and I guess this is part of future work. However, I don't understand the interpretation vs compilation aspect: Any `while` loop runs until a condition is met without the compiler having to understand when or whether at all this happens.

I see the two definitions for the max of infinite streams. There is a name, BTW, for this "the max until now", which is the generalized prefix sum. It accumulates the whole stream seen until now using some accumulation function and returns the current accumulator after each new input. (On FPGAs and in other asynchronous contexts, "max until now" might have a different interpretation, though: "accumulate this input and tell me the max until now *when I ask you*" -- instead of "after each new input".)

webmiche: We will probably either encounter such a problem, if it is one, or we won't and then we know it is not a problem, so I guess we can leave this discussion at that.

My discussion with Christian resulted in that he thinks it should not be a problem as long as you have some sort of coroutines in your targeted language. So basically something that can be easily interrupted and resumed similar to how Python generators yield values (if I recall correctly) and then resume execution later on.

ingomueller-net: OK, cool, thanks for the update!

PR #574, opened 2022-09-21
Iterators: Create data type for columnar representation of tables.

ingomueller-net:
This PR factors out the data type from #553. It  depends on #573, which it currently includes.


ingomueller-net:
Some more thoughts on the topic: I think what I am trying to model is "the equivalent of 2-D memrefs where the row-major dimension can have heterogeneous types." I can easily see how the following ops from the `memref` dialect have direct equivalents on such "row-heterogeneous memrefs": `copy`, `load`, `alloc*`, `dealloc`, `dim`, `dma_*`, `*global`, `prefetch`, `realloc`, `reinterpret_cast` (with limits), `store`, `subview` (with limits). If we think of each "column" as one memref, then most of these ops could be "implemented" with one of the memref-ops per column plus maybe some glue.

I am also getting the impression that there is something similar to the "layout" of memrefs, which could distinguish row-major vs column-major (i.e., array of structs vs struct of arrays) and potentially a mixture of the two (i.e., "struct of arrays of (plain value or struct of values)"). In that case, the type wouldn't prescribe a columnar format as it does now but carry attributes that describe the layout, which I guess would make it even more similar to memrefs.

ingomueller-net:
Summary of an offline discussion:

* This type is like a `memref<?xtuple<T1, ..., Tn>>`.
* A strawman lowering could be `llvm.ptr<llvm.struct<(T1, ..., Tn)>>`. This is the textbook row-major layout.
* Another lowering could be `llvm.struct<(llvm.ptr<T1>, ..., llvm.ptr<Tn>)>`, which is the textbook column-major layout (and can be thought of as the "transposition" of the above.
* The type is independent of iterators, so it should be moved into its own dialect.

ingomueller-net:
Closing in favor of #577.

PR #575, opened 2022-09-27
Iterators: Add test case implementing TPC-H Q6.

ingomueller-net:
This PR adds a test case implementing TPC-H Query 6 run on a tiny data set provided as constant data.

### DIFF
--- a/experimental/iterators/test/Integration/Dialect/Iterators/CPU/tpch-q6.mlir
+++ b/experimental/iterators/test/Integration/Dialect/Iterators/CPU/tpch-q6.mlir
@@ -1,0 +1,131 @@
+// RUN: mlir-proto-opt %s \
+// RUN:   -convert-iterators-to-llvm \
+// RUN:   -convert-func-to-llvm \
+// RUN:   -convert-scf-to-cf -convert-cf-to-llvm \
+// RUN: | mlir-cpu-runner -e main -entry-point-result=void \
+// RUN: | FileCheck %s
+
+// Original query in SQL:
+//
+//   SELECT
+//           SUM(l_extendedprice * l_discount) AS revenue
+//   FROM
+//           lineitem
+//   WHERE
+//           l_shipdate >= date '1994-01-01'
+//           AND l_shipdate < date '1995-01-01'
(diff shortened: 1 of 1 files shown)


experimental/iterators/test/Integration/Dialect/Iterators/CPU/tpch-q6.mlir
```diff
@@ -0,0 +1,131 @@
...
+// Attributes used in the query and their physical representations:
+//
+//   - l_extendedprice: i32 (in cents)
+//   - l_discount: i8 (in 0.01 percent)
+//   - l_shipdate: i16 (in number of days since epoch)
+//   - l_quantity: i8
+//
+// Query constants in physical representation:
+//
+//    - date '1994-01-01' (l_shipdate): 8766 : i16
```
webmiche, line 29: This is just days since, right? Notice that I (and you in your integer version of the TPC-H benchmark) used seconds since making this constant 757382400. Just something to keep in mind :)

webmiche (approved):
LGTM! I see that you have generated short-circuiting code in the filter. This might actually be a potential speedup for me as well. I will think about how/whether to implement this :)

experimental/iterators/test/Integration/Dialect/Iterators/CPU/tpch-q6.mlir
ingomueller-net: Yes, I am representing this as days. Depending on the schema, you may know this, namely if it is of type `DATE` and not `DATETIME`. You may even know the minimum value and use that as "first day" instead of `1970-01-01` and rewrite query constants accordingly, which may allow you to use even fewer bits.

nicolasvasilache (approved):
Generally LGTM but why are we still using the generic op form ("iterators.constantstream") on all iterator ops?

ingomueller-net:
> [...] why are we still using the generic op form ("iterators.constantstream") on all iterator ops?

We initially didn't in order to spare ourselves the work to implement printers and parsers and never revised that decision. Should I start changing that?

PR #576, opened 2022-09-29
SQL: add group and order by

webmiche:
This PR adds the IR constructs for group by and order by. Notice that order by is implemented in ibis as part of a selection. On the other hand, group by gives a so-called GroupedTable. This is actually its own class and not a standard TableExpr like I would have expected. I am not quite sure why that is the case, but I feel that this can basically be lowered to a `group_by` operation in rel_alg.


experimental/sql/dialects/rel_alg.py
```diff
@@ -112,6 +112,23 @@ class Nullable(DataType):
   type: ParameterDef[DataType]
 
 
+@irdl_attr_definition
+class Order(ParametrizedAttribute):
```
ingomueller-net, line 116: The result "type" of `Order` is a big different from the other operators in that it returns a sequence, not a bag. There *are* probably ways to not represent that with a type but we have to think a bit about the implications and semantics of having this op.

experimental/sql/dialects/rel_alg.py
```diff
@@ -224,6 +241,64 @@ class Operator(Operation):
   ...
 
 
+@irdl_op_definition
+class GroupBy(Operator):
```
ingomueller-net, line 245: In relational algebra, you generally don't have a "grouped table"; instead, "grouping" and "aggregation" are represented as a single operation "grouping with aggregation". (There isn't anything else that you could to with it, anyways, at least not in SQL.) So I'd expect the aggregation functions to show up here...

experimental/sql/src/ibis_frontend.py
```diff
@@ -125,7 +126,23 @@ def visit(  #type: ignore
...
+  return id.Selection.get(table, predicates, projections, sort_keys, names)
+
+
+@dispatch(ibis.expr.operations.sortkeys.SortKey)
+def visit(op: ibis.expr.operations.sortkeys.SortKey) -> Operation:
+  return id.SortKey.get(Region.from_operation_list([visit(op.expr)]),
+                        op.ascending)
+
+
+import ibis.expr.types.groupby
```
ingomueller-net, line 139: Shouldn't this be at the top of the file?

experimental/sql/src/ibis_frontend.py
```diff
@@ -125,7 +126,23 @@ def visit(  #type: ignore
...
+
+
+@dispatch(ibis.expr.types.groupby.GroupedTable)
+def visit(op: ibis.expr.types.groupby.GroupedTable) -> Operation:
+  return id.GroupedTable.get(Region.from_operation_list([visit(op.table)]),
+                             visit_ibis_expr_list(op.by))
```
ingomueller-net, line 145: I guess you haven't modelled/you aren't capturing aggregations, yet, at all, right? I think the two have to go together for a relational-algebra-based dialect.

experimental/sql/dialects/rel_alg.py
ingomueller-net: BTW, there seems to be canonical form (?) or optimizations that you can do: You don't loose generality if you require:

1. Each column to occur either as a grouping key or as an input to some aggregation function. If it doesn't occur as either, it is effectively projected away.
2. Each grouping key only occurs once. If it occurs several times, it doesn't modify the definition of the groups and can be "brought back" with a subsequent projection.
3. exploit the idempotence of some aggregation functions in case they are used on a column that is also used as a grouping key. Consider `SELECT MAX(a), ... GROUP BY a ...`. You know that `MAX(a)` is equal to `a` because all values in the `a` group (or *one of the* `a` groups, if there are additional grouping keys) have the value `a` and the `MAX` of those is again `a`. You can thus calculate the "`MAX(a)`" with a projection after the grouping (which, itself) doesn't compute it).

ingomueller-net (commented):
This looks great! I have a few comments on the design though.

I guess one of them basically answers your question from the description of the PR: ibis' group by only does half of the job of relational "grouping with aggregation". They do this, I guess, to (1) allow auto-completion of `aggregate` and similar after having called `by` and (2) restrict what you can do on such a grouped table. I am pretty sure that you can't join it, for example -- you need to aggregate first.

The other (implicit?) question you had is about the ordering being part of the `SELECT`. My guess is this: the `SELECT` somehow model "what you see" as the result of computations and the order is part of that. For the other operations, namely `JOIN` and all variants, `WHERE`, `HAVING`, `GROUP BY`, the order doesn't matter.

I also raised the question of how to model "ordered bags". Maybe ibis' design is their answer to this question.

experimental/sql/dialects/rel_alg.py
webmiche: I feel that this is not important atm, since we do not have types in this dialect. But I agree that I have to think of something for the ssa-based dialects.

experimental/sql/src/ibis_frontend.py
webmiche: We actually have aggregations in the rel_alg dialect, it looks the following:
'''
rel_alg.aggregate() ["col_names = ["b"], "functions" = ["sum"]] {
    rel_alg.pandas_table() ...
  }
'''

Is it really the case that the two have to always go together? Isn't this a valid query?
'''SQL
SELECT * FROM lineitem

ORDER BY l_extendedprice
'''

Or would you describe this as jsut an "identity aggregation" or something like that?

webmiche: We actually have aggregations in the rel_alg dialect, it looks the following:
```
rel_alg.aggregate() ["col_names = ["b"], "functions" = ["sum"]] {
    rel_alg.pandas_table() ...
  }
```

webmiche:
I have implemented a version that fuses group bys and aggregations now. It actually works without any `GroupedTable`s, as ibis actually fuses them as well, if you used together. AFAIU, only things like the below example are represented as GroupedTables in ibis.

``` SQL
SELECT a, b FROM T
GROUP BY a, b
```

experimental/sql/src/ibis_frontend.py
ingomueller-net: Aha, right, we support it for *ungrouped* aggregation! So my formulation was a bit sloppy: the two don't need to go together (because aggregation does not require grouping) but grouping *does* require aggregation.

ingomueller-net:
> I have implemented a version that fuses group bys and aggregations now. It actually works without any `GroupedTable`s, as ibis actually fuses them as well, if you used together. AFAIU, only things like the below example are represented as GroupedTables in ibis.
> 
> ```sql
> SELECT a, b FROM T
> GROUP BY a, b
> ```

I would guess that this is an artifact of the interface, probably due to the fact that when you have written `group_by(...)`, you may or may not add `aggregate(...)` after that -- if you don't, you may still want the query to execute. Note that this only results in a valid SQL query if only grouping attributes (from the `GROUP BY` clause) are used as projections in the `SELECT` clause. (A few dialects also allow other dialects and use the implicit aggregation `ANY` for the other attributes.)

Also note that this corresponds to an "empty" aggregation, so you can probably translate such a case to a fused group-by with an empty list of aggregations.

experimental/sql/dialects/ibis_dialect.py
```diff
@@ -318,17 +291,18 @@ class Aggregation(Operation):
   table = SingleBlockRegionDef()
```
ingomueller-net, line 291: Please update the docu. Make sure you discuss the semantics of what happens if `by` and `metrics` is empty, respectively, vs when it is not.

experimental/sql/test/frontend/group_by.ibis
```diff
@@ -0,0 +1,37 @@
+# RUN: rel_opt.py -f ibis %s | filecheck %s
+
+table = ibis.table([("a", "string"), ("b", "int64"), ("c", "int64")], 't')
+table.group_by([table.a, table.b]).aggregate(table.c.sum().name('c_sum'))
```
ingomueller-net, line 4: Does the "empty" aggregation (i.e., group-by without aggregation) work? Can you add a test case for it?

ingomueller-net (approved):
LGTM! I think that this works. One comment about the documentation...

In lower dialects, we may need to think about whether we still want a single op for scalar aggregation and grouping with aggregation but it makes perfect sense to use only one in ibis, and it is probably good to have just one op in rel alg, too.

experimental/sql/dialects/rel_alg.py
webmiche: Yes, you are right. This is (probably) another downside of this style of tree-shaped dialect and representing information as implicitly as we are atm. In the current design, it might just end up being another tree traversal, trying to figure out whether the stream is order at a given operation.

I have however defined an optimized compilation stack and there we can have this information on the tree-level. Also, this actually gives me a great example for why this representing types on the tree level is useful, thanks a lot! :)

(Spoiler: in the new design, the tree comes in a type-agnostic and a type-aware version, so SSAifying and type inference are split up. This means that, e.g., projection pushwodn can be down type-agnosticly, whereas optimizations that rely on the order can be done on the type-aware tree dialect).

experimental/sql/test/frontend/group_by.ibis
webmiche: I have removed the GroupedTable handling from the prototype since TPC-H queries do not need this (AFAIK). Hence, this will not work anymore.

I can add it again, but I feel that we will throw away (most of) this prototype after the thesis submission and go with the optimized design instead.

webmiche:
How is an aggregation with group by implemented usually? They basically have to be done in the same loop-nest, right? I think of something like maybe a hash table, where the grouped columns are the key, then you aggregate in the bins directly, when collisions occur. 

I feel that it makes sense to implemented them as one op for the moment. It can probably still be split, if we think that would be beneficial. What do you think, @ingomueller-net?

experimental/sql/test/frontend/group_by.ibis
webmiche: And even if we want this behavior, we could still express it with the current version, by using `.aggregate()`, right?

ingomueller-net:
> How is an aggregation with group by implemented usually? They basically have to be done in the same loop-nest, right? I think of something like maybe a hash table, where the grouped columns are the key, then you aggregate in the bins directly, when collisions occur.

Exactly, this is one of the two textbook algorithms. Doing the aggregation in-line has the benefit of keeping the size of intermediate results small. The other textbook algorithm is to sort the data by the grouping columns and then aggregate in one pass, emitting the output of one group every time a new grouping key is detected. Practical algorithms typically do early aggregation in order no reduce the intermediate result sizes as well. In that case, as well as in the hash-table based approach, the two have to be seen as one operation. Since grouping alone cannot occur in SQL, there is also no reason to have an op just for that.

experimental/sql/test/frontend/group_by.ibis
ingomueller-net: Yes, true. We can produce an empty aggregation. We will fail to compile some ibis constructs that are valid for other backends, though, no? But I guess that's OK; just document it.

webmiche:
Wouldn't the second algorithm correspond to an order by and an ungrouped aggregation? This could be implemented as two passes, right? But I guess you are losing some performance if you actually materialize the entire sorted table just to aggregate it afterward...

ingomueller-net:
> Wouldn't the second algorithm correspond to an order by and an ungrouped aggregation? This could be implemented as two passes, right? But I guess you are losing some performance if you actually materialize the entire sorted table just to aggregate it afterward...

It's almost that but not quite: with an ungrouped aggregation, you emit only one result; with the algorithm I described, you emit one every time the grouping key changes. But yes, without early aggregation, materializing the sorted input is most often quite wasteful. (The only times that it may actually be a good idea is when either (1) the result is (partially) pre-sorted or (2) the number of groups is almost or exactly as large as the number of inputs, in which case the grouped result is essentially as big as the sorted result.)

webmiche:
Ah yes, the semantics of the ungrouped aggregation are different, you are right!

Then again, say we had sorted data (from an index scan or w/e), we want to group it, and the sorting and grouping columns are the same. In this case, we want to just aggregate in some sense, so essentially do a grouped aggregation, that doesn't group (in restructuring the data), but does aggregate (in combine rows). I guess this could be either its own operation or this could be some context-sensitive lowering. In this case it might make sense to have split operations for the latter algorithm. I guess this is something for later though :)

ingomueller-net:
Yes, good observation! This is know as "interesting orderings". It's also true for other combinations: if you have a join and a grouping later on the same key, it may be worth to sort by the join key, do the join as a sort-merge join, and then use that ordering for the aggregation. (Similar, if the second operation is another join, or a an order by.) I think that these are physical properties that we probably leave for a different dialect...

webmiche:
These feel like they might be different types to me, but I am not sure we can define a sensible type hierarchy that allows subtyping (like sorted by [a,b] is also sorted by [a]) while also enforcing the necessary invariants...

I guess this also relates to the decision of how/whether to model the output of an order as a bag type or a sequence. Essentially, a sequence is also "just" an interesting ordering, I would argue.

I will implement a version that keeps it as a bag as I would consider a all sequences to also be bags. We can then iterate in that PR on whether that makes sense or not. Also considering how much this adds for the thesis (which I feel is certainly not nothing since the possibility of representing interesting orderings as type hierarchies has not been described in it yet and this would even give us a first example of something like this idea).

PR #577, opened 2022-09-29
Iterators: Add Tabular dialect with TabularView type and initial set of ops.

ingomueller-net:


experimental/iterators/lib/Dialects/Tabular/IR/Tabular.cpp
```diff
@@ -0,0 +1,124 @@
...
+                            [&](Type type) { type.print(stream); });
+    }
+    std::string memrefColumnTypes;
+    {
+      llvm::raw_string_ostream stream(memrefColumnTypes);
+      llvm::interleaveComma(memrefs().getTypes(), stream, [&](Type type) {
+        type.cast<MemRefType>().getElementType().print(stream);
+      });
+    }
+    return emitOpError()
```
nicolasvasilache, line 73: I would just iterate with an index and report that `element type of operand "index" (i.e. memref<...>) does not match the result "index" column type (i.e. resulttype<...>))`.
The message is both more actionnable and the code quite simpler as you just use vanilla type printing and avoid managing your strings.

experimental/iterators/lib/Dialects/Tabular/IR/Tabular.cpp
```diff
@@ -0,0 +1,124 @@
...
+           << "type mismatch: should return a tabular view with the column "
+           << "types of the elements of the input memrefs (namely: "
+           << "tabular_view<" << memrefColumnTypes << ">) "
+           << "but returns a tabular_view<" << viewColumnTypes << ">.";
+  }
+
+  // Verify memrefs have rank 1 and contiguous memory layout.
+  {
+    for (const auto &indexedMemref : llvm::enumerate(memrefs().getTypes())) {
```
nicolasvasilache, line 82: You can use StridedMemRefRankOf mlir/include/mlir/IR/OpBase.td and trim most of this out.
Feel free to add extra support for contiguous and what else may be missing for your use case and send it for review upstream.

experimental/iterators/lib/Dialects/Tabular/IR/Tabular.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+      // Only support contiguous memory layout.
+      if (!isStaticShapeAndContiguousRowMajor(memrefType))
+        return emitOpError()
+               << "unsupported type: input memref #" << indexedMemref.index()
+               << " does not have a static shape and a contiguous memory "
+                  "layout, both of which are required.";
+    }
+
+    // Verify all memrefs are of equal static length.
```
nicolasvasilache, line 98: please check OpBase.td for support for similar things (not sure it exists but at least let's check)

nicolasvasilache (approved):

experimental/iterators/lib/Dialects/Tabular/IR/Tabular.cpp
ingomueller-net: OK, I added some tablegen constructs to `TabularOps.td` in the commit "Move some type constraints to tablegen code." What do you think? Is this worth/ready for upstreaming?

ingomueller-net:
Closing in favor of #590, #592, #593, #594, and #595.

PR #579, opened 2022-10-03
Iterators: Implement TabularViewToStreamOp.

ingomueller-net:
This PR depends on and therefor currently includes #600. The PR implements an iterator op that allows iterating over the "tabular view" introduced in the previous PR, which now lives in its own `Tabular` dialect.

The lowering of this op currently has a dependency on the lowering of the tabular view: the iterator needs the LLVM representation of the view in order to be able to store it in its state, so the view needs to be lowered at the same time as the iterator op. I have thought of overcoming this eventually by breaking up the lowering of iterators into a lowering into built-in `tuple`s (which can store any type, including `TabularViewType`) such that the tabular dialect can be lowered in a separate pass (and the `tuple`s would be lowered to LLVM structs again later). However, this would be a complete overhaul of the current pipeline, which I am not sure I should do right away.

### DIFF
--- a/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
+++ b/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
@@ -267,11 +267,12 @@ def Iterators_ReduceOp : Iterators_Op<"reduce",
 }
 
 def Iterators_TabularViewToStreamOp : Iterators_Op<"tabular_view_to_stream",
-    [AllMatch<["$input.getType().cast<TabularViewType>().getColumnTypes()",
-               "$result.getType().cast<StreamType>().getElementType()"
-               "    .cast<LLVM::LLVMStructType>().getBody()"],
-              "the field types of the LLVM struct in the result stream must "
-              "match the column types of the tabular view given as input">]> {
+    [TypesMatchWith<"element type of input stream must match result type",
+                    "result", "input",
+                    "TabularViewType::get("
+                    "    $_self.getContext(),"
+                    "    $_self.cast<StreamType>().getElementType()"
+                    "          .cast<LLVM::LLVMStructType>().getBody())">]> {
   let summary = "Extracts the tuples from a tabular view as LLVM structs";
   let description = [{
     Produces a stream of LLVM structs from a given `tabular_view` (i.e., "scans"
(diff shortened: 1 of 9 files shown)


experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -265,6 +266,29 @@ def Iterators_ReduceOp : Iterators_Op<"reduce",
...
+  let description = [{
+    Produces a stream of LLVM structs from a given `tabular_view` (i.e., "scans"
+    the given `tabular_view`). Each LLVM struct represents one row, and the op
+    produces all rows in ascending order.
+
+    Example:
+    ```mlir
+    %stream = "iterators.tabular_view_to_stream"(%view)
+      : (!iterators.tabular_view<!t1, ..., !tn>)
+        -> !iterators.stream<!llvm.struct<(!t1, ..., !tn)>>
```
nicolasvasilache, line 285: The dependency on LLVM here is very unfortunate.

experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
```diff
@@ -99,6 +99,24 @@ LLVMStructType StateTypeComputer::operator()(
...
+/// template <typename TabularViewType>
+/// struct { int64_t currentIndex; TabularViewType view; }
+template <>
+LLVMStructType StateTypeComputer::operator()(
+    TabularViewToStreamOp op,
+    llvm::SmallVector<LLVM::LLVMStructType> /*upstreamStateTypes*/) {
+  MLIRContext *context = op->getContext();
+  Type i64 = IntegerType::get(context, /*width=*/64);
+  Type viewType = typeConverter.convertType(op.input().getType());
+  return LLVM::LLVMStructType::getNewIdentified(
```
nicolasvasilache, line 115: Same here re LLVM dependence.

What is tricky here is that you already seem to depend on the LLVM dialect for this.

We can keep this as a later TODO but please be sure to cleanup before embarking on more abstractions.

nicolasvasilache (approved):

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -265,6 +266,29 @@ def Iterators_ReduceOp : Iterators_Op<"reduce",
...
+              "match the column types of the tabular view given as input">]> {
+  let summary = "Extracts the tuples from a tabular view as LLVM structs";
+  let description = [{
+    Produces a stream of LLVM structs from a given `tabular_view` (i.e., "scans"
+    the given `tabular_view`). Each LLVM struct represents one row, and the op
+    produces all rows in ascending order.
+
+    Example:
+    ```mlir
+    %stream = "iterators.tabular_view_to_stream"(%view)
```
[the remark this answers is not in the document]
ingomueller-net: Done in latest commit.

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
ingomueller-net: Yes! Added TODO in latest commit.

PR #583, opened 2022-10-04
SQL: add min, mean, max, count, and count_distinct reductions

webmiche:
This PR adds the min, mean, max, count, and count_distinct reductions. For the count, I decided to have an empty string as the col_name since we can take any column of the input operator to count, as all have the same number of rows. I also decided to rename `mean` to `avg` in rel_alg, as this corresponds to SQLs keyword (AFAIK).

This PR is the second in a row of PR's trying to make 9 tpc-h queries run with our prototype. This will let us represent 8 queries with one easy operation to go to handle the 9 simple queries. And with 13 operations to go in order to represent all of tpc-h.

### DIFF
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -490,6 +490,134 @@ class Sum(Operation):
 
 
 @irdl_op_definition
+class Min(Operation):
+  """
+  Takes the minimum of all the elements of the column given in arg based on the encompassing
+  aggregation operator.
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/operations/reductions.py#L204
+
+  Example:
+
+  '''
+  ibis.min() {
+    ibis.table_column() ...
+  }
+  '''
+  """
+  name = "ibis.min"
+
+  arg = SingleBlockRegionDef()
+
+  @staticmethod
+  @builder
+  def get(arg: Region) -> 'Min':
+    return Min.create(regions=[arg])
+
+
+@irdl_op_definition
+class Max(Operation):
+  """
+  Takes the maximum of all the elements of the column given in arg based on the encompassing
+  aggregation operator.
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/operations/reductions.py#L197
+
+  Example:
+
+  '''
+  ibis.max() {
+    ibis.table_column() ...
+  }
+  '''
+  """
+  name = "ibis.max"
+
+  arg = SingleBlockRegionDef()
+
+  @staticmethod
+  @builder
+  def get(arg: Region) -> 'Max':
+    return Max.create(regions=[arg])
+
+
+@irdl_op_definition
+class Mean(Operation):
+  """
+  Takes the mean of all the elements of the column given in arg based on the encompassing
+  aggregation operator.
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/operations/reductions.py#L108
+
+  Example:
+
+  '''
+  ibis.mean() {
+    ibis.table_column() ...
+  }
+  '''
+  """
+  name = "ibis.mean"
+
+  arg = SingleBlockRegionDef()
+
+  @staticmethod
+  @builder
+  def get(arg: Region) -> 'Mean':
+    return Mean.create(regions=[arg])
+
+
+@irdl_op_definition
+class Count(Operation):
+  """
+  Counts the elements in arg based on the encompassing aggregation operator.
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/operations/reductions.py#L18
+
+  Example:
+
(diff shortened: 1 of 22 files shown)


experimental/sql/dialects/ibis_dialect.py
```diff
@@ -489,6 +489,134 @@ def get(arg: Region) -> 'Sum':
...
+  arg = SingleBlockRegionDef()
+
+  @staticmethod
+  @builder
+  def get(arg: Region) -> 'Mean':
+    return Mean.create(regions=[arg])
+
+
+@irdl_op_definition
+class Count(Operation):
```
ingomueller-net, line 571: `COUNT` (with or without `DISTINCT`) in SQL is a bit weird: `COUNT(*)` is the special case that counts the number of rows; `COUNT(x)` counts the number of *non-NULL values* of `a` (see [Postgres doc](https://www.postgresql.org/docs/8.2/functions-aggregate.html)). Maybe we should model the column name after all, no?

ingomueller-net (commented):
Nice! I think the column name is required in `COUNT`, though. Let me know what you think.

experimental/sql/dialects/ibis_dialect.py
webmiche: I think I am capturing this distinction. Count can come with an empty string (meaning COUNT(*)) or with a column in COUNT(*that column*). I guess I could replace the `""` with a `"*"`, which might be more intuitive. What do you think?

webmiche: I will add a test case for the other behavior.

ingomueller-net: Aha, I see. I think I got confused by your introductory comment and then didn't understand the semantics of the arguments of that op. I don't mind, empty string or `"*"` work fine. I guess `"*"` could, in principle, be a valid column name, no? So that would be an argument against using that. Could `""` be a valid column name? I guess not?

What I'd ask you to do, though, is to define the semantics (and assumptions) of the arg. I guess it has to be a table with exactly one schema element, which defines the column name given as the argument to the function, right?

webmiche: We are talking about `Count(*)`, right?

I think in this case, the input can be any operator. And the decision which column to count this on could be made when some time during lowering, where we could predict which column might be most efficient. For example, it probably makes more sense to count on a fixed size column like an `i64` compared to counting on a `string` column.

In terms of the name, ibis does not care (so both "" and "*" are valid, also "_" and " "). Though there seem to be tools that do add restrictions (e.g. [ingres](https://docs.actian.com/ingres/11.0/index.html#page/SQLRef/SQL_Naming_and_Statement_Rules.htm)).

ingomueller-net: My question is about `arg`. What does that mean? Is that the "input table"? If yes, where do you specify the name of the column you want the count/min/max/... of?

For `COUNT(*)`, I think you don't actually need any column, do you? You can project away all of them, at least in the grouped version, where you still have the grouping attributes, and then simply count the number of times you see that group.

I am wondering whether `COUNT(*)` shouldn't be it's own function? It has a slightly different semantic, right? And that would solve the distinction problem via the column name.

webmiche: Ah, AFAIU, the arg is the "thing that is counted", so the column in one case and the operator in the other.

Yes, I agree. This feels like something I will revisit when working on the projection pushdown.

Hmm, yes that could certainly be an option. I will still need the `""` in the other list in order to properly align functions with their column names, but that would certainly solve the naming problem. I will look into this :)

webmiche: In the case that we have the ith element of functions as count, there can be two cases: either the ith element of col_names is "". Then this is a count over all the rows in a (grouped or ungrouped) operator. On the other han, if the ith element of col_names is the name of a column, that this is the count over all non null values in that column (of a grouped or ungrouped operator).

In both cases, the operator is just the thing that this aggregation consumes from.

ingomueller-net: I still don't fully understand it, sorry 😞 

> In both cases, the operator is just the thing that this aggregation consumes from.

... and that operator is what `arg` is used for, right? How do you pass in the name from `COUNT(name)`? I *think* that the  answer is that you don't; the name is provided by aggregate, which "zips" functions and column names, right?

If that's the case, why do we provide the function with one but not with the other? Why do we need to pass `arg`? Can't aggregate provide that as well?

webmiche: So this is still the ibis level, so there is a level of repetition present. Here, the arg correponds to what is aggregated. Notice that the same thing (at least without subqueries) is consumed from by the aggregation operator. So it is repeated in the input for the aggregation.

On this level, if we have a COUNT(*), this is visibile as:
```
ibis.count() {
   ibis.table() ...
}
```

whereas the count(name) is represented as:

```
ibis.count() {
  ibis.table_column() ...
}
```


On later levels, we change how aggregation is modeled, yo you are right there in that we zip function and the column name lists.

webmiche: Notice that we now know that the repetition in ibis is actually not q repetition, but can also be a subquery, so we will have to rethink our lowering approach here. But I think the modeling should still check out.

ingomueller-net: OK, thanks! I think I understand what `arg` is now. Please, please, please do add some explanation to the code, though!

PR #584, opened 2022-10-04
SQL: add GreaterThan to ibis and lowering to rel alg.

webmiche:
This PR adds handling of the GreaterThan node in ibis and its lowering to rel_alg. 

With this and #583, we can represent 9 queries of tpc-h. There would then be 12 more operations to go.


ingomueller-net (approved):
Nice! LGTM!

PR #585, opened 2022-10-04
SQL: add limit to the entire stack

webmiche:
This PR adds the limit node to the entire compilation stack. This is one of the more advanced operations that are still left, but it will give us another 3 queries, so I considered it worth it.

Notice that I do not model the `offset` part of the ibis expression as this is not needed for TPC-H at all.

### DIFF
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -764,6 +764,7 @@ class Ibis:
     self.ctx.register_op(Divide)
     self.ctx.register_op(UnboundTable)
     self.ctx.register_op(SortKey)
+    self.ctx.register_op(Limit)
     self.ctx.register_op(SchemaElement)
     self.ctx.register_op(Selection)
     self.ctx.register_op(CartesianProduct)
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -668,6 +668,33 @@ class UnboundTable(Operation):
 
 
 @irdl_op_definition
+class Limit(Operation):
+  """
+  Limits the number of tuples in `table` to `n` .
+
+  https://github.com/ibis-project/ibis/blob/f3d267b96b9f14d3616c17b8f7bdeb8d0a6fc2cf/ibis/expr/operations/relations.py#L337
+
+  Example:
+
+  ```
+  ibis.limit() ["n" = 10 : !i64] {
+    ...
+  }
+  ```
+  """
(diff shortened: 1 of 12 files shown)


ingomueller-net (approved):
In the rel alg dialect and possibly later, limit really only makes sense in direct combination with order by or on sequences. I guess that this continues the same discussion we already started but somehow emphasizes the need to think this through. What are the exact semantics now? The most precise I can think of is this: If there is an order by, it has to come as the root op, or the child of a limit op that is the root op, and the result of the two is really a sequence not a bag. That isn't very elegant is it? 😉

As before, I guess that modelling this as is for now should be fine but these open questions absolutely need discussion in the thesis!

Note that limit and order by can be fused to "top-k", and it is beneficial to do so at the execution layer. (In that case, you only need to keep and sort the top k elements seen so far while scanning the input and you can drop every new element that is smaller than the currently smallest.)

webmiche:
I am not quite sure that this is the case. Couldn't you just want 10 random samples from a table for example? Then you would also use limit, right? 

At least in the ibis tutorial, they are using limit on an unordered table (`countries['name', 'continent', 'population'].limit(3)`). Surely, the question is how useful that is and whether people would actually use it.

ingomueller-net:
> I am not quite sure that you limit without order doesn't make sense. Couldn't you just want 10 random samples from a table for example? Then you would also use limit, right?

True, you are right! The point is more subtle, then, and maybe more important: limit has a different meaning if preceded by order (and on a bag-only type system) than if not! You *don't* want 10 random samples if you have limit + order, right?

webmiche:
Yes, In terms of implementation, I would have implemented it as a boundary on the number of times you call next in your top-level iterator. In this case, you will just get 10 random samples in the unordered case and the top 10 otherwise.

But you are right. There might be optimization opportunities here for the 10 random samples one, as you can read "any" data and don't have to read all of it.

ingomueller-net:
If the pull-based execution works well, it will often do the right thing already: if you call 10 times pull on the child of "limit", this will trigger *just* the computations necessary to produce these 10 elements.

This doesn't quite work for sorting: the sort operator needs to read all input before producing any output. (Otherwise, any tuple that it may not have read might come before one of the tuples it already returned.) So if you have a naive order + limit, you probably order everything and then just read 10 tuples from the result. (The only way to avoid that without fusing is to implement some "lazy" sorting like heap sort, but that wouldn't be very efficient in most other case.)

But most of this is irrelevant for this PR 😉 What I do find important to keep in mind and discuss in the thesis are the shaky semantics with respect to ordering.

PR #588, opened 2022-10-05
SQL: add float64 types, fix small error

webmiche:
This PR continues the series of PRs to handle TPC-H queries. This one in particular adds the float64 type and its lowering. It also fixes an issue where count would only work on base tables and not on any operator.

### DIFF
--- a/experimental/sql/dialects/rel_alg.py
+++ b/experimental/sql/dialects/rel_alg.py
@@ -83,6 +83,20 @@ class Int64(DataType):
 
 
 @irdl_attr_definition
+class Float64(DataType):
+  """
+  Models a float64 type in a relational query.
+
+  Example:
+
+  ```
+  !rel_alg.float64
+  ```
+  """
+  name = "rel_alg.float64"
+
+
+@irdl_attr_definition
 class String(DataType):
   """
   Models a string type in a relational query, that can be either nullable or
@@ -488,6 +502,7 @@ class RelationalAlg:
     self.ctx.register_attr(Int64)
     self.ctx.register_attr(Nullable)
     self.ctx.register_attr(Order)
+    self.ctx.register_attr(Float64)
 
     self.ctx.register_op(Table)
     self.ctx.register_op(Limit)
--- a/experimental/sql/dialects/rel_impl.py
+++ b/experimental/sql/dialects/rel_impl.py
@@ -57,6 +57,20 @@ class Int64(DataType):
 
 
 @irdl_attr_definition
+class Float64(DataType):
+  """
+  Models a float64 type in a relational implementation query.
+
+  Example:
+
+  ```
+  !rel_impl.float64
+  ```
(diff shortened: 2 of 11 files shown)


ingomueller-net (approved):
Cool -- LGTM!

Where do you need this in TPC-H?

webmiche:
For Q11. One of the parameters is a fraction there.

```SQL
select
	ps_partkey,
	sum(ps_supplycost * ps_availqty) as value
from
	partsupp,
	supplier,
	nation
where
	ps_suppkey = s_suppkey
	and s_nationkey = n_nationkey
	and n_name = 'GERMANY'
group by
	ps_partkey having
		sum(ps_supplycost * ps_availqty) > (
			select
				sum(ps_supplycost * ps_availqty) * 0.0001000000
			from
				partsupp,
				supplier,
				nation
			where
				ps_suppkey = s_suppkey
				and s_nationkey = n_nationkey
				and n_name = 'GERMANY'
		)
order by
	value desc;

```

ingomueller-net:
I see, the parameter is called fraction; I didn't know. Are you sure it isn't a decimal, still? The query constant in the query you posted could be represented as a decimal...

webmiche:
I actually decided to drop Q11 for the moment as it is very weird. AFAIU, subqueries in general are modeled in ibis as part of the columns. So us dropping that part of the column node drops half of the queries for some examples :sweat_smile: 
I will look into doing this in a smarter way. Maybe with multiple queries in one module?

This query in particular is very strange, as it bascially computes a result in the inner query that is then used to filter in the outer query...

ingomueller-net:
In Oracle, `0.0001` is a `NUMBER` (aka decimal), see [this SO post](https://stackoverflow.com/questions/50534568/return-data-type-of-expression-in-oracle) and [this SQL fiddle](http://sqlfiddle.com/#!4/62d824/1).

ingomueller-net:
> I actually decided to drop Q11 for the moment as it is very weird. AFAIU, subqueries in general are modeled in ibis as part of the columns. So us dropping that part of the column node drops half of the queries for some examples 😅 I will look into doing this in a smarter way. Maybe with multiple queries in one module?
> 
> This query in particular is very strange, as it bascially computes a result in the inner query that is then used to filter in the outer query...

Aha, I'd be interested in these subqueries, how ibis represents them, and how to unnest them. Some of these nested queries turn into joins or anti-joins, and I am curious what the best way is to do it. I have never done this myself and I think there is quite some complexity to it, so I wouldn't recommend you to start that in the last week 😉

webmiche:
Yes, I agree. I think I have an understanding of what ibis does, but how to unnest them is certainly not feasible anymore.

I have thrown away queries with subqueries, so we are down to 5 queries again. I feel that this is still enough. I will look at the representations of the queries again to make sure we are really handling the full query, but I am quite confident.

PR #591, opened 2022-10-06
SQL: adjust proj pushdown

webmiche:
This PR adjusts the projection pushdown to handle the case of cartesian product, limit and order by. It also adds the ProjectionSimplifier that optimizes projections before new projections are added.

Currently, this is only tested on the TPC-H queries. I will add some testcases now.

### DIFF
--- a/experimental/sql/src/projection_pushdown.py
+++ b/experimental/sql/src/projection_pushdown.py
@@ -85,10 +85,21 @@ class ProjectionOptimizer(RewritePattern):
 @dataclass
 class ProjectionSimplifier(ProjectionOptimizer):
 
+  # This rewriter simpifies projections such that they only project columns that
+  # are acutlaly needed upstream.
+
   @op_type_rewrite_pattern
   def match_and_rewrite(self, op: RelAlg.Project, rewriter: PatternRewriter):
     cols = list(dict.fromkeys(self.find_cols_upstream(op.parent_op())))
     if None in cols or len(op.names.data) == len(cols):
+      return
+
+    input_schema = self.find_schema(op)
+    output_schema = self.flatten(
+        [self.find_cols_in_expr(o) for o in op.projections.ops])
+
+    if input_schema != output_schema or [s.data for s in op.names.data
+                                        ] != output_schema:
       return
 
     for o in op.projections.ops:
--- a/experimental/sql/src/projection_pushdown.py
+++ b/experimental/sql/src/projection_pushdown.py
@@ -213,6 +213,7 @@ def projection_pushdown(ctx: MLContext, query: ModuleOp):
       apply_recursively=False,
       walk_reverse=False)
   simplify_projections_walker.rewrite_module(query)
+
   infer_projections_walker = PatternRewriteWalker(GreedyRewritePatternApplier(
       [ProjectionInference()]),
                                                   walk_regions_first=False,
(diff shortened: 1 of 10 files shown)


experimental/sql/src/projection_pushdown.py
```diff
@@ -66,6 +74,42 @@ def find_schema(self, op: RelAlg.Operator) -> list[str]:
...
+      return self.find_schema(op.input.op)
+    if isinstance(op, RelAlg.Limit):
+      return self.find_schema(op.input.op)
+
+
+@dataclass
+class ProjectionSimplifier(ProjectionOptimizer):
+
+  # This rewriter simpifies projections such that they only project columns that
+  # are acutlaly needed upstream.
```
ingomueller-net, line 89:
```suggestion
  # are actually needed upstream.
```

experimental/sql/src/projection_pushdown.py
```diff
@@ -66,6 +74,42 @@ def find_schema(self, op: RelAlg.Operator) -> list[str]:
...
+  def match_and_rewrite(self, op: RelAlg.Project, rewriter: PatternRewriter):
+    cols = list(dict.fromkeys(self.find_cols_upstream(op.parent_op())))
+    if None in cols or len(op.names.data) == len(cols):
+      return
+
+    input_schema = self.find_schema(op)
+    output_schema = self.flatten(
+        [self.find_cols_in_expr(o) for o in op.projections.ops])
+
+    if input_schema != output_schema or [s.data for s in op.names.data
```
ingomueller-net, line 101: What does this test do? If the input is *not* different from the output, they are the same, so the projection didn't do anything, no? Is this the case that is handled here? If so, I don't understand it from the documentation/comments...

experimental/sql/src/projection_pushdown.py
```diff
@@ -84,12 +128,26 @@ def match_and_rewrite(self, op: RelAlg.Operator, rewriter: PatternRewriter):
     if None in cols:
       return
 
-    new_proj = RelAlg.Project.get(
-        Region.from_operation_list([op.input.op.clone()]),
-        Region.from_operation_list([RelAlg.Column.get(s) for s in cols]),
-        ArrayAttr.from_list([StringAttr.from_str(s) for s in cols]))
-
-    rewriter.replace_op(op.input.op, new_proj)
+    if not isinstance(op, RelAlg.CartesianProduct):
```
ingomueller-net, line 131: Isn't is more future-proof to do it the other way around? Handle the `isinstance(op, CartesianProduct)` case first, then handle other special cases in the future, and finally handle the default case?

ingomueller-net (commented):
Probably looks good but there are a couple of things I don't really understand. I have added questions in some places but more comments would be helpful in others as well.

experimental/sql/src/projection_pushdown.py
webmiche: I added more comments to clarify.

experimental/sql/test/projection_pushdown_tests/order_by_extended.xdsl
```diff
@@ -0,0 +1,32 @@
+// RUN: rel_opt.py -p projection-pushdown %s | filecheck %s
+
+module() {
+  rel_alg.aggregate() ["col_names" = ["a"], "functions" = ["sum"], "res_names" = ["a_sum"], "by" = []] {
+    rel_alg.order_by() ["by" = [!rel_alg.order<"a", "desc">, !rel_alg.order<"b", "asc">]] {
```
ingomueller-net, line 5: Completely off-topic but there is an opportunity to drop order-bys that don't have an effect, such as this one. 

With aggregate, you have to know that the aggregate function is commutative, which, if you consider overflows, might not actually be the case. But there might be other cases, such as two subsequent order-bys, or an order-by followed by a join (if we assume that the join produces its own order and isn't required to preserve the existing one.)

Never mind, just my brain producing ideas... 😉

ingomueller-net (approved):
Cool, thanks for the update, this LGTM now!

PR #593, opened 2022-10-06
Iterators: Add lowering to LLVM for new TabularViewType

ingomueller-net:
This is a first of a series of PR that should replace #577. It depends on and therefore currently includes #590 and #592.

PR #594, opened 2022-10-06
Iterators: Add CreateTabularViewOp

ingomueller-net:
This is a first of a series of PR that should replace #577. It depends on and therefore currently includes #590, #592, and #593.

### DIFF
--- a/experimental/iterators/include/iterators/Dialect/Tabular/IR/TabularOps.td
+++ b/experimental/iterators/include/iterators/Dialect/Tabular/IR/TabularOps.td
@@ -17,4 +17,50 @@ class Tabular_Op<string mnemonic, list<Trait> traits = []> :
   Op<Tabular_Dialect, mnemonic, traits> {
 }
 
+/// MemRefType that has static shape and contiguous row-major layout.
+def IsStaticShapeAndContiguousRowMajorMemRefTypePred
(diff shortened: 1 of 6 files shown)


experimental/iterators/include/iterators/Dialect/Tabular/IR/TabularOps.td
```diff
@@ -17,4 +17,50 @@ class Tabular_Op<string mnemonic, list<Trait> traits = []> :
...
+    contiguous memory layout. Furthermore, no dynamic dimension is allowed
+    currently (in order to avoid runtime checks).
+
+    Example:
+    ```mlir
+      %t1 = arith.constant dense<[0, 1, 2]> : tensor<3xi32>
+      %t2 = arith.constant dense<[3, 4, 5]> : tensor<3xi64>
+      %m1 = bufferization.to_memref %t1 : memref<3xi32>
+      %m2 = bufferization.to_memref %t2 : memref<3xi64>
+      %view = "tabular.create_tabular_view"(%m1, %m2)
```
nicolasvasilache, line 51: I'd just drop the "create" prefix which may suggest "alloc".

nicolasvasilache: view_as_tabular returns a tabular_view

PR #597, opened 2022-10-07
SQL: fix projection pushdown and partial loading

webmiche:
This PR addresses an issue I introduced when "fixing" the handling of renames in the projection pushdown, in particular the projection simplifier. What I introduced was to restrictive to handle the TPC-H queries. So, I rewrote it to be able to handle some renames.

I also fixed a problem in the partial loading where projections would not be fused at all if there are complex projections present. I rewrote this to a version that does he partial loading, but also keeps the projection around.

### DIFF
--- a/experimental/sql/dialects/rel_impl.py
+++ b/experimental/sql/dialects/rel_impl.py
@@ -585,7 +585,7 @@ class Project(Operator):
   def get(input: Operation, res_names: List[str], res_types: List[DataType],
           projection: Region) -> 'Project':
     return Project.create(operands=[input.result],
-                          result_types=[Bag.get(res_names, res_types)],
+                          result_types=[Bag.get(res_types, res_names)],
                           regions=[projection])
 
   @staticmethod
--- a/experimental/sql/src/fuse_proj_into_scan.py
+++ b/experimental/sql/src/fuse_proj_into_scan.py
@@ -25,29 +25,45 @@ class ProjScanFuser(RewritePattern):
     if not isinstance(op.input.op, RelImpl.FullTableScanOp):
       return
 
-    # This rewrite expects that the projection only maps onto a subset of the
-    # columns of the input. Therefore, we first check that only IndexByName is
-    # present and then that there are equally as many cols that are yielded as
-    # there are in the result.
-    if not all([
-        isinstance(o.op, RelImpl.IndexByName) for o in op.projection.ops[-1].ops
-    ]):
-      return
-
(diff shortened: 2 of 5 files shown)


experimental/sql/dialects/rel_impl.py
```diff
@@ -585,7 +585,7 @@ class Project(Operator):
   def get(input: Operation, res_names: List[str], res_types: List[DataType],
           projection: Region) -> 'Project':
     return Project.create(operands=[input.result],
-                          result_types=[Bag.get(res_names, res_types)],
```
webmiche, line 588: This is an unrelated change. I can remove it, if you prefer.

ingomueller-net: Leave it as is now but I prefer separate PRs for the future.

experimental/sql/src/fuse_proj_into_scan.py
```diff
@@ -25,29 +25,45 @@ def match_and_rewrite(self, op: RelImpl.Project, rewriter: PatternRewriter):
...
-    ]):
-      return
-
     assert isinstance(op.projection.ops[-1], RelImpl.YieldTuple)
     yielded_ssa_values = op.projection.ops[-1].ops
 
-    col_names = [o.op.col_name.data for o in yielded_ssa_values]
-    res_names = [s.elt_name.data for s in op.input.typ.schema.data]
-    if len(col_names) == len(res_names):
+    # Get the list of used columns. The dict magic removes duplicates.
```
ingomueller-net, line 31: Isn't this the typical idiom: `list(set([1,1,1,1,1]))`?

ingomueller-net: Ah, and then you can avoid the construction of the inner list by making it a generator expression: `list(set((1 for i in range(4))))`

ingomueller-net (approved):
LGTM! Much easier to understand as well :)

experimental/sql/src/fuse_proj_into_scan.py
webmiche: I think the main difference is that the dict version keeps the order, that's why its my default go to. Order doesn't matter here though, so I am fine with changing to set, if you feel its necessary.

ingomueller-net: Ah, good point about the order. No, pick whatever you think fits best. (I think the generator expression in favor of the list comprehension doesn't have any disadvantages, though.)

PR #598, opened 2022-10-11
SQL: Add between to ibis and lower it to rel_alg

webmiche:
This PR adds the modeling for the ibis `Between` node and lowers it to two comparisons in rel_alg.

### DIFF
--- a/experimental/sql/dialects/ibis_dialect.py
+++ b/experimental/sql/dialects/ibis_dialect.py
@@ -858,7 +858,7 @@ class SchemaElement(Operation):
 @irdl_op_definition
 class Between(Operation):
   """
-  Returns wheter `arg` is in between `lower_bound` and `upper_bound`.
+  Returns whether `arg` is in between `lower_bound` and `upper_bound`. These bounds are inclusive.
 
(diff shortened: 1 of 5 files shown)


experimental/sql/dialects/ibis_dialect.py
```diff
@@ -855,6 +855,36 @@ def get(name: str, type: DataType):
     })
 
 
+@irdl_op_definition
+class Between(Operation):
+  """
+  Returns wheter `arg` is in between `lower_bound` and `upper_bound`.
```
ingomueller-net, line 861:
```suggestion
  Returns whether `arg` is in between `lower_bound` and `upper_bound`.
```

ingomueller-net: Please also define whether the two bounds are inclusive or exclusive.

ingomueller-net (approved):
Nice -- LGTM!

Philosophical question: This can be done as part of the lowering but also as a rewrite *on the ibis dialect*. This seems to be an instance of a more general question: simplify on a given dialect or do that simplification on the fly when lowering to the next one. Any thoughts on your side? Was this a conscious decision or just the first thing that came to your mind?

webmiche:
IMO, having lowerings that are more involved than just code expansion can become extremely tricky and has the same issue as the transformation-based design: If you change a dialect, you need to touch and update more rewrites than otherwise required. I think this is something that I would count as "separation of concerns".

PR #599, opened 2022-10-11
SQL. add support for and

webmiche:
This PR adds support for the `And` not. In our current prototype, this can only occur in selections over lists of conditions, something like:

```python
table.filter([table.a == 5 & table.b == 7])
```

However, this is equivalent to writing:

```python
table.filter([table.a == 5, table.b == 7])
```

With this change, we lower to the latter form, if the former is encountered.

This might be a bit hacky, but it works for the TPC-H queries nad it closes the second not-so-nice thing about our representations.

### DIFF
--- a/experimental/sql/src/ibis_frontend.py
+++ b/experimental/sql/src/ibis_frontend.py
@@ -72,7 +72,11 @@ def visit_schema(schema: ibis.expr.schema.Schema) -> Region:
 def visit_ibis_expr_list(l: List[ibis.expr.types.Expr]) -> Region:
   ops = []
   for op in l:
-    ops.append(visit(op))
+    if isinstance(op.op(), ibis.expr.operations.logical.And):
+      ops.append(visit(op.op().left))
+      ops.append(visit(op.op().right))
+    else:
+      ops.append(visit(op))
   return Region.from_operation_list(ops)
 
 
--- a/experimental/sql/src/ibis_frontend.py
+++ b/experimental/sql/src/ibis_frontend.py
@@ -72,6 +72,10 @@ def visit_schema(schema: ibis.expr.schema.Schema) -> Region:
 def visit_ibis_expr_list(l: List[ibis.expr.types.Expr]) -> Region:
   ops = []
   for op in l:
+    # TODO: This is a hack that rewrites top-level Ands. To do this properly,
+    # and handle all adds, first model And in ibis, then either rewrite this to
+    # a version using an optimization on the ibis dialect or model And in
+    # rel_alg and lower to that.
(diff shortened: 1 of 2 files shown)


experimental/sql/src/ibis_frontend.py
```diff
@@ -72,7 +72,11 @@ def visit_schema(schema: ibis.expr.schema.Schema) -> Region:
 def visit_ibis_expr_list(l: List[ibis.expr.types.Expr]) -> Region:
   ops = []
   for op in l:
-    ops.append(visit(op))
+    if isinstance(op.op(), ibis.expr.operations.logical.And):
```
ingomueller-net, line 75: What if there are several nested `And`s?

Similar question here as before: this could also be a simplification/normalization pass on the ibis dialect -- or on a later one. Note that there might be slightly more complicated cases where it is possible to detect a conjunction, for example, by applying a distributive equivalence first (`ab | bc` --> `b & (a | c)`). Any thoughts?

ingomueller-net (commented):
A good thing to add but I'd like to hear your thoughts on where this should live.

I have the feeling that this really just handles one specific corner case of a much broader class of cases for detecting conjunctions in a filter. Unlike #598, I am thus inclined to prefer this to be done as an optimization pass rather than doing it as part of the lowering. What do you think?

ingomueller-net:
Anticipating your reply: if you want to do a hacky thing now given the tight timeline, please mark it as such in the code and document briefly what a better solution could look like.

webmiche:
I agree again with this being an optimization and not a lowering. Again, I took the route of least resistance for this prototype, as this is the last thing I want to add and this really only being something that is added to make one plot a bit nicer and delete one sentece claiming a difference.

I feel like I would have the time to implement this differently, i.e., modeling the `And` and then rewriting it during lowering or doing it as an optimization pass, or even a step further, modeling `And` in rel_alg and getting a proper design this way. I don't really mind which way we go here.

Do you feel this is necessary? Or do you think explaining the optimized designs in a todo would suffice?

experimental/sql/src/ibis_frontend.py
webmiche: This is something that we do not handle atm. It is also not something actually needed for the TPC-H queries, that's why I made the decision to ignore it.

webmiche:
I will do some other things that I feel are more important atm, like getting the tables in shape, and then I will revisit this closer to the deadline again.

ingomueller-net:
> Do you feel this is necessary? Or do you think explaining the optimized designs in a todo would suffice?

No, as I said, marking it as a hack with a brief description of how you imagine it being done correctly is enough.

PR #600, opened 2022-10-11
[NFC] Iterators: Make TabularToLLVM compose with IteratorsToLLVM.

ingomueller-net:
This commit makes TabularTypeConverter part of the public API of TabularToLLVM such that other conversions can reuse that converter. It also turns TabularToLLVM into a partial conversion such that it can be run before IteratorsToLLVM (i.e., with ops from that dialect present).

### DIFF
--- a/experimental/iterators/lib/Conversion/TabularToLLVM/TabularToLLVM.cpp
+++ b/experimental/iterators/lib/Conversion/TabularToLLVM/TabularToLLVM.cpp
@@ -35,39 +35,30 @@ struct ConvertTabularToLLVMPass
 };
 } // namespace
 
-/// Maps types from the Tabular dialect to corresponding types in LLVM.
-class TabularTypeConverter : public TypeConverter {
-public:
-  TabularTypeConverter(LLVMTypeConverter &llvmTypeConverter)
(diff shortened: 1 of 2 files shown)


experimental/iterators/lib/Conversion/TabularToLLVM/TabularToLLVM.cpp
```diff
@@ -174,7 +165,7 @@ void ConvertTabularToLLVMPass::runOnOperation() {
   typeConverter.addSourceMaterialization(addUnrealizedCast);
   typeConverter.addTargetMaterialization(addUnrealizedCast);
 
-  if (failed(applyFullConversion(module, target, std::move(patterns))))
```
nicolasvasilache, line 177: This line should be in a different commit ?

nicolasvasilache (approved):

experimental/iterators/lib/Conversion/TabularToLLVM/TabularToLLVM.cpp
ingomueller-net: Done! Made it part of #579.

PR #604, opened 2022-10-20
Iterators: Fix handling of operands produced by non-iterator ops.

ingomueller-net:
The current conversion driver was handling operands produced by non-iterator ops incorrectly. In case they needed a type conversion, an unrealized cast was inserted correctly; in case they don't, they were simply dropped. That case didn't occur anywhere until now but I ran into it in a work-in-progress PR.


nicolasvasilache (changes requested):
This needs a test, maybe fold this change as part of the PR that needs it

ingomueller-net:
This PR will be part of #614 (or some successor of it). Closing.

PR #606, opened 2022-10-21
Iterators: Create iterator state type and extract/insert/undef ops.

ingomueller-net:
This PR depends on and therefore includes #605.

One point for discussion is the assembly format, which we may want to decide to change in order to make it similar to other ops with similar semantics.

### DIFF
--- a/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
+++ b/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
@@ -310,7 +310,7 @@ def Iterators_SinkOp : Iterators_Base_Op<"sink"> {
 def Iterators_UndefStateOp : Iterators_Base_Op<"undefstate", [NoSideEffect]> {
   let summary = "Create an undefined iterator state";
   let results = (outs Iterators_State:$result);
-  let assemblyFormat = "attr-dict `:` type($result)";
+  let assemblyFormat = "attr-dict `:` qualified(type($result))";
   let description = [{
     Creates an iterator state of the given type with undefined field values.
     All fields have to be set individually with `insertvalue` before the whole
@@ -340,7 +340,8 @@ def Iterators_ExtractValueOp : Iterators_Base_Op<"extractvalue", [
   let arguments = (ins Iterators_State:$state, IndexAttr:$index);
   let results = (outs AnyType:$result);
   let assemblyFormat = [{
-    $state `[` $index `]` attr-dict `:` type($state) `->` type($result)
+    $state `[` $index `]` attr-dict `:`
+      qualified(type($state)) `->` qualified(type($result))
   }];
   let description = [{
     Extracts the value of the given iterator state at the given index.
@@ -372,7 +373,8 @@ def Iterators_InsertValueOp : Iterators_Base_Op<"insertvalue", [
   let arguments = (ins Iterators_State:$state, IndexAttr:$index, AnyType:$value);
   let results = (outs Iterators_State:$result);
   let assemblyFormat = [{
-    $state `[` $index `]` `(` $value `:` type($value) `)` attr-dict `:` type($result)
+    $state `[` $index `]` `(` $value `:` qualified(type($value)) `)`
+      attr-dict `:` qualified(type($result))
   }];
   let description = [{
     Inserts the given value into the given iterator state at the given index.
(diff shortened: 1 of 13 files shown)


experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -303,4 +303,90 @@ def Iterators_SinkOp : Iterators_Base_Op<"sink"> {
...
+//===----------------------------------------------------------------------===//
+// Ops related to Iterator bodies.
+//===----------------------------------------------------------------------===//
+
+def Iterators_UndefStateOp : Iterators_Base_Op<"undefstate", [NoSideEffect]> {
+  let summary = "Create an undefined iterator state";
+  let results = (outs Iterators_State:$result);
+  let assemblyFormat = "attr-dict `:` type($result)";
+  let description = [{
+    Creates an iterator state of the given type with undefined field values.
```
nicolasvasilache, line 315: Please take a look at this as background info: https://discourse.llvm.org/t/rfc-poison-semantics-for-mlir/66245

experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsOps.td
```diff
@@ -303,4 +303,90 @@ def Iterators_SinkOp : Iterators_Base_Op<"sink"> {
...
+    All fields have to be set individually with `insertvalue` before the whole
+    state is fully defined.
+
+    This is similar to `llvm.undef` for `llvm.struct`.
```
nicolasvasilache, line 319: Related to the above, I am not sure whether llvm.under for llvm.struct is a "good thing" or a "mistake from the past".

Would be nice to dig a bit and understand.

nicolasvasilache (approved):

ingomueller-net:
From the offline discussion just now:

1. Enforce qualified type to avoid unreadable `<i32>` abbreviations.
2. Make an attempt to elide the return type in the assembly format of `iterators.extractvalue`/value type of `iterators.insertvalue`, which are redundant.

ingomueller-net:
TODOs for next iteration:

- [ ] Add tests for round-tripping printing and parsing.
- [ ] Use qualified type instead of abbreviated type.
- [ ] Attempt to elide redundant types in custom assembly format.
- [ ] Understand undef/poison issue and decide whether/when to apply.

ingomueller-net:
Here are some things I found out about `poison`:

* Disclaimer: I am not 100% sure that I understand things correctly, so take everything with a grain of salt.
* `poison` is similar to `undef` in that it allows to express "deferred undefined behavior." This allows IR to express behavior that may be undefined "locally", but which may never occur at runtime and therefore be correct.
* `poison` is different from `undef` in that the latter can be thought of "any valid value from the domain" whereas the former can be thought of as "an invalid value." This may make a difference: Consider `a > INT_MAX` where `a` is of type `int`, then if `a` were to become `undef`, we can still know that the comparison is `false` (because no value of the `int` domain is larger than `INT_MAX`) but if `a` were to become `poison`, the comparison is also `poison` (by definition).
* If I understand things correctly, the goal is to use `poison` only to define the semantics of instructions and *not* at runtime. If instructions are defined to return `poison` instead of `undef`, their undefined behavior is "more undefined", so more optimizations are valid.
* The ["Taming undefined behavior in LLVM" paper](https://www.cs.utah.edu/~regehr/papers/undef-pldi17.pdf) argues to stop using `undef` and to use only `poison` instead. The main problem seems to be that the exact semantics of both `poison` and `undef` haven't been defined (or respected) in LLVM, so there are inconsistent assumptions in different optimizations. (I have not fully understood why their proposal solves theses, though.)
* [This recent RFC](https://discourse.llvm.org/t/rfc-poison-semantics-for-mlir/66245) proposes a way to allow `poison` values in arbitrary dialects.
* It is meant to be opt-in, and it should be possible to adopt it on a rolling basis.
* While the idea is to start implementing/show-casing this proposition in the `LLVM` and other low-level dialects, I believe that this hasn't happened (or even started) yet.

In light of all this, I got the impression that implementing `poison` in this PR is premature. We aren't doing any optimizations on our type yet, so there is no benefit. Also, due to the recency of the MLIR proposal, there aren't any examples yet that we could take inspiration from. Finally, it seems like `poison` semantics can be adopted over time and, in particular in our case, it seems like we can simply replicate LLVM's dealing with `poison` on its structs once it has adopted it.

@nicolasvasilache: What do you think?

nicolasvasilache:
SGTM thanks for looking into this @ingomueller-net !

PR #607, opened 2022-10-21
Iterators: Implement lowering of iterator state type and ops.

ingomueller-net:
This PR depends on and therefore includes #605 and #606.


nicolasvasilache (commented):
We need some ops parsing/printing test here.

nicolasvasilache (changes requested):
Sorry spoke too soon, we need to ensure the roundtripping works properly and we need tests for that: `mlir-opt %s | mlir-opt`

nicolasvasilache (approved):

ingomueller-net:
> We need some ops parsing/printing test here.

This is actually a comment for #606 and I have addressed it there.

ingomueller-net:
> Sorry spoke too soon, we need to ensure the roundtripping works properly and we need tests for that: `mlir-opt %s | mlir-opt`

Dito: related to #606 and addressed there.

PR #608, opened 2022-10-21
Iterators: Use new iterator states in IteratorsToLLVM.

ingomueller-net:
This PR depends on and therefore includes #605, #606, and #607.

This PR finalizes the removal of one big dependency on LLVM, namely that for using LLVM structs for the states of iterators. Instead, the new type introduced for that purpose is used.

We may consider renaming the pass, either now, or after we removed the other dependencies to LLVM. (There are still several other dependencies, but they aren't related to iterators *in general* but rather to specific iterator ops such as `TabularViewToStreamOp` returning each row as an LLVM struct instead of a dedicated row type.)

### DIFF
--- a/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
+++ b/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorAnalysis.cpp
@@ -2,14 +2,12 @@
 
 #include "iterators/Dialect/Iterators/IR/Iterators.h"
 #include "iterators/Utils/NameAssigner.h"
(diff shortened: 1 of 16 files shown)


nicolasvasilache (approved):
Please take a look at the review discussion in this PR and make sure that this abides by the conclusion: https://reviews.llvm.org/D136377.

Accepting conditioned on that.

ingomueller-net:
> Please take a look at the review discussion in this PR and make sure that this abides by the conclusion: https://reviews.llvm.org/D136377.

This is a comment for #615, right?

PR #610, opened 2022-10-26
WIP: Enable cache of setup-python Github action.

ingomueller-net:


ingomueller-net:
Doesn't seem to cache if `requirements.txt` doesn't fix versions. That step only takes about 1.5 minutes. Not worth spending more time...

PR #615, opened 2022-10-28
Iterators: Use arith.constant where unrelated to LLVM.

ingomueller-net:
This PR depends on and therefore includes #608 (and, transitively, others). It removes some minor dependencies to the LLVM dialect via `llvm.mlir.constant`, which weren't necessary in any way.

### DIFF
--- a/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
+++ b/experimental/iterators/lib/Conversion/IteratorsToLLVM/IteratorsToLLVM.cpp
@@ -157,7 +157,7 @@ struct ConstantTupleLowering : public OpConversionPattern<ConstantTupleOp> {
       // Create constant value op.
       Attribute field = values[i];
       Type fieldType = field.getType();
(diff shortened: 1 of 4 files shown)


nicolasvasilache (commented):
Please take a deep look at the discussion in https://reviews.llvm.org/D136377 and decide accordingly.

ingomueller-net:
> Please take a deep look at the discussion in https://reviews.llvm.org/D136377 and decide accordingly.

@nicolasvasilache: Thanks for the pointer! It seems like that discussion is about a change in the the opposite direction, namely using some conversions from `arith` to `LLVM` as part of some lowering pass. In contrast, this PR is about lowering to `arith` instead of lowering directly to `LLVM`. I have the feeling that, if anything, this PR "undoes" a mangling that was considered bad in the discussion you linked to. (Though it never really did that in the first place: it didn't pull in patterns from a different conversion pass but rather created `LLVM` ops directly where `arith` ops would have been more appropriate.)

nicolasvasilache (approved):

PR #614, opened 2022-10-27
Iterators: Implement ops to convert values to and from streams.

ingomueller-net:
This PR ~~depends on and therefore~~ includes #616 and #617.

PR #616, opened 2022-10-31
Iterators: Implement ops to convert values to and from streams.

ingomueller-net:


ingomueller-net:
As discussed offline: will merge as part of #614.

PR #617, opened 2022-10-31
Iterators: Implement lowering for ValueToStreamOp.

ingomueller-net:
This PR depends on and therefore includes #616. It includes the fix of the abandoned #604, for which it now also provides a test case.


ingomueller-net:
As discussed offline: will merge as part of #614.

PR #626, opened 2022-12-09
CI: Update ninja dependency

webmiche:
This is something I did for a couple of other repos recently, so let me mention it here as well:

It seems that the llvm action to install Ninja (and also checkout v2) uses deprecated nodes (visibile [here](https://github.com/iree-org/iree-llvm-sandbox/actions/runs/3628671385)). This updates to a properly configured action. Alternatively, one could probably also update the llvm action and close this PR...

I just did it for a single CI workflow, the others would obviously have to be updated as well, if you decide to use this.

### DIFF
--- a/.github/workflows/buildAndTestDatabaseIteratorsStandalone.yml
+++ b/.github/workflows/buildAndTestDatabaseIteratorsStandalone.yml
@@ -22,12 +22,12 @@ jobs:
         python-version: 3.9
 
     - name: Checkout project
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         path: sandbox
 
     - name: Install Ninja
-      uses: llvm/actions/install-ninja@55d844821959226fab4911f96f37071c1d4c3268
+      uses: lukka/get-cmake@latest
 
     - name: Install doxygen
       run: |


.github/workflows/buildAndTestDatabaseIteratorsStandalone.yml
```diff
@@ -22,12 +22,12 @@ jobs:
...
 
     - name: Checkout project
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         path: sandbox
 
     - name: Install Ninja
-      uses: llvm/actions/install-ninja@55d844821959226fab4911f96f37071c1d4c3268
+      uses: lukka/get-cmake@latest
```
ingomueller-net, line 30: I am not sure if we'll want to change to some third-party action. I looked at the action upstream and it could be a single-character change. I'll look into that...

ingomueller-net (commented):
Thanks for looking into that! No problem about the `checkout` action but I am skeptical about using some third-party action for ninja...

ingomueller-net:
FYI, I have just submitted [this](https://github.com/llvm/actions/pull/4) PR to the `llvm/actions` repo.

webmiche:
Ah, nice. Thanks!

PR #630, opened 2022-12-14
[DO NOT MERGE]: Test updated LLVM GitHub actions 

ingomueller-net:


ingomueller-net:
The updated action is being used for 10 days now and nobody seems to have had problems.

PR #636, opened 2023-01-18
[Iterators] Update LLVM submodule to a version from September 27th (df7606a0) and fix breakages.

ingomueller-net:


ingomueller-net:
The broken CI target seems to come from a CMake problem in the IREE repo. I can make the build work with the following patch:

```patch
diff --git a/llvm-external-projects/iree-dialects/lib/Dialect/LinalgExt/Passes/CMakeLists.txt b/llvm-external-projects/iree-dialects/lib/Dialect/LinalgExt/Passes/CMakeLists.txt
index 68991a666..3f84ad4cc 100644
--- a/llvm-external-projects/iree-dialects/lib/Dialect/LinalgExt/Passes/CMakeLists.txt
+++ b/llvm-external-projects/iree-dialects/lib/Dialect/LinalgExt/Passes/CMakeLists.txt
@@ -11,6 +11,7 @@ add_mlir_library(IREELinalgExtPasses
   LINK_LIBS PUBLIC
   IREEInputDialect
   IREELinalgExtDialect
+  IREELinalgExtTransforms
   MLIRAffineDialect
   MLIRIR
   MLIRLinalgDialect
```

I don't know yet whether this problem persists in the current `main`. If so, I'll provide a fix there.

ingomueller-net:
I can't reproduce the CI breakage yet but I can rule out that it is related to above patch. The patch fixes a problem that appears if building with `BUILD_SHARED_LIBS=ON`, which I do on my machine but CI does not. My impression is that the CMake files of the main IREE repository are broken in several places but the breakage hasn't been apparent since it doesn't have any consequences for static linking...

ingomueller-net:
Since I could not make the integration through IREE work, this is now based on https://github.com/iree-org/iree-llvm-sandbox/pull/641. I also removed all changes that are unrelated to the iterators dialect such that the remainder of the sandbox continues to work with the version of IREE and LLVM it currently uses.

ingomueller-net:
Abandoning in favor of #640.

PR #638, opened 2023-01-18
[Iterators] Update LLVM submodule to a version from November 23rd (61aed52c) and fix breakages.

ingomueller-net:


ingomueller-net:
Abandoning in favor of #640.

PR #641, opened 2023-01-20
[Iterators] Make build, test, and CI infra independent of remaining repo.

ingomueller-net:
I have run into problems building the IREE dependencies of newer than the current version that I couldn't easily solve. Since the iterators don't actually depend on the rest of the sandbox and IREE, it makes sense to separate its infrastructure. This PR thus turns the `experimental/iterators` folder into a dedicated CMake project (using the external project mechanism of LLVM) with its dedicated git submodule for LLVM, and dedicated entry points for tests.


nicolasvasilache (approved):
Makes sense, did not go into details

PR #644, opened 2023-01-21
[broken][IREE] Update IREE submodule and make compile (but not pass CI).

ingomueller-net:
This PR collects the changes I made while attempting to update the whole sandbox to the latest (Jan 10, 2023) version of the IREE submodule. I eventually gave up because of issues in the CMake files of that repository that I couldn't resolve in CI. The commits here can still serve if an update should be done at a later stage.


ingomueller-net:
@nicolasvasilache: FYI.

PR #646, opened 2023-01-25
[Iterators] Introduce pass that decomposes iterator states into its constituent values.

ingomueller-net:
~~This PR depends on and, therefore, includes #650 and #655. I might also factor out the update of LLVM into a different PR before landing.~~

This PR implements a pass that decomposes iterator states into its constituent values instead of lowering them to (nested) LLVM structs. The problem with the previous approach is that it only supports storing LLVM types in the state, which includes many interesting things such as `tensor`s or `memref`s. By decomposing the state into individual fields, other passes can run on this fields without knowing anything about iterators (or LLVM structs) and, thus, makes iterators composable with other dialects.

- [x] Move boilerplate changes to first commit.
- [x] Remove forward declarations of `populate*`, which are now in a different compilation unit.

### DIFF
--- a/experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
+++ b/experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
@@ -10,8 +10,13 @@
 
 #include "iterators/Dialect/Iterators/IR/Iterators.h"
 #include "iterators/Dialect/Iterators/Transforms/Passes.h"
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Func/Transforms/OneToNFuncConversions.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/Dialect/SCF/Transforms/Transforms.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/OneToNTypeConversion.h"
 
 namespace mlir {
 #define GEN_PASS_CLASSES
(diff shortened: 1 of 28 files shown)


experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
```diff
@@ -0,0 +1,251 @@
...
+    auto stateType = op.getState().getType().cast<StateType>();
+    TypeRange originalFieldTypes = stateType.getFieldTypes();
+    OneToNTypeMapping fieldMapping(originalFieldTypes);
+    if (failed(typeConverter->convertSignatureArgs(originalFieldTypes,
+                                                   fieldMapping)))
+      return failure();
+
+    // Extract converted operands.
+    ValueRange convertedState =
+        operandMapping.getConvertedValues(convertedOperands, 0);
```
nicolasvasilache, line 85: is there an opportunity to use the Adaptor to get to the magic constant index?

ingomueller-net: Yeah, that'd be nice. This is an aspect where the 1:N type conversion could be improved. However, I think that the operand indices are currently not exposed. The are *used* by the accessor methods but they aren't exposed.

What would also be nice is if I could create an adaptor with the converted operands insight like the traditional dialect conversion does -- but then an accessor (like `getState()`) would have to return a value *range* rather than a single value. Is that something you think I should investigate/push for?

ingomueller-net: Ha! I looked at the adaptors again: they look more powerful from what I remember from when I looked last time. There is (now?) a templated `XOpGenericAdaptor`, which you can wrap around a range of anything and then have the accessors access that. I believe can wrap that around a range of ranges such that the accessor returns the *range* of converted values. That would be an awesome solution upstream. I'll report back this afternoon.

ingomueller-net: OK, this issue will be fixed upstream if https://reviews.llvm.org/D147225 is accepted. Can we allow this PR with the magic constants and improve upon it when we integrate the upstream changes?

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
```diff
@@ -0,0 +1,243 @@
...
+  patterns.add<
+      // clang-format off
+      DecomposeExtractValueOp,
+      DecomposeInsertValueOp,
+      DecomposeCreateStateOp
+      // clang-format on
+      >(typeConverter, patterns.getContext());
+}
+
+static std::optional<Value> buildCreateStateOp(OpBuilder &builder,
```
nicolasvasilache, line 144: doc plz

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+    operands.push_back(operand);
+  }
+  auto createStateOp = builder.create<CreateStateOp>(loc, type, operands);
+  assert(createStateOp->getNumResults() == 1);
+  return createStateOp->getResult(0);
+}
+
+std::optional<SmallVector<Value>>
+buildExtractValueOps(OpBuilder &builder, TypeConverter &typeConverter,
```
nicolasvasilache, line 178: static plz

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
```diff
@@ -0,0 +1,243 @@
...
+                     TypeRange resultTypes, Value input, Location loc) {
```
nicolasvasilache, line 179: comments plz

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
```diff
@@ -0,0 +1,243 @@
...
+      extractedValues.push_back(extractedValue);
+      continue;
+    }
+
+    // For states, we recurse to extract the nested fields.
+    SmallVector<Type> nestedResultTypes;
+    if (failed(typeConverter.convertType(extractedValue.getType(),
+                                         nestedResultTypes)))
+      return {};
+    std::optional<SmallVector<Value>> maybeResults = buildExtractValueOps(
```
nicolasvasilache, line 200: this is recursive but the depth is small ?

experimental/iterators/test/Integration/Dialect/Iterators/CPU/constant-stream.mlir
```diff
@@ -1,6 +1,6 @@
 // RUN: iterators-opt %s \
 // RUN:   -convert-iterators-to-llvm \
-// RUN:   -convert-states-to-llvm \
+// RUN:   -decompose-iterator-states -canonicalize \
```
nicolasvasilache, line 3: Do we need all these canonicalize ?
IOW what breaks if we remove them ?

For the purpose of unit testing, it is best to avoid canonicalize unless you are actually testing the effect of mixing canonicalizations with other things (that have also been tested independently).

nicolasvasilache (approved):

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
ingomueller-net: Fixed in latest commit.

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
ingomueller-net: Fixed in latest commit.

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
ingomueller-net: Fixed in latest commit.

experimental/iterators/lib/Dialect/Iterators/Transforms/DecomposeIteratorStates.cpp
ingomueller-net: Fixed in latest commit.

experimental/iterators/test/Integration/Dialect/Iterators/CPU/constant-stream.mlir
ingomueller-net: OK, makes sense. Fixed in the latest commit.

ingomueller-net (commented):
@nicolasvasilache: Thanks for the review. I addressed all issues and will land as soon as the tests pass.

PR #652, opened 2023-01-31
[Iterators] Replace utility builders for LLVM ops by upstreamed builders (NFC).

ingomueller-net:
This PR currently includes #650 (but doesn't depend on it in principle).

We have implemented some time ago utility builder functions for LLVM's ExtractValueOp and InsertValueOp with the plan to upstream them as builders eventually. In the meantime, somebody else has done that, so we can now use the upstreamed builders and remove our utility functions.


ingomueller-net:
Closing in favor of #653, which includes this PR.

PR #653, opened 2023-01-31
[Iterators] Replace utility builders for LLVM+SCF ops by upstreamed builders (NFC).

ingomueller-net:
This PR includes #650, which pulls in an upstream fix for the builder of `scf.while`.

We have implemented some time ago utility builder functions for LLVM's ExtractValueOp and InsertValueOp as well as SCF's WhileOp with the plan to upstream them as builders eventually. In the meantime, somebody else has done that, so we can now use the upstreamed builders and remove our utility functions.


ingomueller-net:
CI still fails due to an upstream bug. It should work once [my fix](https://reviews.llvm.org/D142952) has landed and I updated the LLVM dependency of this repository.

nicolasvasilache (approved):

PR #655, opened 2023-02-08
[Iterators] Add utils for 1:N type conversion

ingomueller-net:
- [ ] Add documentation to functions and classes.
- [ ] Add test cases for SCF ops.
- [ ] Add test cases for tuple op conversions.
- [ ] Add test cases for combined conversion.


experimental/iterators/lib/Utils/OneToNTypeConversionSCF.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include "iterators/Utils/OneToNTypeConversionSCF.h"
+
+#include "iterators/Utils/OneToNTypeConversion.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+
+using namespace mlir;
+using namespace mlir::iterators;
+using namespace mlir::scf;
+
+class DecomposeStatesInSCFIfOp : public OneToNOpConversionPattern<IfOp> {
```
ingomueller-net, line 24: Rename.

ingomueller-net: Done.

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+      if (!maybeResults)
+        return failure();
+      materializedResults = maybeResults.value();
+    } else if (areOperandTypesLegal && !areResultsTypesLegal &&
+               resultTypes.size() == 1) {
+      // This is a source or an argument materialization.
+      std::optional<Value> maybeResult;
+      if (llvm::all_of(operands, [&](Value v) { return v.getDefiningOp(); })) {
+        // This is an source materialization.
+        maybeResult = typeConverter.materializeArgumentConversion(
```
ftynse, line 299: The code doesn't match the comment. Form the check, it appears that the comment is correct and the code is wrong.

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+          typeConverter.materializeTargetConversion(
+              rewriter, castOp->getLoc(), resultTypes, operands.front());
+      if (!maybeResults)
+        return failure();
+      materializedResults = maybeResults.value();
+    } else if (areOperandTypesLegal && !areResultsTypesLegal &&
+               resultTypes.size() == 1) {
+      // This is a source or an argument materialization.
+      std::optional<Value> maybeResult;
+      if (llvm::all_of(operands, [&](Value v) { return v.getDefiningOp(); })) {
```
ftynse, line 297: A slightly cheaper check is to do `isa<OpResult>(v)`, you don't actually need the defining op itself.

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+        // This is an source materialization.
+        maybeResult = typeConverter.materializeArgumentConversion(
+            rewriter, castOp->getLoc(), resultTypes.front(),
+            castOp.getOperands());
+      } else {
+        // This is an argument materialization.
```
ftynse, line 303: I'd have an assert here to catch the potential weird situation where operands are a mix of block arguments and op results because of, e.g., interplay of incorrect partial conversion foldings that folded away only a part of the IR.

ingomueller-net (commented):
Thanks a lot for the review, @ftynse! I have just pushed a couple of commits that fix the smaller comments. I'll contact you directly for the few remaining design issues.

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
ingomueller-net: I am now unsure for which version of the code you made this comment but for *source* materializations, operands may actually be block arguments and there may be a mix of those and op results. I thus removed the cast again in the latest version.

If a cast is inserted due to a type conversion of an op and the operands of this op are (forwarded) block arguments, then that cast is a *source* conversion but its arguments are still block arguments. Example:

```mlir
func.func @materializations(%arg0: i1) -> tuple<i1> {
  %0 = "test.make_tuple"(%arg0) : (i1) -> tuple<i1>
  return %0 : tuple<i1>
}
```

The conversion pattern for `make_tuple` simply forwards its N operands as N results (without any op), so we get:

```mlir
func.func @materializations(%arg0: i1) -> tuple<i1> {
  %0 = builtin.unrealized_conversion_cast %arg0 : i1 to tuple<i1> {"kind" = "source"}
  return %0 : tuple<i1>
}
```

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+  if (failed(result))
+    return failure();
+  SmallVector<Value> &replacementValues = result.value();
+
+  // If replacementValues consist of the results of the original op, assume
+  // in-place update.
+  // TODO: This isn't particularly elegant. Not sure how else to handle that
+  //       case without tracking modifications through the rewriter, which
+  //       would require a custom pattern application driver.
```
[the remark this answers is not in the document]
ingomueller-net: I have reworked this part of the API by introducing a sub-class of `PatternRewriter`, which avoids this problem.

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+  if (ValueRange{op->getResults()} == replacementValues)
+    return success();
+
+  // Cast op results back to the original types and use those.
+  SmallVector<Value> castResults = buildUnrealizedBackwardsCasts(
+      replacementValues, resultConversion, rewriter);
+  rewriter.replaceOp(op, castResults);
```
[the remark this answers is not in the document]
ingomueller-net: The new design with the sub-class of `PatternRewriter` should avoid this problem as well.

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+// applyPatternsAndFoldGreedily and then replaces all newly inserted
+// UnrealizedConversionCastOps that haven't folded away. ("Backward" casts from
+// target to source types inserted by a OneToNConversionPattern normally fold
+// away with the "forward" casts from source to target types inserted by the
+// next pattern.) To understand which casts are "newly inserted", we save a list
+// of all casts existing before the patterns are applied and assume that all
+// casts not in that list after the application are new. (This is probably not
+// correct: It might be possible that an existing cast is folded away and a new
+// cast happens to be allocated with exactly the same pointer. Dealing with that
+// possiblity is an open TODO.) Also, we do not track which inserted casts are
```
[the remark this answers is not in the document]
ingomueller-net: I have now done the reverse: I annotate all casts that my pass inserts with a particular attribute -- only those casts that have that attribute after the patterns are applied will need replacing with a materialization. This way, I can also track which kind of cast (source, argument, or target) it was to invoke the corresponding user function more reliably.

I have still left the collecting of pre-existing casts as well as some properties to distinguish kinds of casts but only as asserts. I can probably remove most of that...

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+// needed for source, target, or argument materialization, so we do some
+// educated guessing to recover that information. Fixing both issues would
+// require to use a PatternRewriter that overloads various `notify*` functions
+// and similar and tracks all changes there. However, that would require a
+// dedicated pattern application driver, which is currently also left as an open
```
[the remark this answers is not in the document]
ingomueller-net: I have initially thought so, but I realized that `RewriterBase::replaceOp` does not work in the presence of 1:N type conversions. I thus think that I need a subclass of `RewriterBase` anyways in order to provide an overload for that purpose. That solves the problem for automatically inserted the "backward" cast better, IMO.

I also wanted to track the inserted casts in order to replace them with materializations. For the moment, I think that attributes are an elegant solution to that. The super clean way to do this is probably to use pass-specific rewriter class (for the whole pass, not a short-lived rewriter for each pattern application) that tracks all inserted casts like the upstream dialect conversion does. However, that seems like a lot of boilerplate code for a relatively small improvement, so would only like to do it once I get more confidence that the code is on a way to get upstreamed.

experimental/iterators/test/lib/Conversion/OneToNTypeConversion/OneToNTypeConversion.cpp
```diff
@@ -0,0 +1,322 @@
...
+LogicalResult applyOneToNConversion(Operation *op,
+                                    OneToNTypeConverter &typeConverter,
+                                    const FrozenRewritePatternSet &patterns) {
+  // Remember existing unrealized casts.
+  SmallSet<UnrealizedConversionCastOp, 4> existingCasts;
+  op->walk(
+      [&](UnrealizedConversionCastOp castOp) { existingCasts.insert(castOp); });
+
+  // Apply provided conversion patterns.
+  if (failed(applyPatternsAndFoldGreedily(op, patterns)))
```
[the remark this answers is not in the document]
ingomueller-net: Yes, that's probably true. Again, I don't want to invest time to write my own pattern application driver just yet.

ingomueller-net (commented):
@ftynse: I just pushed four more commits that address most of the remaining issues with some more profound changes in the API. Let me know what you think.

ingomueller-net:
Closing in favor of (pending) upstream changes ([here](https://reviews.llvm.org/D144469) and [here](https://reviews.llvm.org/D146959)).

PR #659, opened 2023-02-17
[Iterators] Add lookupOrCreateFn with optional isVarArg flag.

ingomueller-net:
This PR is in conflict with #653, which removes the `MLIRSupport` file.


ingomueller-net:
Abandoning in favor of landed upstream patch: https://reviews.llvm.org/D144256.

PR #666, opened 2023-03-09
[Iterators] Extend the set of types that are supported for printing.

ingomueller-net:
This PR depends on and currently includes #665.

(Off-topic: Oooohh... evil PR number 😈)

### DIFF
--- a/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsTypes.td
+++ b/experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsTypes.td
@@ -40,20 +40,11 @@ class Iterators_Type<string name, string typeMnemonic, list<Trait> traits = []>
 def Iterators_PrintableFloatType :
   FloatOfWidths<[16, 32, 64]>;
 
-/// List of bit-widths of printable integer types.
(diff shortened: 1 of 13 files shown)


experimental/iterators/include/iterators/Dialect/Iterators/IR/IteratorsTypes.td
```diff
@@ -36,29 +36,73 @@ class Iterators_Type<string name, string typeMnemonic, list<Trait> traits = []>
 //     well.
 //===----------------------------------------------------------------------===//
 
+/// List of printable float types.
+def Iterators_PrintableFloatType :
+  FloatOfWidths<[16, 32, 64]>;
+
+/// List of bit-widths of printable integer types.
+def Iterators_PrintableIntegerBitWidths {
+  list<int> widths = [ 1,  2,  3,  4,  5,  6,  7,  8,  9, 10,
```
nicolasvasilache, line 45: there should be a way to get a confined CPred in tablgen

ingomueller-net: Yes, good idea. Fixed in new commit.

PR #673, opened 2023-03-16
[Iterators] Add prefix and suffix attributes to PrintOp.

ingomueller-net:
~~This PR depends on and currently includes #666.~~


experimental/iterators/test/Integration/Dialect/Iterators/CPU/print.mlir
```diff
@@ -7,7 +7,45 @@ func.func @print_empty_tuple(%tuple : tuple<>) -> () {
   return
 }
 
-func.func @main() {
+func.func @main() {  %i32 = arith.constant 42
```
ingomueller-net, line 10: New line

experimental/iterators/lib/Dialect/Iterators/IR/Iterators.cpp
```diff
@@ -61,6 +61,111 @@ void IteratorsDialect::initialize() {
 // Iterators operations
 //===----------------------------------------------------------------------===//
 
+static ParseResult
+parsePrintOpArgs(OpAsmParser &parser, StringAttr &prefixAttr,
```
nicolasvasilache, line 65: Can we just drop all this C++ tech debt ?

The default assembly language should have what you need to avoid this, even if you need to massage your syntax a bit around commas, it's better than writing and maintaining all this.

nicolasvasilache (changes requested):

experimental/iterators/lib/Dialect/Iterators/IR/Iterators.cpp
ingomueller-net: **Objective**: The op has an optional string attribute `prefix`, an optional operand `element`, and an optional string attribute `suffix`. I want a format with three arguments (a) in that order, (b) ideally separated with commas, (c) elided if not present/not different from their default respective value (`""` for `prefix` and `\0A` for `suffix`), and (d) identified by their position (i.e., not as named arguments).

There are two distinct challenges with implementing this:

1. If the prefix has the default value (`""`) and is therefore elided, there is no SSA operand, but there is a non-default suffix, then the result of printing might be ambiguous and the *suffix* value might be parsed as *prefix*. Example:
   ```mlir
   iterators.print("" "suffix\0A")
   // gets parsed into
   "iterators.print"() {prefix = "", suffix = "suffix\0A"} : () -> ()
   // which may get printed as
   iterators.print("suffix\0A")
   // which gets parsed into
   "iterators.print"() {prefix = "suffix\0A", suffix = "\0A"} : () -> ()
   ```
   This can be prevented by exceptionally printing the prefix when it is ambiguous (i.e., iff there is no operand but a suffix). However, `custom<X>($prefix)` does not have this information, `custom<X>($prefix, ref($element), ref($suffix))` is invalid, and `custom<X>($prefix, $element, $suffix)` is what I am doing currently.
1. If I want commas (or any other separator), I need an "interleave comma" logic. Such a logic doesn't exist as a built-in directive. Also, nested optional groups aren't able to express this:
   ```
   (
     $prefix^ (`,` $element^ (`,` $suffix^)? ):($suffix)?
   ):(
     $element^ (`,` $suffix^)? ):($suffix)?
   )?
   ```
   `$element` and `$suffix` are bound several times, which the assembly format does not allow (even though the different occurrences are mutually exclusive).

I can solve the second challenge by giving up (b) and only separating the arguments by spaces. (I get random spaces in the printed result but I guess that's acceptable.) However, this leaves the first challenge unsolved.

The only way to solve the first challenge (without changing the op) I can see is to give up (d) and introduce keywords, but then the format gets verbose:
```
iterators.print(pre "prefix" %arg0 suf "suffix\0A") : i64
iterators.print(pre "prefix" suf "suffix\0A")
iterators.print(pre "prefix")
iterators.print( suf "suffix\0A")  // notice the spurious space introduced by the printer
iterators.print()
```

Finally, I can change the op itself and drop the prefix (or the suffix). Then, the ambiguous situation cannot occur. (I'd also need to separate by spaces instead of commas.) This still allows to:

```mlir
iterators.print(%i32 "suffix\0A") : i64
iterators.print(%i32) : i64
iterators.print("suffix\0A")
iterators.print()
```

The only disadvantage is that for printing only a string, you *have* to add the new line character (`\0A`) manually if you want one rather than it being the default value of the suffix.

ingomueller-net: OK, I think I found a good compromise: I simplified the op such that it either takes one `StringAttr` or an SSA operand. This has no ambiguity problem, no comma problem, and is concise in the common cases. By default, it emits an implicit new line, but in the *uncommon* case where this is not wanted, it can be suppressed with a `nonl` `UnitAttr`. (Also, printing a value *and* a suffix and/or prefix string now requires several ops, but since this is not very common either, I guess it is also acceptable.) The only (very minor) imperfection are now spurious space characters in the output but I guess we can live with that 😉

nicolasvasilache (approved):

PR #687, opened 2023-04-14
Indexing dialect

makslevental:


python/indexing/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+# and add it as an include directory in order to extend/play with the bindings
+get_target_property(mlir_python_sources MLIRPythonExtension.Core INTERFACE_SOURCES)
+list(GET mlir_python_sources 0 HEADER)
+get_filename_component(header_dir ${HEADER} DIRECTORY)
+string(REPLACE "$<BUILD_INTERFACE:" "" header_dir_clean "${header_dir}")
+set(header_dir_interface "${header_dir}>")
+
+execute_process(COMMAND python ${CMAKE_CURRENT_SOURCE_DIR}/regex_mlir_bindings.py ${header_dir_clean}/IRModule.h)
+execute_process(COMMAND python ${CMAKE_CURRENT_SOURCE_DIR}/regex_mlir_bindings.py ${header_dir_clean}/IRCore.cpp)
+execute_process(COMMAND python ${CMAKE_CURRENT_SOURCE_DIR}/regex_mlir_bindings.py ${header_dir_clean}/IRAffine.cpp)
```
makslevental, line 17: The reason for all this wackiness is because you want to use the template classes `PyConcreteType`, `PyConcreteValue`, `PyConcreteAttribute`, and etc. in [IRCore.cpp](https://github.com/llvm/llvm-project/blob/0ae342f45bedd29e34690de087011a9da4db6a65/mlir/lib/Bindings/Python/IRCore.cpp#L1974). Right now that can't be done for two reasons:

1. The symbols aren't exported in C++ (i.e., needs `PYBIND11_EXPORT __attribute__((visibility("default")))` and so they're not visible at runtime (when you load `_mlir.cpython.so`);
2. The C++ classes are bound (using pybind) with `py::module_local()` annotation which prevents them from being subclassed in Python (that annotation has the effect of hiding the constructor in Python);
3. The headers in `mlir/lib/Bindings/Python` are not in an `include/` directory.

Overall very annoying and cumbersome to hackaround, but the hacks work - I've been running them for [months now](https://github.com/makslevental/nelli/actions). Like I mentioned in the PR description, longterm (I'm hoping) they'll go away.

python/indexing/mlir_indexing/dialects/indexing.py
```diff
@@ -0,0 +1,63 @@
...
+  sys.setdlopenflags(old_flags | ctypes.RTLD_GLOBAL)
+  yield
+  sys.setdlopenflags(old_flags)
+
+
+with dl_open_guard():
+  # noinspection PyUnresolvedReferences
+  from .._mlir_libs import _mlir
+  from ..ir import (register_attribute_builder, DenseI64ArrayAttr, Context,
+                    Location, Module, InsertionPoint)
```
makslevental, line 25: Adding the attribute `__attribute__ ((visibility("default")))` does what it says - makes the symbol visibility _default_, not _public_ (with the attribute they are "hidden"). On Linux the typeinfo (re [RTTI](https://en.cppreference.com/w/cpp/types) that pybind employs) appear in the symbol table in the ELF but they are "weak" symbols

```console
$ readelf -s -W -C _mlir.cpython-310-x86_64-linux-gnu.so | grep typeinfo | grep PyOperation
687: 000000000036c4e8    56 OBJECT  WEAK   DEFAULT   23 typeinfo for mlir::python::PyOperation
635: 0000000000260bf0    28 OBJECT  WEAK   DEFAULT   16 typeinfo name for mlir::python::PyOperation
```
or "vague" symbols:

```console
$ nm -gDC _mlir.cpython-310-x86_64-linux-gnu.so | grep typeinfo | grep PyOperation
000000000036c4e8 V typeinfo for mlir::python::PyOperation
0000000000260bf0 V typeinfo name for mlir::python::PyOperation
```

That means that means they are not available for symbol resolution/relocation of subsequently loaded shared objects ([man dlopen](https://linux.die.net/man/3/dlopen)), i.e., a C extension that's loaded after the upstream `_mlir.cpython` extension is loaded. This corresponds to a default loader flag `RTLD_LOCAL`; `RTLD_GLOBAL` makes them available.

Yes I have spent an inordinate amount of time fighting linkers (even [just this week](https://reviews.llvm.org/D148058)).

python/indexing/regex_mlir_bindings.py
```diff
@@ -0,0 +1,44 @@
...
+          r"^class ((?!PYBIND11_EXPORT).*) {$",
+          r"class PYBIND11_EXPORT \1 {",
+          f,
+          flags=re.MULTILINE,
+      )
+      res = res.replace(
+          "class DefaultingPyMlirContext",
+          "class PYBIND11_EXPORT DefaultingPyMlirContext",
+      )
+      ff.write(res)
```
makslevental, line 23: See [this comment](https://github.com/iree-org/iree-llvm-sandbox/pull/687/files#r1166204774).

python/indexing/regex_mlir_bindings.py
```diff
@@ -0,0 +1,61 @@
...
+      res = re.sub(
+          r"^class ((?!MY_PYBIND11_EXPORT).*) {$",
+          r"class MY_PYBIND11_EXPORT \1 {",
+          f,
+          flags=re.MULTILINE,
+      )
+      res = res.replace(
+          "class DefaultingPyMlirContext",
+          "class MY_PYBIND11_EXPORT DefaultingPyMlirContext",
+      )
```
makslevental, line 38: See [this comment](https://github.com/iree-org/iree-llvm-sandbox/pull/687/files#r1166204774).

python/indexing/regex_mlir_bindings.py
```diff
@@ -0,0 +1,61 @@
...
+#endif
+"""
+
+module_guard = """\
+#ifdef MY_PYBIND11_EXPORT_MODULE
+#  define MY_PYBIND11_LOCAL_MODULE
+#else
+#  define MY_PYBIND11_LOCAL_MODULE , py::module_local()
+#endif
+"""
```
makslevental, line 19: I'm not sure if this is working correctly (will test further) but the idea is to prevent Iterators from taking on the same hacks as Indexing (currently) does.

nicolasvasilache:
I'd be +1 on Ingo's proposal.
I would just use the work `structured` instead of `sandbox` at the top-level.

python/indexing/CMakeLists.txt
ftynse: These symbols are intentionally unexported. For attributes and types, we should use `mlir_attr/type_subclass` which are exported. Nobody just cared enough to update the built-in types from `PyConcrete*` to these.

For values, I think https://reviews.llvm.org/D147758 adds a similar mechanism so we won't this dance anymore. Or is there something else?

examples/indexing/fancy_indexing.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
+from numpy.random import randint
+
+# %out = tensor.gather %input[%indices] gather_along_dimension(1) :
+#   tensor<2x3x4xf32>[tensor<5x6x 1xindex>] -> tensor<5x6x2x1x4xf32>
+
+input = randint(0, 10, (2, 3, 4))
+indices = randint(0, 2, (5, 6))
+out = input[:, indices, :]
+out_ = np.moveaxis(out, (1, 2), (0, 1))
```
ingomueller-net, line 10: How about asserting the result shape here? Mainly for documentation purpose such that one can see it without executing the code...

examples/indexing/fancy_indexing.py
```diff
@@ -0,0 +1,67 @@
...
+
+for index in np.ndindex(5, 6):
+  coord = indices[index]
+  assert np.allclose(out_[index], input[:, coord, :])
```
ingomueller-net, line 14: There are no numerical issues here. Why not use [`array_equal`](https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html)?

examples/indexing/fancy_indexing.py
```diff
@@ -0,0 +1,67 @@
...
+
+# %out = tensor.gather %input[%indices] gather_along_dimensions(0, 2) :
+#   (tensor<2x3x4xf32>, tensor<5x6x 2xindex>) -> tensor<5x6x3xf32>
+
+indices = randint(0, 1, (5, 6, 2))
```
ingomueller-net, line 19: Again, I am still wrapping my head around the basics here. But `randint(0, 1, ...)` produces only zeroes... is that intended? It also seems like in the next line, the indices used for the first axis of `input` can be from `[0..1]` whereas those used for the third axis can be from `[0..3]`, so even if we change the line above to `randint(0, 2, ...)`, there still seems to be something off because that wouldn't ever read the whole input. Is that intended?

ingomueller-net (commented):
@makslevental: I started with trying to understand some basics. Here are some comments that I had in the process but I need to have another look at that and the rest of the PR...

examples/indexing/fancy_indexing.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
+from numpy.random import randint
+
+# %out = tensor.gather %input[%indices] gather_along_dimension(1) :
+#   tensor<2x3x4xf32>[tensor<5x6x 1xindex>] -> tensor<5x6x2x1x4xf32>
```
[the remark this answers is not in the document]
makslevental: Yes the RFC mostly has `x1x` i.e., "rank-preserving" semantics but also mentions "rank-reducing" semantics. All the examples in this file (which, btw, is just me wrapping my head around the semantics discussed in the RFC and in upstream `TensorOps.td`) are rank-reducing (because `numpy` just naturally corresponds). For the actual implementation (wrt to `Indexing` Python bindings) there will be a way to specify `rank_reducing=True/False`.

examples/indexing/fancy_indexing.py
makslevental: TIL about `array_equal`.

examples/indexing/fancy_indexing.py
makslevental: Ah whoops I thought it was `[low, high]` but you're right it's `[low, high)`.

>the indices used for the first axis of input can be from [0..1] whereas those used for the third axis can be from [0..3]

Yes your intuition is right on - the `indices` tensor should be exercising indexing anywhere along the 0th dim and 2nd dim (where the extent of the 2nd dim is `[0..3]`).  I just didn't want to have some verbose numpy line that created an `indices` array that was something like `5 x 6 x randint(0,2) x randint(0,4)`. But you're right I should change it so that all the numbers line up intuitively.

ingomueller-net: Thanks for clarifying/confirming! As I said, I still working on understanding the basics and I asked these questions because I these subtleties might, in fact, be me fundamentally misunderstanding what is going on... BTW, this file was hugely helpful for me! Thanks :)

python/indexing/CMakeLists.txt
makslevental: Yea I think this is valid - I had in mind possibly some other extension points but no need to frontload them.

makslevental:
Closing in favor of https://github.com/iree-org/iree-llvm-sandbox/pull/689 and https://github.com/iree-org/iree-llvm-sandbox/pull/690.

PR #689, opened 2023-04-15
[Structured] Rename Iterators -> Structured

makslevental:
Implements the alternative discussed in https://github.com/iree-org/iree-llvm-sandbox/pull/687.

Then next up is https://github.com/iree-org/iree-llvm-sandbox/pull/690.


ingomueller-net (approved):
@makslevental: LGTM! (I have the impression I hadn't been particularly consistent with the namespaces of the different dialects, so I might have to clean up some things once this gets merged...)

nicolasvasilache:
Great, thanks @makslevental !

nicolasvasilache (approved):

ingomueller-net (commented):
Actually, two minor change requests: 

1. Can you run clang-format on all files in order to re-order the include statements according to their new names? Otherwise, we'll get spurious changes in subsequent commits.
2. Can you fix the header guards (`#ifndef ITERATORS_DIALECT_...` -> `#ifndef STRUCTURED_DIALECT_`)?

PR #694, opened 2023-04-18
[Indexing] Add Python bindings for indexing dialect.

makslevental:
Finally we arrive at being ready for some of the Python acrobatics. In particular, this PR adds Python bindings for a `mlir_value_subclass(..., TensorValue,...)` (using my recently upstream small [diff](https://reviews.llvm.org/D147758)) which is then used as a wrapper that maps native Python indexing to (currently) `tensor.extract`. The near-term grand plan is to route as much of `tensor.scatter` and `tensor.gather` semantics as possible through this API.

All comments/questions/concerns welcome and appreciated.

cc @kylechard

NB: In the previous PR I added a [short notebook](https://github.com/iree-org/iree-llvm-sandbox/blob/00babefe3abedd66d8a50e8be986c69e95bdcae4/examples/advanced-indexing.ipynb) that explains numpy's advanced indexing and its relation to `tensor.scatter`/`tensor.gather`. I will keep adding to that notebook as our API matures, but since there's no real addition to the API in this PR (I'm counting `tensor.extract` as too small to discuss), there's no change to that notebook here.

Edit: 

@nicolasvasilache @ingomueller-net 

I've added some code that sketches out "tensor-like" semantics on `tensor<...>` via pybinding. One choice I think we should discuss is whether we should map such operations to `linalg` or `arith` (or both). For example, I think it's reasonable that operations on tensors of data (i.e. `tensor<...xf32>`) should be expressed in terms of `linalg` as such:

```python
    ten = idx.Tensor.empty([10, 10], f32)
    twenty = ten + ten
    one_hundred = ten * ten
```

mapping to

```mlir
  %0 = tensor.empty() : tensor<10x10xf32>
  %1 = tensor.empty() : tensor<10x10xf32>
  %2 = linalg.elemwise_binary {cast = #linalg.type_fn<cast_signed>, fun = #linalg.binary_fn<add>} ins(%0, %0 : tensor<10x10xf32>, tensor<10x10xf32>) outs(%1 : tensor<10x10xf32>) -> tensor<10x10xf32>
  %3 = tensor.empty() : tensor<10x10xf32>
  %4 = linalg.elemwise_binary {cast = #linalg.type_fn<cast_signed>, fun = #linalg.binary_fn<mul>} ins(%0, %0 : tensor<10x10xf32>, tensor<10x10xf32>) outs(%3 : tensor<10x10xf32>) -> tensor<10x10xf32>
```

On the other hand it might be useful to map "index tensors" (i.e., tensor of indices) directly to `arith` as such:

```python
    ten = idx.IndexTensor.empty([10, 10])
    twenty = ten + ten
    one_hundred = ten * ten
```

mapping to

```mlir
  %0 = tensor.empty() : tensor<10x10xindex>
  %1 = arith.addi %0, %0 : tensor<10x10xindex>
  %2 = arith.muli %0, %0 : tensor<10x10xindex>
```

My thinking there is that index arithmetic should be treated distinctly/separately from tensor-of-data operations (maybe making constant-folding simpler).

### DIFF
--- a/lib/CAPI/Dialects.cpp
+++ b/lib/CAPI/Dialects.cpp
@@ -12,19 +12,48 @@
 #include "mlir/CAPI/IR.h"
 #include "mlir/CAPI/Registration.h"
 #include "mlir/IR/Types.h"
+#include "structured/Dialect/Indexing/IR/Indexing.h"
 #include "structured/Dialect/Iterators/IR/Iterators.h"
 #include "structured/Dialect/Tabular/IR/Tabular.h"
 #include "structured/Dialect/Tuple/IR/Tuple.h"
 #include "llvm/ADT/StringRef.h"
+#include <mlir-c/BuiltinTypes.h>
+#include <mlir/CAPI/Support.h>
+
+using namespace mlir;
+using namespace mlir::indexing;
+using namespace mlir::iterators;
+using namespace mlir::tabular;
+using namespace mlir::tuple;
+
+//===----------------------------------------------------------------------===//
+// Indexing dialect and attributes
+//===----------------------------------------------------------------------===//
+
+MLIR_DEFINE_CAPI_DIALECT_REGISTRATION(Indexing, indexing, IndexingDialect)
+
+bool mlirTypeIsAIndexingCustom(MlirType type) {
+  return unwrap(type).isa<CustomType>();
+}
+
+MlirType mlirIndexingCustomTypeGet(MlirContext ctx, MlirStringRef str) {
+  return wrap(CustomType::get(unwrap(ctx), unwrap(str)));
+}
+
+bool mlirIsATensorValue(MlirValue value) {
+  return mlirTypeIsATensor(mlirValueGetType(value));
+}
+
+bool mlirIsAnArithValue(MlirValue value) {
+  MlirType type = mlirValueGetType(value);
+  return mlirTypeIsABF16(type) || mlirTypeIsAComplex(type) ||
+         mlirTypeIsAF16(type) || mlirTypeIsAF32(type) || mlirTypeIsAF64(type) ||
+         mlirTypeIsAInteger(type) || mlirTypeIsAIndex(type);
+}
 
 //===----------------------------------------------------------------------===//
 // Iterators dialect and types
 //===----------------------------------------------------------------------===//
-
-using namespace mlir;
-using namespace mlir::iterators;
-using namespace mlir::tabular;
-using namespace mlir::tuple;
 
 MLIR_DEFINE_CAPI_DIALECT_REGISTRATION(Iterators, iterators, IteratorsDialect)
 
--- a/python/StructuredDialects.cpp
+++ b/python/StructuredDialects.cpp
@@ -6,18 +6,34 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include <mlir-c/BuiltinAttributes.h>
 #include <pybind11/pybind11.h>
 
-#include <array>
-#include <initializer_list>
+#include <string>
 
-#include "mlir-c/BuiltinAttributes.h"
 #include "mlir-c/IR.h"
 #include "mlir/Bindings/Python/PybindAdaptors.h"
 #include "structured-c/Dialects.h"
+#include <mlir-c/BuiltinTypes.h>
+#include <vector>
 
 namespace py = pybind11;
 using namespace mlir::python::adaptors;
+
+static MlirStringRef toMlirStringRef(const std::string &s) {
+  return mlirStringRefCreate(s.data(), s.size());
+}
+
+namespace {
+// TODO(max): this is plucked from PybindUtils.cpp upstream - refactor upstream
+pybind11::error_already_set SetPyError(PyObject *excClass,
+                                       const llvm::Twine &message) {
+  auto messageStr = message.str();
+  PyErr_SetString(excClass, messageStr.c_str());
+  return pybind11::error_already_set();
+}
+
+} // namespace
 
 PYBIND11_MODULE(_structuredDialects, mainModule) {
   //===--------------------------------------------------------------------===//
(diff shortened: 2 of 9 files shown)


lib/CAPI/Dialects.cpp
```diff
@@ -12,20 +12,49 @@
...
+
+MlirType mlirIndexingCustomTypeGet(MlirContext ctx, MlirStringRef str) {
+  return wrap(CustomType::get(unwrap(ctx), unwrap(str)));
+}
+
+bool mlirIsATensorValue(MlirValue value) {
+  return mlirTypeIsATensor(mlirValueGetType(value));
+}
+
+bool mlirIsAnArithValue(MlirValue value) {
```
ingomueller-net, line 47: Super-tiny nit: the conventions don't seem to respect English grammar and never change "a" to "an" in symbol names (e.g., `mlirTypeIsAInteger`).

lib/CAPI/Dialects.cpp
```diff
@@ -12,20 +12,49 @@
...
+  MlirType type = mlirValueGetType(value);
+  return mlirTypeIsABF16(type) || mlirTypeIsAComplex(type) ||
+         mlirTypeIsAF16(type) || mlirTypeIsAF32(type) || mlirTypeIsAF64(type) ||
+         mlirTypeIsAInteger(type) || mlirTypeIsAIndex(type);
```
ingomueller-net, line 51: Where does this selection of types come from? In the file where these functions are defined, there is also a `f8E5M2` type... Add a comment with the rationale?

python/mlir_structured/dialects/indexing.py
```diff
@@ -0,0 +1,112 @@
...
+    dims = list(dims)
+    for i, d in enumerate(dims):
+      if isinstance(d, int):
+        dims[i] = arith_dialect.ConstantOp.create_index(d).result
+
+    return Scalar(tensor_dialect.ExtractOp(self, dims).result)
+
+  @classmethod
+  def __binary_op(cls, op: str, lhs: "Tensor", rhs: "Tensor") -> "Tensor":
+    assert op in {"Add", "Sub", "Mul"}
```
ingomueller-net, line 95: Unlike (probably) all others, this looks like a good use for `assert`.

python/mlir_structured/dialects/indexing.py
```diff
@@ -0,0 +1,112 @@
...
+    assert lhs.element_type == rhs.element_type
+    assert lhs.shape == rhs.shape
+
+    if _is_floating_point_type(lhs.element_type):
+      return cls(getattr(arith_dialect, f"{op}FOp")(lhs, rhs).result)
+    if _is_integer_type(lhs.element_type) or _is_index_type(lhs.element_type):
+      return cls(getattr(arith_dialect, f"{op}IOp")(lhs, rhs).result)
+    raise NotImplementedError("Unsupported 'add' operands: {lhs}, {rhs}")
```
ingomueller-net, line 103: The message says "'add'" but it might be used for a different op.

python/mlir_structured/runtime/util.py
```diff
@@ -0,0 +1,21 @@
+import contextlib
+from typing import Optional
+
+from mlir_structured.ir import Context, Module, InsertionPoint, Location
+
+
+@contextlib.contextmanager
+def mlir_mod_ctx(src: Optional[str] = None,
+                 context: Context = None,
+                 location: Location = None):
```
ingomueller-net, line 10: Nice! That looks useful -- thanks!

I am wondering whether this isn't a testing util (and the file should, thus, live under `test/`)...?

test/python/dialects/indexing/dialect.py
```diff
@@ -0,0 +1,114 @@
...
+  i32 = IntegerType.get_signless(32)
+  with mlir_mod_ctx() as module:
+
+    @func.FuncOp.from_py_func()
+    def test_tensor_value():
+      ten = idx.Tensor.empty((10, 10), i32)
+      # CHECK: Tensor(%[[TEN:.*]], tensor<10x10xi32>)
+      print(ten)
+
+      twenty = ten + ten
```
ingomueller-net, line 73: These variable names are slightly misleading... What is there in `ten + ten` that matches "twenty"?

python/mlir_structured/dialects/indexing.py
```diff
@@ -0,0 +1,112 @@
...
+from .._mlir_libs import _mlirStructuredPasses as _cextStructuredPasses
+from .._mlir_libs._structuredDialects.indexing import *
+from ..ir import Value, Type, RankedTensorType, ShapedType, Operation, IndexType
+
+res_val_reg = re.compile(r"(%\w+) =")
+
+_body_builder = _BodyBuilder({}, {}, {})
+
+
+class Scalar(ArithValue):
```
ingomueller-net, line 26: This and the next class are (amazingly elegant!) syntactic sugar for creating MLIR. Please document the high-level idea of doing so in a paragraph or two, probably at the beginning of this file.

ingomueller-net (commented):
@makslevental: Thanks a lot for the PR and sorry for letting you wait for my review! It's OK to go and merge after some time -- whenever I have time, I try to do a post-factum review later.

So I added a few comments through-out the PR. You can address them in several small PRs or one or two big PRs that refer to this one.

python/StructuredDialects.cpp
```diff
@@ -6,19 +6,35 @@
...
+}
+
+namespace {
+// TODO(max): this is plucked from PybindUtils.cpp upstream - refactor upstream
+pybind11::error_already_set SetPyError(PyObject *excClass,
+                                       const llvm::Twine &message) {
+  auto messageStr = message.str();
+  PyErr_SetString(excClass, messageStr.c_str());
+  return pybind11::error_already_set();
+}
```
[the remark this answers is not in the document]
makslevental: This function isn't used (it was during experimentation) but it couldn't be pulled out of PybindUtils.cpp because PybindUtils.cpp is not exported/exposed because the [directory it lives in (as well as its header)](https://github.com/llvm/llvm-project/tree/main/mlir/lib/Bindings/Python) technically isn't "includable" (you can play [tricks](https://github.com/makslevental/nelli/blob/2dc46b3655b5aa7f4ba4816cc094c7b7ee1bb644/cpp_ext/CMakeLists.txt#L70)). I've discussed with @ftynse moving various things out of that directory but we just haven't come to a consensus yet on what and where.

lib/CAPI/Dialects.cpp
makslevental: My 4th grade english teacher will be very disappointed in me...

lib/CAPI/Dialects.cpp
makslevental: This is better labeled/called `ScalarValue`.

test/python/dialects/indexing/dialect.py
makslevental: play on words :) - if `ten` is short for `tensor` then two such `tens` could be a `twenty`? Anyway I'll rename.

python/StructuredDialects.cpp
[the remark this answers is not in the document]
ingomueller-net: Ah, OK, got it. I knew this kind of problem and I *thought* I had checked whether it was the case here but apparently came to the wrong conclusion...

PR #696, opened 2023-04-26
[Indexing] Add `ConcatenateOp`

makslevental:
This is a wholesale lift/grab from [stablehlo](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#concatenate). Feel free to tell me you want more tests, or more examples in the td description, or to make a suggestion about a different asm.

Stack (bottom to top):

https://github.com/iree-org/iree-llvm-sandbox/pull/702

[here](https://github.com/iree-org/iree-llvm-sandbox/pull/696)

### DIFF
--- a/include/structured/Dialect/Indexing/IR/Indexing.h
+++ b/include/structured/Dialect/Indexing/IR/Indexing.h
@@ -17,10 +17,6 @@
 
 #include "structured/Dialect/Indexing/IR/IndexingOpsDialect.h.inc"
 
-namespace mlir {
-namespace indexing {} // namespace indexing
-} // namespace mlir
-
 #define GET_TYPEDEF_CLASSES
 
 #include "structured/Dialect/Indexing/IR/IndexingOpsTypes.h.inc"
--- a/include/structured/Dialect/Indexing/IR/IndexingOps.td
+++ b/include/structured/Dialect/Indexing/IR/IndexingOps.td
@@ -37,4 +37,32 @@ def Indexing_GatherOp : Indexing_Op<"gather", [
   }];
 }
 
+def Indexing_ConcatenateOp : Indexing_Op<"concatenate",
+    [Pure, SameOperandsAndResultElementType,
+     DeclareOpInterfaceMethods<InferTypeOpInterface>]> {
+  let summary = "Concatenate operation";
+  let description = [{
+    Concatenates a variadic number of tensors in `inputs` along `dimension`
+    dimension in the same order as the given arguments and produces a `result`
+    tensor.
+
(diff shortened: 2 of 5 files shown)


test/python/dialects/indexing/dialect.py
```diff
@@ -91,7 +91,7 @@ def test_tensor_value():
   print(module)
 
   pm = PassManager.parse(
-      'builtin.module(func.func(convert-elementwise-to-linalg))')
+      "builtin.module(func.func(convert-elementwise-to-linalg))")
```
ingomueller-net, line 94: Why this change?

include/structured/Dialect/Indexing/IR/IndexingOps.td
```diff
@@ -37,4 +37,33 @@ def Indexing_GatherOp : Indexing_Op<"gather", [
...
+
+  let arguments = (ins
+    Variadic<AnyRankedTensor>:$inputs,
+    I64Attr:$dimension
+  );
+
+  let results = (outs AnyRankedTensor);
+
+  let assemblyFormat = [{
+     custom<VariadicOperandWithAttribute>($inputs) `dim` `=` $dimension attr-dict
```
ingomueller-net, line 64: Maybe you can change the assembly format slightly in order to avoid the custom directive. Maybe like this:

```suggestion
     `(` $inputs `)` `dim` `=` $dimension attr-dict
```

(Maybe add braces around `dim = ...`.)

include/structured/Dialect/Indexing/IR/Indexing.h
```diff
@@ -18,7 +18,16 @@
 #include "structured/Dialect/Indexing/IR/IndexingOpsDialect.h.inc"
 
 namespace mlir {
-namespace indexing {} // namespace indexing
```
ingomueller-net, line 21: I *think* that you shouldn't need these forward declarations (even if you needed the custom directive, see below). If it doesn't compile without these, you might miss an include to a generated header file...

ingomueller-net (approved):
LGTM!

test/python/dialects/indexing/dialect.py
makslevental: either yapf or black did this. i'll undo it

include/structured/Dialect/Indexing/IR/IndexingOps.td
makslevental: i was trying to figure out how to get it to be something like this but didn't think of delimiting with parens (not much experience yet with the parser. thanks!

PR #702, opened 2023-05-04
[Indexing] Rename gather_dims

makslevental:
Stack (bottom to top):

[here](https://github.com/iree-org/iree-llvm-sandbox/pull/702)

Note this is kind of frippery because this op is just a 1-1 duplication of @nicolasvasilache's upstream `tensor.gather`, so in reality soon (if I don't think of any reason not to) I'll just upstream the only addition/delta I've made so far (`inferReturnTypes` as a trait) and get rid of this op entirely.


ingomueller-net (approved):
LGTM!

PR #703, opened 2023-05-04
[Indexing] Refactor `mlir_value_subclass`es

makslevental:
This PR refactors adds the `mlir_value_subclass`es `TensorValue` and `ScalarValue` to 

1. Support wrapping arbitrary Python objects, in particular objects corresponding to existing IR (`OpView, Operation, Value`) and Python values that can be mapped to corresponding MLIR types (`int, float, bool, np.ndarray`).
2. Perform constant folding; i.e., arithmetic operations on `TensorValue`s and `ScalarValue`s that are the results of `arith.constant` will be evaluated at the time of execution of the Python script.
3. Carry (cached) convenience methods for querying metadata (such as shape and dtype).

A battery of tests is also added to exercise this new functionality.

### DIFF
--- a/python/mlir_structured/dialects/indexing.py
+++ b/python/mlir_structured/dialects/indexing.py
@@ -3,116 +3,369 @@
 # Licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-import re
-from typing import Union, Tuple
+import operator
+from copy import deepcopy
+from functools import cached_property, lru_cache, partialmethod
+from typing import Optional, Tuple, Union
 
-from . import arith as arith_dialect
-from . import tensor as tensor_dialect
-# noinspection PyUnresolvedReferences
+import numpy as np
+
+from . import arith, tensor
+from ._arith_ops_ext import _is_integer_like_type
 from ._indexing_ops_gen import *
-from .linalg.opdsl.lang.emitter import _BodyBuilder, _is_floating_point_type, _is_integer_type, \
-  _is_index_type, _is_complex_type
-# noinspection PyUnresolvedReferences
-from .._mlir_libs import _mlirStructuredPasses as _cextStructuredPasses
+from ._ods_common import get_op_result_or_value
+from .linalg.opdsl.lang.emitter import _is_floating_point_type
 from .._mlir_libs._structuredDialects.indexing import *
-from ..ir import Value, Type, RankedTensorType, ShapedType, Operation, IndexType
+from ..ir import (
+    DenseElementsAttr,
+    F16Type,
+    F32Type,
+    F64Type,
+    IndexType,
+    IntegerType,
+    OpView,
+    Operation,
+    RankedTensorType,
+    ShapedType,
+    Type,
+    Value,
+)
 
-res_val_reg = re.compile(r"(%\w+) =")
 
-_body_builder = _BodyBuilder({}, {}, {})
+def infer_mlir_type(
+    py_val: Union[int, float, bool, np.ndarray]
+) -> Union[IntegerType, F64Type, RankedTensorType]:
+  """Infer MLIR type (`ir.Type`) from supported python values.
 
-# The classes in this file implement syntactic sugar for creating MLIR corresponding to arithmetic operations
-# on `tensor`s (MLIR Value whose type is `tensor<...>`) and `scalar`s (`f*` floats, `i*` integers, and `index`).
-#
-# The implementation uses the Python operator overloading mechanism to overload various "dunder" methods (`__add__`, `__mul__`, etc.)
-# to dispatch to the corresponding MLIR builders (`arith.AddFOp`, `arith.MulFOp`, etc.) through their Python bindings.
-# The classes also provide some convenience factory methods and properties for accessing metadata (data type, shape, etc.).
+  Note ints and floats are mapped to 64-bit types.
+
+  Args:
+    py_val: Python value that's either a numerical value or numpy array.
+
+  Returns:
+    MLIR type corresponding to py_val.
+  """
+  if isinstance(py_val, bool):
+    return IntegerType.get_signless(1)
+  elif isinstance(py_val, int):
+    return IntegerType.get_signless(64)
+  elif isinstance(py_val, float):
+    return F64Type.get()
+  elif isinstance(py_val, np.ndarray):
+    dtype = {
+        np.int8: IntegerType.get_signless(8),
+        np.int16: IntegerType.get_signless(16),
+        np.int32: IntegerType.get_signless(32),
+        np.int64: IntegerType.get_signless(64),
+        np.uintp: IndexType.get(),
+        np.float16: F16Type.get(),
+        np.float32: F32Type.get(),
+        np.float64: F64Type.get(),
+    }[py_val.dtype.type]
+    return RankedTensorType.get(py_val.shape, dtype)
+  else:
+    raise NotImplementedError(
+        f"Unsupported Python value {py_val=} with type {type(py_val)}")
 
 
-class Scalar(ScalarValue):
+def constant(
+    value: Union[int, float, bool, np.ndarray],
+    type: Optional[Type] = None,
+    index: Optional[bool] = None,
+) -> arith.ConstantOp:
+  """Instantiate arith.constant with value `value`.
 
+  Args:
+    value: Python value that determines the value attribute of the
+      arith.constant op.
+    type: Optional MLIR type that type of the value attribute of the
+      arith.constant op; if omitted the type of the value attribute
+      will be inferred from the value.
+    index: Whether the MLIR type should be an index type; if passed the
+      type argument will be ignored.
+
+  Returns:
+    ir.OpView instance that corresponds to instantiated arith.constant op.
+  """
+  if index is not None and index:
+    type = IndexType.get()
+  if type is None:
+    type = infer_mlir_type(value)
+  assert type is not None
+
+  if isinstance(value, np.ndarray):
+    value = DenseElementsAttr.get(
+        value,
+        type=type,
+    )
+
+  return arith.ConstantOp(type, value)
+
+
+class ArithValueMeta(type(Value)):
+  """Metaclass that orchestrates the Python object protocol
+  (i.e., calling __new__ and __init__) for Indexing dialect extension values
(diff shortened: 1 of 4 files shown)


python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+        type=type,
+    )
+
+  cst = arith.ConstantOp(type, cst)
+
+  return cst
+
+
+# TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
+class ArithValueMeta(type(Value)):
```
makslevental, line 97: Using a metaclass enables wrapping whatever we want by calling the constructors of `Tensor`, `Scalar` (which are subclasses again of `TensorValue`, `ScalarValue`). So this accomplishes deftly wrapping and subclassing `Value`.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+    else:
+      raise NotImplementedError(
+          f"{cls.__name__} doesn't support wrapping {arg}.")
+
+    cls_obj = cls.__new__(cls, val)
+    cls.__init__(cls_obj, val, arg_copy)
+    return cls_obj
+
+
+class ArithValue(metaclass=ArithValueMeta):
```
makslevental, line 122: Mixin class that collects together common functionality and connects to the metaclass.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+    # The call chain after this is the __init__ here *and then* Value.__init__
+    return super().__new__(cls, val)
+
+  def __init__(self,
+               val,
+               arg: Optional[Union[int, float, bool, np.ndarray]] = None):
+    self.__arg = arg
+    super().__init__(val)
+
+  @lru_cache(maxsize=1)
```
makslevental, line 134: `@cached_property` and `@lru_cache(maxsize=1)` are effectively equivalent.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+  def is_constant(self) -> bool:
+    return isinstance(self.owner.opview, arith.ConstantOp)
+
+  @lru_cache(maxsize=1)
   def __str__(self):
     s = res_val_reg.findall(super().__str__())
```
makslevental, line 140: This can be simplified when https://reviews.llvm.org/D149902 lands.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+    return str(self)
+
+  def __binary_op(self, other: "Tensor", op: str) -> "Tensor":
+    assert op in {"add", "sub", "mul"}
+    if self.type != other.type:
+      raise ValueError(f"{self=} {other=} must have the same type.")
+
+    if self.is_constant() and other.is_constant():
+      lhs, rhs = self.literal_value, other.literal_value
+      op = operator.attrgetter(op)(operator)
```
makslevental, line 160: This is critical - effectively "constant folding at runtime" - for index arithmetic.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+  __add__ = partialmethod(__binary_op, op="add")
+  __sub__ = partialmethod(__binary_op, op="sub")
+  __mul__ = partialmethod(__binary_op, op="mul")
 
-  def __sub__(self, other) -> "Scalar":
-    return Scalar(_body_builder._binary_sub(self, other))
 
-  def __mul__(self, other) -> "Scalar":
-    return Scalar(_body_builder._binary_mul(self, other))
+class Scalar(ArithValue, ScalarValue):
```
makslevental, line 177: Order here matters - determines [method resolution order](https://gist.github.com/igormolybog/4487d8bf236005fc0ebefa9160c91ba8) - e.g., you want the `__new__`, `__init__`, and `__str__` from `ArithValue` not `ScalarValue`.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+    arg = args[0]
+    if isinstance(arg, (OpView, Operation, Value)):
+      if isinstance(arg, (Operation, OpView)):
+        assert len(arg.results) == 1
+      val = get_op_result_or_value(arg)
+      arg_copy = None
+    elif isinstance(arg, (int, float, bool, np.ndarray)):
+      dtype = kwargs.get("dtype")
+      index = kwargs.get("index")
+      arg_copy = deepcopy(arg)
```
makslevental, line 111: In particular this is for `np.ndarray`, i.e., decouple from the numpy value since `DenseElementsAttr.get(arr)` takes a [view](https://github.com/llvm/llvm-project/blob/040a41a852933d3d1b855aebc8b054baa60f61e2/mlir/lib/Bindings/Python/IRAttributes.cpp#L556) rather than making a copy.

python/mlir_structured/runtime/util.py
```diff
@@ -1,13 +1,18 @@
 import contextlib
 from typing import Optional
 
-from mlir_structured.ir import Context, Module, InsertionPoint, Location
+from mlir_structured.ir import (
+    Context,
+    Module,
+    InsertionPoint,
+    Location,
```
ingomueller-net, line 8: Sorted, please.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+        np.float16: F16Type.get(),
+        np.float32: F32Type.get(),
+        np.float64: F64Type.get(),
+    }[py_val.dtype.type]
+  else:
+    raise NotImplementedError(
+        f"Unsupported Python value {py_val=} with type {type(py_val)}")
+
+
+def constant(
```
ingomueller-net, line 72: Can you add a brief comment with what this and the previous function do?

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+        cst,
+        type=type,
+    )
+
+  cst = arith.ConstantOp(type, cst)
+
+  return cst
+
+
+# TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```
ingomueller-net, line 96: What does this comment mean?

python/mlir_structured/dialects/indexing.py
ingomueller-net: Same again: please document in the code.

python/mlir_structured/dialects/indexing.py
ingomueller-net: Same again: please document in the code.

python/mlir_structured/dialects/indexing.py
```diff
@@ -29,90 +46,199 @@
...
+class ArithValueMeta(type(Value)):
+
+  def __call__(cls, *args, **kwargs):
```
ingomueller-net, line 99: I don't understand what this function does.

Also, and this might be my limited understanding of metaclasses, why do you implement the `__call__` function? The examples of metaclasses I see [here](https://stackoverflow.com/a/6581949/651937) implement the `__new__` function...

ingomueller-net (approved):
LGTM (I think)!

This seems to make writing IR yet more convenient, which is nice! I think it is quite similar to Triton's [`constexpr`](https://github.com/openai/triton/blob/deb2c71/python/triton/language/core.py#L361) mechanism.

What I would like to ask you for the future is to make this kind of PRs smaller. This one does a couple of things, that could be independent PRs: introduce the "run time/compile time constants", improve the subclassing of the `Value` class, and several smaller clean-ups. All but the first, if isolated, can probably be reviewed in a very short amount of time, but they obstruct the review of the complicated first PR. (Some reviewers refuse to review such commits and just ask for a split before doing anything...)

nicolasvasilache (commented):
Can you improve the commit messages to explain what this pr does ?

nicolasvasilache (commented):
Can you improve the commit messages to explain what this pr does ?

python/mlir_structured/dialects/indexing.py
makslevental: > Also, why does this inherit from `type(Value)` and not from `type` (as the examples [here](https://stackoverflow.com/a/6581949/651937) do).

The comment just above (that transcribes the `TypeError`) alludes to why but I'll explain in detail.

python/mlir_structured/dialects/indexing.py
```diff
@@ -3,116 +3,369 @@
...
+      # see mlir/lib/Bindings/Python/IRAttributes.cpp#L556
+      arg_copy = deepcopy(arg)
+      val = constant(arg, dtype, index).result
+    else:
+      raise NotImplementedError(
+          f"{cls.__name__} doesn't support wrapping {arg}.")
+
+    # The mlir_value_subclass mechanism works through __new__
+    # (see mlir/Bindings/Python/PybindAdaptors.h#L502)
+    # So we have to pass the wrappee Value to the __new__ of the subclass
```
ingomueller-net, line 188:
```suggestion
    # So we have to pass the wrapped Value to the __new__ of the subclass
```

ingomueller-net (approved):
LGTM! Awesome -- thanks a lot for the update! I think the level of detail of the documentation is just perfect now :)

A small feedback on the process (and I have to confess that I don't know whether this is accepted best practice or not, so take it as personal preference for now): I like to be able to review the diff between the current version and the version from my last review, such that I can concentrate on the changes that happened since then and skip over the unchanged parts. The only way I know how Github supports that well is to leave the previously reviewed commit intact and add new commits rather than force-pushing an updated commit.

makslevental:
> A small feedback on the process (and I have to confess that I don't know whether this is accepted best practice or not, so take it as personal preference for now): I like to be able to review the diff between the current version and the version from my last review, such that I can concentrate on the changes that happened since then and skip over the unchanged parts. The only way I know how Github supports that well is to leave the previously reviewed commit intact and add new commits rather than force-pushing an updated commit.

I think that's reasonable - I've gotten into the habbit of force pushing because other projects have complained about exactly the opposite (not being able to see all the changes at once in order to get a holistic idea). But I think GitHub UI has been improving on this; e.g., you can view a subrange of commits on a branch/PR:
<p align="center"> 
image
</p>

so I'll do as you suggest from now on.

PR #706, opened 2023-05-10
[Indexing] Advanced indexing with pointwise gather

makslevental:
This PR implements the simplest case of advanced indexing: pointwise and slicing `gather`, i.e., instances like `ten[0, 0, ...]`,`ten[0, :, :, 0]`, etc. The approach I took here was to crib/emulate the implementation of `index_to_gather` in [jax](https://github.com/google/jax/blob/main/jax/_src/numpy/lax_numpy.py#L4199). 

Note, [originally the PR was much larger](https://github.com/iree-org/iree-llvm-sandbox/blob/0a8499e9237b273fa951238dce3fb6ba50491b9a/python/mlir_structured/dialects/indexing.py) but this is my best effort at splitting it up into minimal chunks[^1]. Thus, many of the functions here will be extended to handle more complex cases in the follow-on PRs.

Note also that currently this is still stacked on top https://github.com/iree-org/iree-llvm-sandbox/pull/703 because I haven't merged that one yet (waiting on re-review from you guys).

[^1]: In fact I used a coverage tool to be certain the included functionality is all exercised.

### DIFF
--- a/python/mlir_structured/dialects/indexing.py
+++ b/python/mlir_structured/dialects/indexing.py
@@ -219,6 +219,10 @@ class ArithValue(metaclass=ArithValueMeta):
     super().__init__(val)
 
   @lru_cache(maxsize=1)
+  def fold(self) -> bool:
+    return self._fold
+
+  @lru_cache(maxsize=1)
   def is_constant(self) -> bool:
     return isinstance(self.owner.opview, arith.ConstantOp)
 
@@ -253,7 +257,7 @@ class ArithValue(metaclass=ArithValueMeta):
     if self.type != other.type:
       raise ValueError(f"{self=} {other=} must have the same type.")
 
-    if self.is_constant() and other.is_constant() and self._fold:
+    if self.is_constant() and other.is_constant() and self.fold():
       # if both operands are constants (results of an arith.constant op)
       # then both have a literal value (i.e. Python value).
       lhs, rhs = self.literal_value, other.literal_value
--- a/python/mlir_structured/dialects/indexing.py
+++ b/python/mlir_structured/dialects/indexing.py
@@ -6,7 +6,7 @@
 import operator
 from copy import deepcopy
 from functools import cached_property, lru_cache, partialmethod
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, NamedTuple, Sequence, Any, List
 
 import numpy as np
 
@@ -21,7 +21,9 @@ from ..ir import (
     F16Type,
     F32Type,
     F64Type,
+    FloatAttr,
     IndexType,
+    IntegerAttr,
     IntegerType,
     OpView,
     Operation,
@@ -29,6 +31,7 @@ from ..ir import (
     ShapedType,
     Type,
     Value,
+    register_attribute_builder,
 )
 
 
@@ -58,6 +61,7 @@ def infer_mlir_type(
         np.int32: IntegerType.get_signless(32),
         np.int64: IntegerType.get_signless(64),
         np.uintp: IndexType.get(),
+        np.longlong: IndexType.get(),
         np.float16: F16Type.get(),
         np.float32: F32Type.get(),
         np.float64: F64Type.get(),
@@ -219,12 +223,12 @@ class ArithValue(metaclass=ArithValueMeta):
     super().__init__(val)
 
   @lru_cache(maxsize=1)
-  def fold(self) -> bool:
-    return self._fold
-
-  @lru_cache(maxsize=1)
   def is_constant(self) -> bool:
     return isinstance(self.owner.opview, arith.ConstantOp)
+
+  @lru_cache(maxsize=1)
+  def fold(self) -> bool:
+    return self.is_constant() and self._fold
 
   @lru_cache(maxsize=1)
   def __str__(self):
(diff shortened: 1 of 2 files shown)


test/python/dialects/indexing/dialect.py
```diff
@@ -225,16 +225,211 @@ def testConcatenateOp():
   i32 = IntegerType.get_signless(32)
   with mlir_mod_ctx() as module:
 
-    @func.FuncOp.from_py_func()
```
ingomueller-net, line 123: (Not overly important but the changes to this function could have been a different commit/PR... 😛)

python/mlir_structured/dialects/indexing.py
```diff
@@ -366,10 +370,268 @@ def __class_getitem__(
...
 
-    return Scalar(tensor.ExtractOp(self, dims))
+    if all(isinstance(d, Scalar) and d.is_constant()
+           for d in idx) and len(idx) == len(self.shape):
+      return Scalar(tensor.ExtractOp(self, idx))
+    else:
+      return build_gather(self, tuple(idx))
+
+
+#################
```
ingomueller-net, line 389:
```suggestion
###################
```

python/mlir_structured/dialects/indexing.py
```diff
@@ -366,10 +370,268 @@ def __class_getitem__(
...
+# advanced indexing
+#################
```
ingomueller-net, line 391:
```suggestion
###################
```

python/mlir_structured/dialects/indexing.py
```diff
@@ -366,10 +370,268 @@ def __class_getitem__(
...
+  (produced from simple or advanced indices) and scatter/gather dims.
+
+  Args:
+    indices: The final index tensor.
+    collapsed_dims: The set of dimensions i in operand that have
+      shape[i] == 1. Must be a tuple of integers in ascending order.
+      These determine the gather or scatter dims.
+    newaxis_dims: axes created by `np.newaxis` or None. These must be inserted
+      for gathers and eliminated for scatters.
+    unique_indices: Describes whether indices known to be non-overlapping.
```
ingomueller-net, line 405:
```suggestion
    unique_indices: Describes whether indices are known to be non-overlapping.
```

python/mlir_structured/dialects/indexing.py
```diff
@@ -366,10 +370,268 @@ def __class_getitem__(
...
+    Whether any of the dimensions of `ten` are degenerate.
+  """
+  shape = ten.shape
+  if isinstance(shape,
+                (tuple, list)) and all(isinstance(s, int) for s in shape):
+    return any(s == 0 for s in shape)
+  raise NotImplementedError(shape)
+
+
+def concatenate(tens: Sequence[Tensor], dim):
```
ingomueller-net, line 431: (Not overly important but I guess that adding this function could have been its own commit/PR... 😜)

python/mlir_structured/dialects/indexing.py
```diff
@@ -366,10 +370,268 @@ def __class_getitem__(
...
+    tens: Sequence of tensors all having the same shape, except for along
+      dimension `dim`.
+    dim: Dimension to concatenate along.
+
+  Returns:
+    Tensor that wraps a value that's either the result of index.concatenate
+    or (if all tensors are constant and foldable) the result of arith.constant.
+  """
+  if len(tens) <= 1:
+    raise ValueError(f"Expected multiple {tens=}")
```
ingomueller-net, line 444: Is this a necessary/useful restriction? The unix tool `cat` is also able to "concatenate" a single file...

python/mlir_structured/dialects/indexing.py
```diff
@@ -366,10 +370,268 @@ def __class_getitem__(
...
+      GatherOp(
+          source=operand,
+          indices=indices,
+          gather_dims=gather_dims,
+          unique=unique_indices,
+      ))
+
+
+def build_gather(
+    tens,
```
ingomueller-net, line 617: Nit: why "ten*s*"? This doesn't hold *multiple* tensors, right?

python/mlir_structured/dialects/indexing.py
```diff
@@ -366,10 +370,268 @@ def __class_getitem__(
...
+):
+  # Early bail for most trivial corner case (all full slices)
+  if all(i == slice(None) for i in idx):
+    return tens
+
+  indexer = _indices_to_indexer(idx, tens.shape)
+  out = tens
+
+  # We avoid generating a gather when indexer.gather_indices.size is empty.
+  if not _is_empty(indexer.indices):
```
ingomueller-net, line 629: I think the logic would be simpler to test for `_is_empty` which `return tens` as another early exit and then handle the general case after that...

ingomueller-net (approved):
LGTM! This starts to look like something really complete :)

I pointed out a few opportunities that would have made the PR a bit smaller. Another one would be to reduce the feature set in the first shot, for example, by only supporting numeric indices, and only later also slice indices, and only later also (explicit and implicit) ellipses etc.

python/mlir_structured/dialects/indexing.py
makslevental: Turns out it's not: 

```python
concat_single_ten_first_dim = indexing.ConcatenateOp((ten,), 0).result
print(concat_single_ten_first_dim.owner)
```
```
%1 = indexing.concatenate(%0) {dim = 0} : (tensor<10x10xi32>) -> tensor<10x10xi32>
```

python/mlir_structured/dialects/indexing.py
makslevental: lol true

PR #708, opened 2023-05-10
[Triton] Add as submodule and make basic connections.

ingomueller-net:
@makslevental, @nicolasvasilache: FYI

This PR sets up Triton as a git submodule and makes some basic connections to the rest of the sandbox. For now, I make it compile through our top-level `CMakeLists.txt` and I register the dialects in the `opt` and LSP server executables. Next I want to see whether/how I can access the Python bindings/module.


ingomueller-net:
I did not 🙈 Let me check in more detail...

makslevental:
> I did not see_no_evil Let me check in more detail...

Don't worry - you did better than me haha (I couldn't figure out how to use their existing CMake).

ingomueller-net:
> PR LGTM but it seems Triton is on an older MLIR branch and I see compilation errors. We may have to use our own fork here..

I have submitted PRs ([here](https://github.com/ptillet/triton-llvm-releases/pull/10) and [here](https://github.com/openai/triton/pull/1668)) for updating Triton's dependency to LLVM in the official repo and they have just landed. I updated this PR to use the latest version of Triton, so now our repo and Triton use the exact same version of LLVM.

That should work for now and it should work OK if we don't update LLVM too frequently but I agree that we may want to reconsider this strategy in the future.

PR #711, opened 2023-05-10
[Indexing] Remove redundant `index: bool` from Scalar, Tensor ctors

makslevental:
Small refactor of `ArithMetaValue`

PR #712, opened 2023-05-10
[Indexing] Add advanced index casting in `_canonicalize_tuple_index`

makslevental:
Paves the way for advanced indexing by canonicalizing Python lists of ints to `tensor<...xindex>` values.

Stacked on https://github.com/iree-org/iree-llvm-sandbox/pull/711

### DIFF
--- a/python/mlir_structured/dialects/indexing.py
+++ b/python/mlir_structured/dialects/indexing.py
@@ -275,7 +275,7 @@ class ArithValue(metaclass=ArithValueMeta):
         op = getattr(arith, f"{op}IOp")
       else:
         raise NotImplementedError(f"Unsupported '{op}' operands: {lhs}, {rhs}")
-    return self.__class__(op(lhs, rhs))
+    return self.__class__(op(lhs, rhs), dtype=self.dtype)
 
   # partialmethod differs from partial in that it also binds the object instance
   # to the first arg (i.e., self)
@@ -500,8 +500,43 @@ def _is_scalar(e: Any) -> bool:
   return isinstance(e, Scalar) or isinstance(e, (int, float, bool))
 
 
-def _canonicalize_tuple_index(idx, rank: int):
-  """Helper to remove Ellipsis and replace with implicit trailing slice(None)s.
+def _is_index_tensor(x):
+  """Returns True if x is a Tensor with index dtype, False otherwise."""
+  return (isinstance(x, Value) and Tensor.isinstance(x) and
+          IndexType.isinstance(x.dtype))
+
+
+def _is_int_arraylike(x):
+  """Returns True if x is array-like with integer dtype, False otherwise.
+
+  Positive (i.e., return True) examples are e.g., [[0], [1]], [[0, 1]],
+  [[[0, 1]], [[0, 1]]].
+  """
+  return (isinstance(x, int) and not isinstance(x, bool) or
+          isinstance(x, (list, tuple)) and all(_is_int_arraylike(e) for e in x))
+
+
+def _is_advanced_int_indexer(idx):
+  """Returns True if idx is array-like or slice or Ellipsis or Tensor with index
+  dtype, False otherwise.
+
+  Positive (i.e., return True) examples are e.g., [[0], [1]], [[0, 1]],
+  [[[0, 1]], [[0, 1]]], [1, 1, 1, :], [1, :, :, 1], [1, :, :, ...],
+  [idx, :, :, :], [:, idx, :, :], [:, :, idx, :].
+
+  """
+  assert isinstance(idx, tuple)
+  if all(e is None or e is Ellipsis or isinstance(e, slice) or
+         _is_scalar(e) and _has_index_type(e) for e in idx):
+    return False
+  return all(e is None or e is Ellipsis or isinstance(e, slice) or
+             _is_int_arraylike(e) or _is_index_tensor(e) for e in idx)
+
+
+def _canonicalize_tuple_index(idx: Tuple[Any], rank: int):
+  """Helper to
+  1. remove Ellipsis and replace with implicit trailing slice(None)s.
+  2. cast Python lists of lists or numpy arrays to index Tensors
 
   Args:
     rank: Rank of tensor.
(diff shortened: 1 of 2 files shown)


test/python/dialects/indexing/dialect.py
```diff
@@ -439,3 +440,37 @@ def testSimpleLiteralIndexing():
...
+@run
+def testCanonicalizeTupleIndexCastListLiteral():
+  with mlir_mod_ctx() as module:
+
+    for n_tens in range(1, 4):
+      uniqs = set()
+      n_slices = 4 - n_tens
+      ten_idx = [[0], [1]]
+      slice_idx = slice(None)
+      for p in permutations([str(ten_idx)] * n_tens +
```
ingomueller-net, line 455: A more pythonic way would be `uniqs = set(permuations(...))`. (Though you probably don't even need the `set`, do you?)

test/python/dialects/indexing/dialect.py
```diff
@@ -439,3 +440,37 @@ def testSimpleLiteralIndexing():
     print(Tensor(w.owner.operands[1]))
     # CHECK: %{{.*}} = indexing.gather %[[TEN]][%[[CST0]]] gather_dims([0, 1, 2]) unique : (tensor<10x10x10x10xi32>, tensor<3xindex>) -> tensor<10xi32>
     print(w.owner)
+
+
+# CHECK-LABEL: TEST: testCanonicalizeTupleIndexCastListLiteral
+@run
+def testCanonicalizeTupleIndexCastListLiteral():
```
ingomueller-net, line 447: I don't understand what this test is doing. Can you explain?

test/python/dialects/indexing/dialect.py
```diff
@@ -439,3 +440,37 @@ def testSimpleLiteralIndexing():
...
+      for p in permutations([str(ten_idx)] * n_tens +
+                            [str(slice_idx)] * n_slices):
+        uniqs.add(p)
+
+      for u in uniqs:
+        u = tuple(u)
+        tens_is = [i for i, t in enumerate(u) if t == str(ten_idx)]
+        slice_is = [i for i, s in enumerate(u) if s == str(slice_idx)]
+
+        tens_slices = _canonicalize_tuple_index(tuple(map(eval, u)), 4)
```
ingomueller-net, line 464: Can you try to find a way to do this without `eval`?

test/python/dialects/indexing/dialect.py
ingomueller-net: Actually, I think that tests should be so simple that they don't need explanation. Otherwise, you'd need to test the test. Even if this gets quite a bit longer, maybe it is better to spell out the cases here?

python/mlir_structured/dialects/indexing.py
```diff
@@ -503,8 +500,32 @@ def _is_scalar(e: Any) -> bool:
   return isinstance(e, Scalar) or isinstance(e, (int, float, bool))
 
 
-def _canonicalize_tuple_index(idx, rank: int):
-  """Helper to remove Ellipsis and replace with implicit trailing slice(None)s.
+def _is_index_tensor(x):
+  """Returns True if x is a Tensor with index dtype, False otherwise."""
+  return (isinstance(x, Value) and Tensor.isinstance(x) and
```
ingomueller-net, line 505: Is the first test needed? Doesn't `Tensor.isinstance` subsume that test?

python/mlir_structured/dialects/indexing.py
```diff
@@ -503,8 +500,32 @@ def _is_scalar(e: Any) -> bool:
...
+          IndexType.isinstance(x.dtype))
+
+
+def _is_int_arraylike(x):
+  """Returns True if x is array-like with integer dtype, False otherwise."""
+  return (isinstance(x, int) and not isinstance(x, bool) or
+          isinstance(x, (list, tuple)) and all(_is_int_arraylike(e) for e in x))
```
ingomueller-net, line 512: This allows nested lists. That isn't intended, is it?

ingomueller-net (commented):
I currently have trouble what the PR exactly does because I don't understand what the test does.

test/python/dialects/indexing/dialect.py
makslevental: You do need set because you want to shuffle `ten_idx` around but you want the `n_tens` copies to distinguishable.

test/python/dialects/indexing/dialect.py
makslevental: This test tests whether the canonicalization works for all permutations of non-trivial indexer and `:`, so e.g., 

```
ten[idx, :, :, :]
ten[:, idx, :, :]
ten[:, :, idx, :]
...
```

You get the idea. I did have lots of these cases spelled out before (I used this same code to generate the cases) but I think you can see that it's the same case just permuted. And, since this is a unit test for an unexposed method, and the only thing that needs to be verified is the `assert` (i.e., no IR prints and such), I changed it to this. I can change it back but it will really will blow up the length of this file.

test/python/dialects/indexing/dialect.py
makslevental: You can't - `slice(None)` and `[[0], [1]]` objects aren't hashable (but their `str` representations are) so you wouldn't be able to find unique permutations.

python/mlir_structured/dialects/indexing.py
makslevental: The first test is needed and `Tensor.isinstance` does not subsume - if you pass something that's not an `ir.Value` you will get a pybind11 exception from failing to cast [here](https://github.com/llvm/llvm-project/blob/2713781b0cdc1af647048ec97d40101664673dee/mlir/include/mlir/Bindings/Python/PybindAdaptors.h#L507).

python/mlir_structured/dialects/indexing.py
makslevental: Why not? You should be able to pass indexers that have shapes like `2x2x2` right? Just btw, in general all of these helper methods should be sensible (I assume) because Jax aims to support numpy advanced indexing as authenticall as possible.

test/python/dialects/indexing/dialect.py
ingomueller-net: Ah, right, some of the permutations will equivalent because some elements of the original list are equivalent...

nicolasvasilache (approved):

test/python/dialects/indexing/dialect.py
nicolasvasilache: This seems fine to me, I would just add comments that you generate and test all possible versions on the fly and add 1-2 examples of MLIR that gets generated.

python/mlir_structured/dialects/indexing.py
nicolasvasilache: Generally, as I look at this PR and its dependent, I would find it helpful to have more doc comments in the form of examples that also hint at the usage, similarly to how we'd write things in MLIR ODS descriptions.

Here we are just talking about supporting underlying mechanics but I already find the polymorphism a bit hard to follow. Just listing the accepted forms in the doc would be useful to me, even if it quickly turns redundant.

PR #714, opened 2023-05-10
[Indexing] Implement mapping of advanced indexing to `GatherOp`

makslevental:
This PR implements advanced indexing (`gather`) with rectangular array-like objects (Python lists of lists, and index `Tensor`s). Again, the approach closely follows the implementation of `index_to_gather` in [jax](https://github.com/google/jax/blob/main/jax/_src/numpy/lax_numpy.py#L4199).

@nicolasvasilache please take a close look at the lit tests to see whether the semantics are correct (at least the type level).

Currently stacked on top of 

https://github.com/iree-org/iree-llvm-sandbox/pull/711

https://github.com/iree-org/iree-llvm-sandbox/pull/712

### DIFF
--- a/test/python/dialects/indexing/dialect.py
+++ b/test/python/dialects/indexing/dialect.py
@@ -4,12 +4,16 @@ from random import random
 
 import numpy as np
 
-from mlir_structured._mlir_libs._mlir.ir import IndexType
+from mlir_structured._mlir_libs._mlir.ir import IndexType, F32Type
 from mlir_structured.dialects import arith, indexing
 from mlir_structured.dialects.indexing import Scalar, Tensor, IndexTensorType, _canonicalize_tuple_index
 from mlir_structured.ir import Context, IntegerType, F64Type
 from mlir_structured.passmanager import PassManager
 from mlir_structured.runtime.util import mlir_mod_ctx
+
+
+def get_array_on_one_line(a):
+  return np.array_str(a, max_line_width=np.inf).replace("\n", ",")
 
 
 def run(f):
@@ -231,7 +235,6 @@ def testTensorValue():
 def testConcatenateOp():
   i32 = IntegerType.get_signless(32)
   with mlir_mod_ctx() as module:
-
     ten = Tensor.empty((10, 10), i32)
     # CHECK: Tensor(%[[TEN:.*]], tensor<10x10xi32>)
     print(ten)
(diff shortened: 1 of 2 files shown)


ingomueller-net:
The link to JAX was super helpful! I still have trouble really understanding what these advanced indexing methods do, and I found the pointer to the great [numpy documentation](https://numpy.org/doc/1.21/reference/arrays.indexing.html#advanced-indexing) that seems to explain it. Maybe we should also link to it?

makslevental:
> The link to JAX was super helpful! I still have trouble really understanding what these advanced indexing methods do, and I found the pointer to the great [numpy documentation](https://numpy.org/doc/1.21/reference/arrays.indexing.html#advanced-indexing) that seems to explain it. Maybe we should also link to it?

That link is the first link here https://github.com/iree-org/iree-llvm-sandbox/blob/main/examples/advanced-indexing.ipynb :)

test/python/dialects/indexing/dialect.py
```diff
@@ -474,3 +477,153 @@ def testCanonicalizeTupleIndexCastListLiteral():
...
+    idx_tensor_operand = Tensor(w.owner.operands[1])
+    # CHECK: %[[IDXTEN:.*]] tensor<2x1xindex> True
+    print(idx_tensor_operand.get_name(), idx_tensor_operand.type,
+          idx_tensor_operand.is_constant())
+    # CHECK{LITERAL}: [[0], [1]]
+    print(get_array_on_one_line(idx_tensor_operand.literal_value))
+    # CHECK: %{{.*}} = indexing.gather %[[TEN]][%[[IDXTEN]]] gather_dims([0]) unique : (tensor<10x10x10x10xf32>, tensor<2x1xindex>) -> tensor<2x10x10x10xf32>
+    print(w.owner)
+
+    w = ten[[[0], [1]], [[0], [1]], :, :]
```
nicolasvasilache, line 502: Would be good to lift some of these examples to the Tensor class doc comment and be pedantic about the flavor of IR being generated.

In particular here, the non-trivial part until you see it, I find, is how the `[[0], [1]], [[0], [1]]` indices get zipped/flattened to form the indexing tensor `[[0 0], [1 1]]`.

It seems one could question the syntax / behavior until one understands that this also allows you to specify non-contiguous gather dims which require a convention to handle the non contiguous indexing "holes".

Besides the ask for more handholding in the comments, I think this looks great!

test/python/dialects/indexing/dialect.py
```diff
@@ -474,3 +477,153 @@ def testCanonicalizeTupleIndexCastListLiteral():
...
+    # CHECK: %[[IDXTEN:.*]] tensor<2x2xindex> True
+    print(idx_tensor_operand.get_name(), idx_tensor_operand.type,
+          idx_tensor_operand.is_constant())
+    # CHECK{LITERAL}: [[0 0], [1 1]]
+    print(get_array_on_one_line(idx_tensor_operand.literal_value))
+    # CHECK: %{{.*}} = indexing.gather %[[TEN]][%[[IDXTEN]]] gather_dims([0, 2]) unique : (tensor<10x10x10x10xf32>, tensor<2x2xindex>) -> tensor<2x10x10xf32>
+    print(w.owner)
+
+    # ensure that the indices aren't unique because the checks below all except non-unique
+    idx_tensor = np.random.randint(low=0, high=10, size=(5, 6, 1))
```
nicolasvasilache, line 523: nit: you could lift these 3-lines into a helper/generator function with a proper name and separate from the actual thing you are testing (the metaprogramming of the IR).

nicolasvasilache: actually, is there a specific value to using randomness here?
Can't you just pick good values and fix them?
I get 30 may be too much but you could chose e.g. 3x4 instead

test/python/dialects/indexing/dialect.py
```diff
@@ -474,3 +477,153 @@ def testCanonicalizeTupleIndexCastListLiteral():
...
+    print(idx_tensor.owner)
+    # CHECK: True
+    print(np.array(idx_tensor.literal_value).dtype == np.int64)
+
+    w = ten[idx_tensor, ...]
+    idx_tensor_operand = Tensor(w.owner.operands[1])
+    # CHECK: %[[IDXTEN:.*]] tensor<5x6x1xindex> True
+    print(idx_tensor_operand.get_name(), idx_tensor_operand.type,
+          idx_tensor_operand.is_constant())
+    # CHECK: %{{.*}} = indexing.gather %[[TEN]][%[[IDXTEN]]] gather_dims([0]) : (tensor<10x10x10x10xf32>, tensor<5x6x1xindex>) -> tensor<5x6x10x10x10xf32>
```
nicolasvasilache, line 539: now that I look at this test I am wondering whether you really want 10x10x10x10 
can you use e.g. `7x77x777x7777` instead so we disambiguate between dimensions in the type?

nicolasvasilache (approved):
Looks good to me modulo more comments and disambiguating all the 10s :)

Thanks!

test/python/dialects/indexing/dialect.py
makslevental: added a long doc string to `Tensor.__getitem__`.

test/python/dialects/indexing/dialect.py
makslevental: done

test/python/dialects/indexing/dialect.py
makslevental: very good idea

PR #718, opened 2023-05-15
Bring some code into alphabetical order. (NFC)

ingomueller-net:
This is a preparation for my PR that adds the Triton dialects, which should also be added in alphabetical order.


makslevental (approved):
LGTM

PR #730, opened 2023-05-22
[Triton] Extend conversion of tt.splat to pointer input.

ingomueller-net:
~~This PR depends on #729.~~

This PR introduces a decision that may require discussion: `tensor<!tt.ptr<T>>`are lowered to `tensor<index>` where the elements have the same integer values (i.e., the pointers are reinterpret-casted to indices).

### DIFF
--- a/lib/Conversion/TritonToLLVM/TritonToLLVM.cpp
+++ b/lib/Conversion/TritonToLLVM/TritonToLLVM.cpp
@@ -209,6 +209,14 @@ void ConvertTritonToLLVMPass::runOnOperation() {
         typeConverter.convertType(type.getPointeeType()),
         type.getAddressSpace());
   });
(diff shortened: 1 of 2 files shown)


nicolasvasilache (approved):

lib/Conversion/TritonToLLVM/TritonToLLVM.cpp
```diff
@@ -225,6 +241,13 @@ void ConvertTritonToLLVMPass::runOnOperation() {
         typeConverter.convertType(type.getPointeeType()),
         type.getAddressSpace());
   });
+  typeConverter.addConversion([&](RankedTensorType type) -> Type {
```
ingomueller-net, line 244: Add a comment that this should be only one out of several modes. Eventually, we want to handle pointers into one allocation as a base pointer with a tensor of offsets.

PR #732, opened 2023-05-22
[Triton] Extend conversion of tt.load and tt.store to 1D tensor inputs.

ingomueller-net:
This PR depends on ~~#727~~, #731, #750, ~~#746~~, and their dependencies, ~~as well as [D150952](https://reviews.llvm.org/D150952)~~. ~~It also depends on #746 but does not include it yet, and CI is expected to fail while it doesn't.~~

TODO:

- [x] Decide when to lower `func` ops to LLVM (resolved by #750).
   Until now, we converted `tt.func` and friends to `func.func` and those into `llvm.func` in one go (by adding patterns for both steps to the same pass). However, this doesn't allow to use `-convert-linalg-to-loops`, which only works on `func.func` and which this PR uses for the first time. I guess that this indicates, that we shouldn't lower everything to LLVM, but changing that probably requires redesigning the whole pipeline (or actually *designing* one rather than just getting a first thing to work).

### DIFF
--- a/lib/Conversion/TritonToLLVM/TritonToLLVM.cpp
+++ b/lib/Conversion/TritonToLLVM/TritonToLLVM.cpp
@@ -122,6 +122,10 @@ struct LoadOpConversion : public OpConversionPattern<triton::LoadOp> {
     }
 
     // Tensor of pointers.
+    // TODO(ingomueller): This is a manual tiling by one. That is fine in order
+    //     to get things running but drops a lot of information. Eventually, we
+    //     want to map this to a vectorized load/gather in order to distribute
+    //     the loading over SIMT threads.
     if (auto tensorType = ptrType.dyn_cast<RankedTensorType>()) {
       if (!tensorType.hasStaticShape())
         return rewriter.notifyMatchFailure(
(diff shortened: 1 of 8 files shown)


lib/Conversion/TritonToLLVM/TritonToLLVM.cpp
```diff
@@ -209,6 +263,53 @@ struct StoreOpConversion : public OpConversionPattern<triton::StoreOp> {
...
+      auto llvmPtrType = typeConverter->convertType(elementPtrType);
+
+      // Compute bounds of for loop.
+      Value lb = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+      Value ub = rewriter.create<arith::ConstantIndexOp>(
+          loc, tensorType.getDimSize(0));
+      Value step = rewriter.create<arith::ConstantIndexOp>(loc, 1);
+
+      // Store each tensor element at a time.
+      rewriter.create<scf::ForOp>(
```
nicolasvasilache, line 289: Please add a TODO that any extension here should go through an reimplementation as a "tiling by 1"-transformation.
We want to avoid any urge to reimplement paralellism, distribution etc in an adhoc fashion.
This is fine for now.

lib/Conversion/TritonToLLVM/TritonToLLVM.cpp
```diff
@@ -108,16 +110,67 @@ struct LoadOpConversion : public OpConversionPattern<triton::LoadOp> {
...
+      // Compute bounds of for loop.
+      Value lb = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+      Value ub = rewriter.create<arith::ConstantIndexOp>(
+          loc, tensorType.getDimSize(0));
+      Value step = rewriter.create<arith::ConstantIndexOp>(loc, 1);
+
+      // Load each tensor element at a time.
+      Value values = rewriter.create<tensor::EmptyOp>(
+          loc, tensorType.getShape(), elementType);
+      auto forOp = rewriter.create<scf::ForOp>(
```
nicolasvasilache, line 149: Please add a TODO that any extension here should go through an reimplementation as a "tiling by 1"-transformation.
We want to avoid any urge to reimplement paralellism, distribution etc in an adhoc fashion.
This is fine for now.

nicolasvasilache (approved):

PR #745, opened 2023-06-02
Replace all occurrences of llvm::Optional with std::optional. (NFC)

ingomueller-net:
The former has been a typedef of the latter for a while but upstream LLVM removes that typedef. We'll have to do this change at the latest in order to update our LLVM dependency; this PR anticipates the change.


makslevental:
I just hit this same thing in [torch-mlir when trying to bump llvm](https://github.com/llvm/torch-mlir/pull/2176#discussion_r1215159805) :)

PR #773, opened 2023-06-21
[WIP] Register all extensions in main functions.

ingomueller-net:


ingomueller-net:
This isn't a regression (yet) -- it's necessary only for updating LLVM, which I do in #778.

PR #781, opened 2023-06-26
[WIP][Triton] Add test case for ptr<ptr> operand for atomicrmw conversion.

ingomueller-net:
If [D153747](https://reviews.llvm.org/D153747) gets accepted, this should be part of #777.


ingomueller-net:
Now part of #777.

PR #790, opened 2023-12-05
Initial import of Jasc.

ingomueller-net:
This PR imports Jasc, an MLIR-based compiler for JAX kernels, to the repo. Most functionality works, but some doesn't, and some shortcuts have been taken. In particular:

* The GPU tests fail, and I haven't investigated why beyond the fact that it can't run on my machine without GPU. Maybe we need a better mechanism to choose the backend before we can make GPUs work in OSS.
* I have completely removed support for sparse and the related tests.
* I have removed several tools that aren't supported in OSS.
* The BUILD files are quite hacky and need to be cleaned up. ~~They currently contain remains of two different approaches I tried, one with several shared object files and one with a single one. The latter is the one that is currently used but I *think* that the former can also work; I just gave up that approach at some point due to a weird error only to find out that the same error also occurs with the other approach and is unrelated.~~ (Update: They BUILD files are still extremely hacky but they are now the closest to minimal I was able to achieve.)
* I have also not made any attempt yet to reduce the diff between this repository and our internal version but there are a few easy things we could do to make potential future syncs easier.
* ~~I have not added copyright and license information yet.~~ (Update: License information added.)

What does work, though, is to compile everything and run the tests (some of which ~~fail~~ are skipped):

    bazel build //...
    bazel test test:*


jasc/.bazelrc
```diff
@@ -0,0 +1,18 @@
+build --announce_rc
+
+build --experimental_repo_remote_exec
+build --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
+build --cxxopt=-w --host_cxxopt=-w
+build --define=grpc_no_ares=true
+build --define=tsl_link_protobuf=true
+build --define open_source_build=true
```
ftynse, line 8: Nit: let's be consistent in using `--define=` and `--define [[:space:]]`

jasc/.bazelrc
```diff
@@ -0,0 +1,18 @@
...
+
+build --define framework_shared_object=true
+build --define tsl_protobuf_header_only=true
```
ftynse, line 11: Any bit of comment as to why these are necessary is helpful.

jasc/.bazelrc
```diff
@@ -0,0 +1,18 @@
...
+build --define=use_fast_cpp_protos=true
+build --define=allow_oversize_protos=true
+
+build --copt=-DMLIR_PYTHON_PACKAGE_PREFIX=jaxlib.mlir.
+
+build -c opt
```
ftynse, line 17: Do we really want to force opt builds by default?

jasc/BUILD
```diff
@@ -0,0 +1,223 @@
+# Schedules for JAX.
+
+load("@rules_python//python:defs.bzl", "py_library")
+load("@pybind11_bazel//:build_defs.bzl", "pybind_extension")
+load("@llvm-project//mlir:build_defs.bzl", "cc_headers_only")
+
+package(
+    # default_applicable_licenses = ["//third_party/mlir_edge:license"],
```
ftynse, line 8: Let's not have unnecessary lines.

jasc/BUILD
```diff
@@ -0,0 +1,223 @@
...
+    default_visibility = ["//visibility:public"],
+)
+
+py_library(
+    name = "jasc",
+    srcs = ["jasc.py", "__init__.py"],
```
ftynse, line 14: Nit: can we run bulidifier from bazeltools to reformat this?  I suspect each file should be on a separate line.

jasc/BUILD
```diff
@@ -0,0 +1,223 @@
...
+        "@jax1//jaxlib/mlir:transform_dialect",
+    ],
+)
+
+cc_library(
+    name = "call_kernel_shared_library_deps",
+    deps = [
+        "@jax1//jaxlib/mlir/_mlir_libs:jaxlib_mlir_capi_shared_library",
+        ":mlir_lowering_shared_library",
+        # "//third_party/gpus/cuda:cuda_headers",
```
ftynse, line 62: Let's not commit commented-out lines.

jasc/BUILD
```diff
@@ -0,0 +1,223 @@
...
+    name = "call_kernel_shared_library_deps_headers",
+    src = "call_kernel_shared_library_deps",
+)
+
+cc_binary(
+    name = "libcallkernel.so",
+    linkopts = [
+        "-Wl,-soname=libcallkernel.so",
+        "-Wl,-rpath='$$ORIGIN'",
+    ],
```
ftynse, line 84: Nit: any comment on what this does and why is helpful to debug this in the future.

jasc/WORKSPACE
```diff
@@ -0,0 +1,207 @@
...
+
+rules_cc_dependencies()
+
+#
+# @llvm-project.
+#
+
+LLVM_COMMIT = "2f17c9f65e7da50a77101431ddf7f6ed7e1ea92c"
+LLVM_SHA256 = "a986740933506ebd1127c8abb64c78655a8c329798f37fd466a8e0f7aa7a5578"
+LLVM_TARGETS = ["X86", "AArch64", "AMDGPU"]
```
ftynse, line 32: Why do we care about AMDGPU but not NVGPU?

jasc/WORKSPACE
```diff
@@ -0,0 +1,207 @@
...
+
+#
+# @jax.
+#
+
+JAX_COMMIT = "32a317f7a43440800e1e39e00ed5f2980e088ab1"
+JAX_SHA256 = "6e2147be7360a5c0672b6ba0d654cdb2ac96113b63ef457dfdc76cd50fe69ff1"
+
+http_archive(
+    name = "jax1",
```
ftynse, line 96: Why is this called "jax1" and not just "jax"?

jasc/dialect/BUILD
```diff
@@ -0,0 +1,194 @@
+# MLIR Dialect to support Jasc transformations.
+
+load("@rules_python//python:defs.bzl", "py_library")
+load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library", "gentbl_filegroup", "td_library")
+load("@llvm-project//mlir:build_defs.bzl", "cc_headers_only")
+load("@pybind11_bazel//:build_defs.bzl", "pybind_extension")
+
+package(
+    # default_applicable_licenses = ["//third_party/mlir_edge:license"],
```
ftynse, line 9: Same as above.

jasc/dialect/BUILD
```diff
@@ -0,0 +1,194 @@
...
+        "//:mlir_lowering_shared_library",
+        "//transform_ops:jasc_transform_ops_shared_library",
+    ],
+)
+
+py_library(
+    name = "python",
+    srcs = [
+        "jasc.py",
+        # "_ods_common.py",
```
ftynse, line 185: Why is this not needed?

jasc/dialect/BUILD
```diff
@@ -0,0 +1,194 @@
...
+        ":ops_py_gen",
+    ],
+    deps = [
+        ":bindings",
+        # "@jax//jaxlib/mlir:core",
+        # "@jax//jaxlib/mlir:ir",
+        # "@jax//jaxlib/mlir:pdl_dialect",
```
ftynse, line 192: Can we just delete these?

jasc/dialect/bindings.cc
```diff
@@ -0,0 +1,26 @@
...
+        registry.insert<jasc::JascDialect>();
+        jasc::registerTransformDialectExtension(registry);
+        context->appendDialectRegistry(registry);
+        context->loadDialect<jasc::JascDialect>();
+      },
+      pybind11::arg("context") = pybind11::none());
+
+  m.def("register_lowering_passes",
+        []() { jasc::registerMLIRLoweringPasses(); });
+}
```
ftynse, line 26: Nit: trailing newline plz.

jasc/dialect/dialect.td
```diff
@@ -0,0 +1,16 @@
...
+  let name = "jasc";
+  let cppNamespace = "::jasc";
+  let dependentDialects = [
+    "::mlir::transform::TransformDialect",
+    "::mlir::gpu::GPUDialect",
+    "::mlir::memref::MemRefDialect",
+  ];
+}
+
+#endif  // JASC_DIALECT_DIALECT
```
ftynse, line 16: Here and below.

jasc/external/requirements.txt
```diff
@@ -0,0 +1,16 @@
+absl-py==2.0.0
+chex==0.1.85
+iniconfig==2.0.0
+jax==0.4.20
+jaxlib==0.4.20
```
ftynse, line 5: Are we `pip install`ing these? We are also building them, which may be unwise.

jasc/gpu_lowering_passes.h
```diff
@@ -0,0 +1,43 @@
+#ifndef THIRD_PARTY_MLIR_EDGE_JASC_GPU_LOWERING_PASSES_H_
```
ftynse, line 1: Macro name looks wrong here.

jasc/mlir_lowering.h
```diff
@@ -0,0 +1,21 @@
+#ifndef THIRD_PARTY_MLIR_EDGE_JASC_MLIR_LOWERING_H_
```
ftynse, line 1: Macro name looks wrong here, and probably in all other headeers.

jasc/patches/apply.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
ftynse, line 1: Is this even needed? Bazel should be applying patches by itself.

ftynse (commented):
I suppose we want license headers in all source files.

jasc/dialect/BUILD
ingomueller-net: In the (even more) hacky version that you reviewed, the whole target wasn't needed. I figured out a way to define the bindings in a way that is more similar to the original version, where this file is required again.

jasc/dialect/BUILD
ingomueller-net: Fixed in the latest commit.

jasc/BUILD
ingomueller-net: I don't know *exactly* what this does. I have copied that pattern from JAX's BUILD files. The latest commit contains a comment with what I know and where the code comes from.

ingomueller-net (commented):
@ftynse: Thanks a lot for the helpful comments! I think I addressed them all, plus a few of the TODOs I had listed, plus some general clean-up. Please give it another look.

jasc/.bazelrc
ftynse: Enyzme-JAX is not the best source of truth, I'm afraid. It also copied these from random places. Let's at least document it was borrowed from there so we can later try and figure out why these are necessary. And let's try removing `-c opt` as it increases compilation time. I suspect this may be to force XLA to enable some optimizations in its compiler, but we are better off without it IMO.

ftynse (approved):

ftynse:
Sorry, I forgot about this change. Don't hesitate to ping me by email when I do.

ingomueller-net:
> Sorry, I forgot about this change. Don't hesitate to ping me by email when I do.

Thanks for the final round! Addressed and merged :)

No worries! I was not in hurry with this PR and had other things on my mind as well...

PR #791, opened 2024-02-15
Remove TritonToLLVM.

ingomueller-net:
It seems like we stopped this effort for good. Let's remove the code in order to make integrating LLVM easier.

I created the [`triton-to-llvm`](https://github.com/iree-org/iree-llvm-sandbox/releases/tag/triton-to-llvm) tag in case we or someone else ever wants to revive the project.


ingomueller-net:
I've been stacking PRs but not very explicit about it. I'll consider this PR approved by @jpienaar by his [comments](https://github.com/iree-org/iree-llvm-sandbox/pull/792#pullrequestreview-1883434866) in #792.

PR #792, opened 2024-02-15
[LLVM] Update submodule to latest version from Feb. 15 (e769fb86).

ingomueller-net:
This PR depends on and, therefor, includes #793 (which removes the indexing dialect).

~~The PR is still missing changes to make the indexing dialect work with the upstream changes:~~

- [x] Make indexing dialect work. @makslevental Could you take a look?

### DIFF
--- a/benchmarks/inner_product/iterators.mlir
+++ b/benchmarks/inner_product/iterators.mlir
@@ -48,7 +48,7 @@ func.func private @sum_float(%lhs : !float_type, %rhs : !float_type) -> !float_t
 // Main program.
 //
 func.func @main(%input: !tabular.tabular_view<!element_type,!element_type>,
-                %output: !llvm.ptr<!element_type>)
+                %output: !llvm.ptr)
     attributes { llvm.emit_c_interface } {
   %stream = iterators.tabular_view_to_stream %input
     to !iterators.stream<!tuple_type>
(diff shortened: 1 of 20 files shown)


makslevental:
I moved `_get_int_int_array_attr` to `_ods_common` recently and that's what broken here. I'll commit ontop and fix in a bit.

ingomueller-net:
Ah, good to know! I saw that it was removed from its old place; I hadn't seen its new place.

But there's more: the transition from Attributes to Properties requires the rewriting of some things and I didn't manage to do it for the canonicalization pattern of `ARangeOp`.

Thanks a lot!

jpienaar (approved):
I'm generally pro dropping these, always there in revision history if we need to revive it.

makslevental:
> I'm generally pro dropping these, always there in revision history if we need to revive it.

Repeating offline comment here: feel free to drop `indexing` (or whatever portions I'm codeowner of) entirely - I doubt I'll have time/bandwidth to come back to it (at least here, in this form). I'd commit on top here but this is a PR from a fork...

ingomueller-net:
OK, I rebased this on #793, which removes the indexing dialect. I kept a branch [here](https://github.com/iree-org/iree-llvm-sandbox/tree/wip-bump-llvm-indexing) with my WIP effort of updating it to the current version of LLVM in case someone ever wants to pick it up again.

PR #794, opened 2024-02-20
[Substrait] Add Substrait dialect stub.

ingomueller-net:
This PR adds an initial stub of a dialect for representing [Substrait](https://substrait.io/) plans. The initial stub essentially consists of implementing the [`Plan` protobuf message type](https://substrait.io/serialization/binary_serialization/) without support for extensions or types. This isn't particularly useful yet but puts the code structure in place.

### DIFF
--- a/include/structured/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/structured/Dialect/Substrait/IR/SubstraitOps.td
@@ -42,6 +42,10 @@ class AnyOf<list<Constraint> opTypes, string summary = ""> : Constraint<
 //===----------------------------------------------------------------------===//
 // Plan
 //===----------------------------------------------------------------------===//
+// The definitions in this section are related to the top-level `Plan` message.
+// See https://substrait.io/serialization/binary_serialization/ and
+// https://github.com/substrait-io/substrait/blob/main/proto/substrait/plan.proto.
+//===----------------------------------------------------------------------===//
 
 def PlanBodyOp : AnyOf<[
     IsOp<"::mlir::substrait::PlanRelOp">
--- a/include/structured/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/structured/Dialect/Substrait/IR/SubstraitOps.td
@@ -1,0 +1,96 @@
+//===-- SubstraitOps.td - Substrait operations definitions -*- tablegen -*-===//
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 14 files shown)


tools/structured-opt/structured-opt.cpp
```diff
@@ -41,6 +42,14 @@ static void registerIteratorDialects(DialectRegistry &registry) {
       >();
 }
 
+static void registerSubstraitDialects(DialectRegistry &registry) {
+  registry.insert<
+      // clang-format off
```
jpienaar, line 47: TBH, I don't think this is worth keeping the format.

include/structured/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -0,0 +1,96 @@
...
+    version information inline as attributes, so it also subsumes the `Version`
+    message type. The body of the op consists of the `relation`s and (once
+    implemented) the extensions and types as well as their URLs defined in the
+    plan.
+  }];
+  let arguments = (ins
+    UI32Attr:$major_number,
+    UI32Attr:$minor_number,
+    UI32Attr:$patch_number,
+    DefaultValuedAttr<StrAttr, "\"\"">:$git_hash,
```
jpienaar, line 66: What are these used for? (non blocking question)

jpienaar (approved):

tools/structured-opt/structured-opt.cpp
ingomueller-net: Done. I think it is good to have this for lists that span multiple lines to make git diffs more readable. I started this format in case the list becomes multiple lines but, admittedly, that may never happen. Let me reintroduce the format suppression if and when I introduce a line break.

include/structured/Dialect/Substrait/IR/SubstraitOps.td
ingomueller-net: These come from the [`Version`](https://github.com/substrait-io/substrait/blob/main/proto/substrait/plan.proto#L60-L82) protobuf message type. The `git_hash` field is used to express in-between and forked versions; the `producer` field is used to document which tool produced the plan and allows to work around quirks of that producer.

I realized that the link between the dialect and the spec wasn't particularly explicit. I added a few sentences and links in the latest version. Let me know if that would have answered your question (or lead you to the right place).

ingomueller-net (commented):
Good pointers/questions -- thanks!

PR #795, opened 2024-02-20
[Substrait] Implement stub of structured-translate and Substrait target.

ingomueller-net:
This PR is based on and, therefor, includes #794.

This PR adds the new tool `structured-translate` with a translation to
and from the text format of Substrait's protobuf serialization. As the
current dialect, this isn't particularly useful in its current form,
yet, but puts the basic code structure into place.

### DIFF
--- a/lib/Target/SubstraitPB/Import.cpp
+++ b/lib/Target/SubstraitPB/Import.cpp
@@ -1,0 +1,101 @@
+//===-- Import.cpp - Import protobuf to Substrait dialect -------*- C++ -*-===//
+//
+// Licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "structured/Target/SubstraitPB/Import.h"
+
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/ImplicitLocOpBuilder.h"
+#include "mlir/IR/OwningOpRef.h"
+#include "structured/Dialect/Substrait/IR/Substrait.h"
+
+#include <google/protobuf/text_format.h>
+#include <substrait/plan.pb.h>
+
+using namespace mlir;
+using namespace mlir::substrait;
+
+namespace pb = google::protobuf;
+
+namespace {
+
+// Forward declaration for the import function of the given message type.
+//
+// We need one such function for most message types that we want to import. The
+// forward declarations are necessary such all import functions are available
+// for the definitions indepedently of the order of these definitions. The
+// message type passed to the function (specified by `MESSAGE_TYPE`) may be
+// different than the one it is responsible for: often the target op type
+// (specified by `OP_TYPE`) depends on a nested field value (such as `oneof`)
+// but the import logic needs the whole context; the message that is passed in
+// is the most deeply nested message that provides the whole context.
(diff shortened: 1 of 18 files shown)


tools/structured-translate/structured-translate.cpp
```diff
@@ -0,0 +1,50 @@
+//===-- structured-translate.cpp - custom mlir-translate --------*- C++ -*-===//
```
jpienaar, line 1: Micro-nit:  "custom" doesn't say much.

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -0,0 +1,101 @@
...
+namespace substrait {
+
+OwningOpRef<ModuleOp> translateProtobufToSubstrait(llvm::StringRef input,
+                                                   MLIRContext *context) {
+  auto plan = std::make_unique<::substrait::Plan>();
+  if (!pb::TextFormat::ParseFromString(input.str(), plan.get()))
+    return {};
+
+  // XXX: Is this really the right way/place to do this?
+  context->loadDialect<SubstraitDialect>();
```
jpienaar, line 86: You probably want to do it in the translate registration yes (its not bad to do it here as you don't have the same "dependent dialect" facility that passes have)

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+static FailureOr<PlanOp> importPlan(ImplicitLocOpBuilder builder,
+                                    const ::substrait::Plan &message) {
+  const ::substrait::Version &version = message.version();
+  auto planOp = builder.create<PlanOp>(
+      version.major_number(), version.minor_number(), version.patch_number(),
+      version.git_hash(), version.producer());
+  planOp.getBody().push_back(new Block());
+
+  for (int i = 0; i < message.relations_size(); i++) {
```
jpienaar, line 51: Nit: https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop , could you use a range based for here?

jpienaar (approved):

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Thanks for the input! I looked how the other translations do it: both [LLVM](https://github.com/llvm/llvm-project/blob/6b62a9135a/mlir/lib/Target/LLVMIR/ModuleImport.cpp#L2079) and [SPIRV](https://github.com/llvm/llvm-project/blob/6b62a9135a/mlir/lib/Target/SPIRV/TranslateRegistration.cpp#L41) load their respective dialect in their "to MLIR" translate function explicitly.

Note that I already use the [corresponding hook](https://github.com/llvm/llvm-project/blob/6b62a9135a/mlir/include/mlir/Tools/mlir-translate/Translation.h#L120-L121) for *registering* the dialects (namely with `registerSubstraitDialects` in `structured-translate.cpp`).

In the latest commit, I simply dropped the comment.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Yep, range-based works!

tools/structured-translate/structured-translate.cpp
ingomueller-net: Done.

ingomueller-net (commented):
Thanks, @jpienaar, for the review! I addressed all issues and will merge as soon as CI passes.

PR #797, opened 2024-02-23
[Substrait] Implement first version of NamedTableOp.

ingomueller-net:
This op corresponds to a `Rel` message with a `ReadRel` message that, in
turn, has a `NamedTable` message. With this op, we are capable of
representing a first complete plan in our dialect that we can serialize
to protobuf and execute (by copy'n'pasting into an external consumer).


lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -33,13 +34,96 @@ void SubstraitDialect::initialize() {
...
+                                 TypeRange fieldTypes) {
+  int numConsumedNames = 0;
+  llvm::SmallSet<llvm::StringRef, 8> currentLevelNames;
+  for (Type type : fieldTypes) {
+    // Check name of current field.
+    if (numConsumedNames >= static_cast<int>(fieldNames.size()))
+      return emitError(loc, "not enough field names provided");
+    auto currentName = llvm::cast<StringAttr>(fieldNames[numConsumedNames]);
+    if (!currentLevelNames.insert(currentName).second)
+      return emitError(loc, llvm::Twine("duplicte field name: '") +
```
jpienaar, line 72: duplicate (should this then be an DictionaryAttr instead?)

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+FailureOr<std::unique_ptr<::substrait::Plan>> exportOperation(ModuleOp op) {
+  if (!op->getAttrs().empty())
+    return failure();
+
+  if (op->getRegions().size() != 1)
+    return failure();
+
+  Region &body = op.getBodyRegion();
+  if (llvm::range_size(body.getOps()) != 1)
```
jpienaar, line 68: This is a little wasteful/one could early exit without computing total size.

jpienaar (commented):
(just skimmed before next talk ... I should probably see how I could view stacked PRs better :) )

ingomueller-net:
> (just skimmed before next talk ... I should probably see how I could view stacked PRs better :) )

Thanks! In the "Files changed" tab, where you're probably doing the review, you can change "Show changes from **all commits**" to a single commit or even a range (by holding shift).

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: How? The only way I can think of is something like `++body.getOps().begin() != body.op_end()` but that's much less readable, isn't it?

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: Typo fixed -- thanks!

Not directly: this models a nested named struct where the unnamed types are given in a nested form and the names in a flat form. Now, the names only need to be unique *per nesting level*, so I can't simply model this type with a dictionary.

I have played with the thought to introduce a type for named structs where each nested level would be a dictionary indeed. But I think the low-tech solution, which doesn't need a custom type, is preferable for now.

jpienaar: SG (you can do a dictionaryattr with entries that are dictionaryattr without introducing anything new, but I'm also fine starting here and evaluating)

lib/Target/SubstraitPB/Export.cpp
jpienaar: Indeed it is, one could also if the first is a terminator, I think your version from comment would be cheaper than that though. One could get creative with zip and seq too , but I think simplest may be a lambda with loop inside from readability point of view (I mean this may not matter, I meant "little" in the non-British use of little :-))

jpienaar (approved):
Nice, thanks

include/structured/Dialect/Substrait/IR/SubstraitInterfaces.td
```diff
@@ -0,0 +1,24 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef SUBSTRAIT_DIALECT_SUBSTRAIT_IR_SUBSTRAITINTERFACES
+#define SUBSTRAIT_DIALECT_SUBSTRAIT_IR_SUBSTRAITINTERFACES
+
+include "mlir/IR/OpBase.td"
+
+def Substrait_RelOpInterface : OpInterface<"RelOpInterface"> {
```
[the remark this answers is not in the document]
ingomueller-net: I haven't made up my mind for this file yet. For `SubstraitOps.td`, I have sections that correspond to the sections in the spec and the protobuf files of Substrait, and then sort alphabetically inside each section. I guess I'll do the same here (once I figure out if we need sections and which they are).

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -50,6 +71,51 @@ FailureOr<std::unique_ptr<Plan>> exportOperation(ModuleOp op) {
...
+  // Build `RelCommon` message.
+  auto relCommon = std::make_unique<RelCommon>();
+  auto direct = std::make_unique<RelCommon::Direct>();
+  relCommon->set_allocated_direct(direct.release());
+
+  // Build `Struct` message.
+  auto struct_ = std::make_unique<::substrait::Type::Struct>();
+  struct_->set_nullability(
+      Type_Nullability::Type_Nullability_NULLABILITY_REQUIRED);
+  auto tupleType = llvm::cast<TupleType>(op->getResult(0).getType());
```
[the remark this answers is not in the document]
ingomueller-net: I am now using `op.getResult()`, which is the accessor for the one result called `result`. That's the case for every Substrait operator: there is always exactly one result. (This goes even further: in databases, a plan node/operator is always thought of as producing exactly one result -- the resulting relation/table.)

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -63,11 +129,29 @@ FailureOr<std::unique_ptr<Plan>> exportOperation(PlanOp op) {
...
   plan->set_allocated_version(version.release());
 
-  // TODO(ingomueller): build plan content once defined.
+  // Add `relation`s to plan.
+  for (auto relOp : op.getOps<PlanRelOp>()) {
+    Operation *terminator = relOp.getBody().front().getTerminator();
+    auto rootOp = llvm::dyn_cast<RelOpInterface>(
+        terminator->getOperand(0).getDefiningOp());
+
+    FailureOr<std::unique_ptr<Rel>> rel = exportOperation(rootOp);
```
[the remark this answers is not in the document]
ingomueller-net: Good catch! I am using a `cast` now because the containing op verifies that all ops in its regions are either a terminator or `PlanRelOp`s.

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -38,8 +40,68 @@ namespace {
...
+static mlir::FailureOr<NamedTableOp>
+importNamedTable(ImplicitLocOpBuilder builder, const Rel &message) {
+  const ReadRel &readRel = message.read();
+  const ReadRel::NamedTable &namedTable = readRel.named_table();
+  MLIRContext *context = builder.getContext();
+
+  // Assemble table name.
+  llvm::SmallVector<FlatSymbolRefAttr> tableNameRefs;
+  tableNameRefs.reserve(namedTable.names_size());
+  for (int i = 0; i < namedTable.names_size(); i++) {
```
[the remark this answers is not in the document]
ingomueller-net: Fixed! I could have caught myself that after the last PR...

test/Dialect/Substrait/named-table-invalid.mlir
```diff
@@ -0,0 +1,33 @@
+// RUN: structured-opt -verify-diagnostics -split-input-file %s
+
+substrait.plan version 0 : 42 : 1 {
```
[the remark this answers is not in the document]
ingomueller-net: Done!

ingomueller-net (commented):
All in-line comments addressed with the latest commits. I'll now look again at the top-level discussion.

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: I have thought about the `DictionaryAttr` idea a bit: I think it would be possible to store all types and their names in a dict attr, indeed, but that would be an *attribute*. I still also need the type, which would be redundant. I could of course verify that the redundancy is at least consistent, or even remove it from the assembly with some custom parser/printer but all that sounds more complicated that I'd like this interim solution to be...

ingomueller-net: I did add a TODO about a `NamedStruct` type, though.

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: Shame! [`std::ranges::take_view`](https://en.cppreference.com/w/cpp/ranges/take_view) would be helpful here but it's C++20.

But, actually, maybe `std::next(body.op_begin()) != body.op_end()` isn't too bad?

ingomueller-net: Yet another thought: if the IR is valid, then the range of the size is 1, so `llvm::range_size` acutally has acceptable cost. Do we really have to optimize for runtime on invalid IR?

ingomueller-net: Based on that last argument, I propose to revert my latest change and go with the initial version. WDYT?

ingomueller-net: Resolving this discussion because (1) the related line wasn't part of this commit, (2) it's the only one left for an otherwise approved PR, (3) we can always come back later: #815.

PR #799, opened 2024-03-06
[Substrait] Add Python bindings.

ingomueller-net:


python/StructuredDialects.cpp
```diff
@@ -60,6 +60,26 @@ PYBIND11_MODULE(_structuredDialects, mainModule) {
...
+  //
+  // Dialect
+  //
+
+  substraitModule.def(
+      "register_dialect",
+      [](MlirContext context, bool doLoad) {
+        MlirDialectHandle handle = mlirGetDialectHandle__substrait__();
+        mlirDialectHandleRegisterDialect(handle, context);
+        if (doLoad) {
```
jpienaar, line 77: Micro nit: MLIR style is to elide braces for single line conditional bodies.

python/mlir_structured/dialects/SubstraitOps.td
```diff
@@ -0,0 +1,12 @@
+// Copyright 2021 The IREE Authors
```
jpienaar, line 1: 2024 ?

jpienaar (approved):

python/mlir_structured/dialects/SubstraitOps.td
ingomueller-net: Fixed by #806.

PR #801, opened 2024-03-07
[Substrait] Add 'using namespace ::substrait' in import/export. (NFC)

ingomueller-net:
This is in line with the LLVM code style and makes the code more concise.


ingomueller-net:
Trivial change -- self-approving.

PR #802, opened 2024-03-07
[Substrait] Emit diagnostics on import/export.

ingomueller-net:
~~This PR is based on and, therefor, includes #801.~~

### DIFF
--- a/lib/Target/SubstraitPB/Export.cpp
+++ b/lib/Target/SubstraitPB/Export.cpp
@@ -38,20 +38,20 @@ DECLARE_EXPORT_FUNC(PlanOp, Plan)
 
 FailureOr<std::unique_ptr<Plan>> exportOperation(ModuleOp op) {
   if (!op->getAttrs().empty()) {
-    op->emitOpError("has attributes.");
+    op->emitOpError("has attributes");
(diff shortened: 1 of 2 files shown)


lib/Target/SubstraitPB/Export.cpp
```diff
@@ -37,16 +37,21 @@ DECLARE_EXPORT_FUNC(ModuleOp, Plan)
 DECLARE_EXPORT_FUNC(PlanOp, Plan)
 
 FailureOr<std::unique_ptr<Plan>> exportOperation(ModuleOp op) {
-  if (!op->getAttrs().empty())
+  if (!op->getAttrs().empty()) {
+    op->emitOpError("has attributes.");
```
jpienaar, line 41: Nit: no trailing punctuation with emit errors.

jpienaar: (here and below)

jpienaar (approved):

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: Fixed.

PR #803, opened 2024-03-07
[Substrait] Add test that executes exported plan in pyarrow.

ingomueller-net:
~~This PR depends on and, therefor, includes #797.~~


jpienaar (approved):
Nice!

PR #804, opened 2024-03-07
[Substrait] Improve Python bindings of PlanOp and PlanRelOp.

ingomueller-net:
~~This PR depends on and, therefor, contains #797.~~

### DIFF
--- a/test/python/dialects/substrait/dialect.py
+++ b/test/python/dialects/substrait/dialect.py
@@ -15,15 +15,26 @@ def run(f):
 # CHECK-LABEL: TEST: testSubstraitDialect
 @run
 def testSubstraitDialect():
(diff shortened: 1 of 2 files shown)


test/python/dialects/substrait/dialect.py
```diff
@@ -18,3 +18,23 @@ def testSubstraitDialect():
   plan = ss.PlanOp(0, 42, 1)
   print(plan)
   # CHECK: substrait.plan
+
+
+# CHECK-LABEL: TEST: testNamedTable
+@run
+def testNamedTable():
+  plan = ss.PlanOp(0, 42, 1)
```
jpienaar, line 26: mmm, could we do named args here? (`(0, 42, 1)` doesn't say too much)

jpienaar (approved):

PR #809, opened 2024-03-18
[Substrait] Fix error messages in protobuf import.

ingomueller-net:


ingomueller-net:
Trivial change, merging without review...

PR #810, opened 2024-03-18
[Substrait] Implement first version of CrossOp.

ingomueller-net:
~~This PR depends on and, therefor, includes #797 and #808~~.

This corresponds to the `CrossRel` message and the ones contained in it.

### DIFF
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -67,7 +67,7 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   llvm::append_range(fieldTypes, rightFieldTypes);
   auto resultType = TupleType::get(context, fieldTypes);
 
-  inferredReturnTypes.emplace_back(resultType);
+  inferredReturnTypes = SmallVector<Type>{resultType};
 
   return success();
 }
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -59,8 +59,8 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   Value leftInput = operands[0];
   Value rightInput = operands[0];
 
-  TypeRange leftFieldTypes = leftInput.getType().cast<TupleType>().getTypes();
-  TypeRange rightFieldTypes = rightInput.getType().cast<TupleType>().getTypes();
+  TypeRange leftFieldTypes = cast<TupleType>(leftInput.getType()).getTypes();
+  TypeRange rightFieldTypes = cast<TupleType>(rightInput.getType()).getTypes();
 
   llvm::SmallVector<mlir::Type> fieldTypes;
   fieldTypes.append(leftFieldTypes.begin(), rightFieldTypes.end());
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -62,9 +62,9 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   TypeRange leftFieldTypes = cast<TupleType>(leftInput.getType()).getTypes();
   TypeRange rightFieldTypes = cast<TupleType>(rightInput.getType()).getTypes();
 
-  llvm::SmallVector<mlir::Type> fieldTypes;
-  fieldTypes.append(leftFieldTypes.begin(), rightFieldTypes.end());
-  fieldTypes.append(rightFieldTypes.begin(), rightFieldTypes.end());
+  SmallVector<mlir::Type> fieldTypes;
+  llvm::append_range(fieldTypes, leftFieldTypes);
+  llvm::append_range(fieldTypes, rightFieldTypes);
   auto resultType = TupleType::get(context, fieldTypes);
 
   inferredReturnTypes.emplace_back(resultType);
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -51,6 +51,27 @@ void SubstraitDialect::initialize() {
 namespace mlir {
 namespace substrait {
 
+LogicalResult
+CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
+                          ValueRange operands, DictionaryAttr attributes,
+                          OpaqueProperties properties, RegionRange regions,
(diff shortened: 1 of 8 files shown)


lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -51,6 +51,27 @@ void SubstraitDialect::initialize() {
...
 
+LogicalResult
+CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
+                          ValueRange operands, DictionaryAttr attributes,
+                          OpaqueProperties properties, RegionRange regions,
+                          llvm::SmallVectorImpl<Type> &inferredReturnTypes) {
+  Value leftInput = operands[0];
+  Value rightInput = operands[0];
+
+  TypeRange leftFieldTypes = leftInput.getType().cast<TupleType>().getTypes();
```
jpienaar, line 62: This is the deprecated form. Prefer `cast<TupleType>(...)` form.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -51,6 +51,27 @@ void SubstraitDialect::initialize() {
...
+  TypeRange rightFieldTypes = rightInput.getType().cast<TupleType>().getTypes();
+
+  llvm::SmallVector<mlir::Type> fieldTypes;
+  fieldTypes.append(leftFieldTypes.begin(), rightFieldTypes.end());
```
jpienaar, line 66: Should this be leftFiledTypes.end() ? (I think ADT has some helpers to work with ranges)

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -51,6 +51,27 @@ void SubstraitDialect::initialize() {
...
+  fieldTypes.append(rightFieldTypes.begin(), rightFieldTypes.end());
+  auto resultType = TupleType::get(context, fieldTypes);
+
+  inferredReturnTypes.emplace_back(resultType);
```
jpienaar, line 70: Nit: You could also do

```
inferredReturnTypes = {resultType};
```

to make more explicit that its single return.

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -51,6 +51,27 @@ void SubstraitDialect::initialize() {
...
+  Value leftInput = operands[0];
+  Value rightInput = operands[0];
+
+  TypeRange leftFieldTypes = leftInput.getType().cast<TupleType>().getTypes();
+  TypeRange rightFieldTypes = rightInput.getType().cast<TupleType>().getTypes();
+
+  llvm::SmallVector<mlir::Type> fieldTypes;
+  fieldTypes.append(leftFieldTypes.begin(), rightFieldTypes.end());
+  fieldTypes.append(rightFieldTypes.begin(), rightFieldTypes.end());
+  auto resultType = TupleType::get(context, fieldTypes);
```
jpienaar, line 68: OOC why use a tuple here vs multiple return values?

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -186,6 +208,9 @@ static mlir::FailureOr<RelOpInterface> importRel(ImplicitLocOpBuilder builder,
 
   Rel::RelTypeCase relType = message.rel_type_case();
   switch (relType) {
+  case Rel::RelTypeCase::kCross: {
+    return importCross(builder, message);
```
jpienaar, line 212: For Read you used ReadRel, but Cross has no Rel suffix.

jpienaar (approved):
Nice, would most (all?) of the ops that return values be single valued with tuple return type?

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: Done.

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: Good catch! I had only looked in the member functions of `SmallVector` and was surprised (annoyed?) that it wasn't there. I do see that there is a free function, and the newest version uses that. Still a bit surprised that it's only a free function...

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: It has to be

```cüü
inferredReturnTypes = SmallVector<Type>{resultType};
```

but I guess that's still more explicit for a single return?

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: I am not using multiple return values because the op really only returns one thing.

However, I do have taken a short-cut here (and everywhere): these ops don't return a single *tuple*, the return a single *relation*. At some point, I should define a `SubstraitRelation` type that, like `Tuple`, will have possibly nested fields.

This is a real necessity, not just for esthetics: Substrait allows a *double* nesting of arrays and tuples (e.g., an array of structs of structs of arrays of atomics) and I need constraints for which atomic values are allowed, which I can't express with `Tuple`s.

I have not done this initially because I `Tuple` provided a sufficiently close alternative and I wanted to make progress with other things first. I have a TODO in `SubstraitTypes.td`, and we can start working on that whenever we feel the time has come:

```
// TODO(ingomueller): Transform this into a proper relation type.
```

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Fixed.

ingomueller-net (commented):
Thanks, Jacques! All very good points! The last series of commits should address them.

ingomueller-net:
> Nice, would most (all?) of the ops that return values be single valued with tuple return type?

Copying from a resolved in-line conversion:

> I am not using multiple return values because the op really only returns one thing.
> 
> However, I do have taken a short-cut here (and everywhere): these ops don't return a single *tuple*, the return a single *relation*. At some point, I should define a `SubstraitRelation` type that, like `Tuple`, will have possibly nested fields.
> 
> This is a real necessity, not just for esthetics: Substrait allows a *double* nesting of arrays and tuples (e.g., an array of structs of structs of arrays of atomics) and I need constraints for which atomic values are allowed, which I can't express with `Tuple`s.
> 
> I have not done this initially because I `Tuple` provided a sufficiently close alternative and I wanted to make progress with other things first. I have a TODO in `SubstraitTypes.td`, and we can start working on that whenever we feel the time has come:
>
> ```
> // TODO(ingomueller): Transform this into a proper relation type.
> ```

PR #812, opened 2024-03-19
[Substrait] Implement first version of FilterOp and Expression ops.

ingomueller-net:
~~This PR depends on and, therefor, includes #817 and its dependencies.~~

This corresponds to the `FilterRel` message and the ones contained in
it. Since one of the messages is an `Expression` message, this PR also
introduces the `ExpressionOpInterface` as well as the `LiteralOp` as
well as the basic import/export infrastructure for such ops/messages.
The `LiteralOp` corresponds to the `Expression.Literal` message type,
and the `ExpressionOpInterface` encapsualtes common properties and
functionality of all types of `Expression` messages. Finally, the PR
also adds attribute constraints for typed attributes of the currently
supported types.

### DIFF
--- a/include/structured/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/structured/Dialect/Substrait/IR/SubstraitOps.td
@@ -138,12 +138,50 @@ def Substrait_PlanRelOp : Substrait_Op<"relation", [
 }
 
 def Substrait_YieldOp : Substrait_Op<"yield", [
-    Terminator, HasParent<"::mlir::substrait::PlanRelOp">
+    Terminator,
+    ParentOneOf<[
+      "::mlir::substrait::PlanRelOp",
+      "::mlir::substrait::FilterOp"
+    ]>
   ]> {
   let summary = "Yields the result of a `PlanRelOp`";
   let arguments = (ins AnyType:$value);
   let assemblyFormat = "$value attr-dict `:` type($value)";
   let builders = [OpBuilder<(ins), [{ /* do nothing */ }]>];
+}
+
+//===----------------------------------------------------------------------===//
+// Expressions
+//===----------------------------------------------------------------------===//
+// The definitions in this section are related to the various `Expression`
+// message types. See https://substrait.io/expressions/field_references/ and
+// https://github.com/substrait-io/substrait/blob/main/proto/substrait/algebra.proto.
+//===----------------------------------------------------------------------===//
+
+// TODO(ingomueller): Make this a proper base for expressions.
+class Substrait_ExpressionOp<string mnemonic, list<Trait> traits = []> :
+  Substrait_Op<mnemonic, traits # [
+    Substrait_ExpressionOpInterface
+  ]>;
+
+def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
+    DeclareOpInterfaceMethods<InferTypeOpInterface>
+  ]> {
+  let summary = "Literal expression";
+  let description = [{
+    Represents a `Literal` message together with all messages it contains and
+    the `Expression` message it is contained in.
+
+    Example:
+
+    ```mlir
+    %0 = literal true
+    ```
+  }];
+  // TODO(ingomueller): extend to other types.
+  let arguments = (ins Substrait_AtomicAttribute:$value);
+  let results = (outs Substrait_AtomicType:$result);
+  let assemblyFormat = "$value attr-dict";
 }
 
 //===----------------------------------------------------------------------===//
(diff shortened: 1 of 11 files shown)


include/structured/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -128,14 +128,52 @@ def Substrait_PlanRelOp : Substrait_Op<"relation", [
 }
 
 def Substrait_YieldOp : Substrait_Op<"yield", [
-    Terminator, HasParent<"::mlir::substrait::PlanRelOp">
+    Terminator,
+    ParentOneOf<[
+      "::mlir::substrait::PlanRelOp",
+      "::mlir::substrait::FilterOp"
```
jpienaar, line 134: Nit: Alphabetical ?

include/structured/Dialect/Substrait/IR/SubstraitTypes.td
```diff
@@ -29,6 +29,17 @@ def Substrait_AtomicTypes {
...
 
+/// Attributes of currently supported atomic types.
+def Substrait_AtomicAttributes {
+  list<Attr> attrs = [
+    SI1Attr, // Boolean
+    SI32Attr // I32
+  ];
+}
+
+/// Attrbitue of one of the currently supported atomic types.
```
jpienaar, line 40: Attribute

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -72,6 +72,56 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   return success();
 }
 
+LogicalResult FilterOp::verifyRegions() {
+  MLIRContext *context = getContext();
+  Type si1 = IntegerType::get(context, /*width=*/1, IntegerType::Signed);
+  Region &condition = getCondition();
+
+  // Verify that type of yielded value is Boolean.
+  auto yieldOp = llvm::cast<YieldOp>(condition.front().getTerminator());
```
jpienaar, line 81: mlir::cast ? (its same thing/alias but just more consistent with other usage)

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -72,6 +72,56 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+                            ValueRange operands, DictionaryAttr attributes,
+                            OpaqueProperties properties, RegionRange regions,
+                            llvm::SmallVectorImpl<Type> &inferredReturnTypes) {
+  auto *typedProperties = properties.as<Properties *>();
+  if (!loc)
+    loc = UnknownLoc::get(context);
+
+  auto attr = llvm::dyn_cast<TypedAttr>(typedProperties->getValue());
+  if (!attr)
+    ::emitError(loc.value()) << "unsuited attribute for literal value: "
```
jpienaar, line 116: Use emitOptionalError here (its technically a bit of a mishap that I need to fix, but the intention was that one can run inference without reporting errors by not specifying a location - I was going to remove it, but folks found it semi useful)

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -137,6 +140,78 @@ FailureOr<std::unique_ptr<Rel>> exportOperation(CrossOp op) {
...
+
+  return rel;
+}
+
+FailureOr<std::unique_ptr<Expression>> exportOperation(LiteralOp op) {
+  auto si1 = IntegerType::get(op.getContext(), 1, IntegerType::Signed);
+  auto si32 = IntegerType::get(op.getContext(), 32, IntegerType::Signed);
+
+  // Build `Literal` message.
+  auto value = llvm::cast<TypedAttr>(op.getValue());
```
jpienaar, line 200: Same

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -137,6 +140,78 @@ FailureOr<std::unique_ptr<Rel>> exportOperation(CrossOp op) {
...
+  mlir::Type literalType = value.getType();
+  auto literal = std::make_unique<Expression::Literal>();
+  if (literalType == si1 || literalType == si32)
+    literal->set_boolean(value.cast<IntegerAttr>().getSInt());
+  else
+    op->emitOpError("has unsupported value");
```
jpienaar, line 206: Wouldn't this have failed verification?

jpienaar (approved):

include/structured/Dialect/Substrait/IR/SubstraitOps.td
ingomueller-net: Done.

include/structured/Dialect/Substrait/IR/SubstraitTypes.td
ingomueller-net: Done.

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: I wasn't aware of the alias! But I don't have the feeling that is being used consistently... `llvm::cast` shows 1249 hits in upstream MLIR, `mlir::cast` shows 10 hits. (That doesn't account for `using mlir` but I guess it gives *some* indication...)

If you think we should prefer `mlir::cast`, ping me and I'll address it for the whole repo. (We have been using `llvm::cast` or whichever thing `cast` was resolved for everywhere.)

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: Good pointer! Yeah, I *do* find it useful, too! Changed!

I'd prefer if it returned an `InflightDiagnostics`, though, such that you can `<<` further information into the result of the call. I did see that you can add other `Args...`, so I can still emit stuff, but I kinda prfer the `<<` version...

I also realized that the original code did not return the error it emitted! I fixed that and added a test...

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: (See above)

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: You caught me somewhat off-guard here: I hadn't asked myself that question too much. I guess it makes sense to only test for things that do verify and I hadn't approached this consciously in the past.

In this particular instance, though, I have been lazy with the definition of the `literal` op and allowed any `TypedAttr`. Only potential users of the resulting values may then complain about getting wrongly typed operands. In short, there are unexportable cases that aren't caught by verification, so we need the test as is.

ingomueller-net (commented):
Thanks for the review! All points should be addressed, with exception of `llvm::cast` vs `mlir::cast`, which I can address in a follow-up PR project wide if you think we should do that.

PR #817, opened 2024-03-21
[Substrait] Add support for SI1 and nested tuple types.

ingomueller-net:
~~This PR depends on and, therefor, includes #810.~~

### DIFF
--- a/lib/Target/SubstraitPB/Export.cpp
+++ b/lib/Target/SubstraitPB/Export.cpp
@@ -45,20 +45,51 @@ DECLARE_EXPORT_FUNC(RelOpInterface, Rel)
 
 FailureOr<std::unique_ptr<proto::Type>> exportType(Location loc,
                                                    mlir::Type mlirType) {
+  MLIRContext *context = mlirType.getContext();
+
+  // Handle SI1.
+  auto si1 = IntegerType::get(context, 1, IntegerType::Signed);
+  if (mlirType == si1) {
+    // TODO(ingomueller): support other nullability modes.
+    auto i1Type = std::make_unique<proto::Type::Boolean>();
+    i1Type->set_nullability(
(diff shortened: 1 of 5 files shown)


lib/Target/SubstraitPB/Export.cpp
```diff
@@ -45,20 +45,51 @@ DECLARE_EXPORT_FUNC(RelOpInterface, Rel)
...
+  auto si1 = IntegerType::get(context, 1, IntegerType::Signed);
+  if (mlirType == si1) {
+    // TODO(ingomueller): support other nullability modes.
+    auto i1Type = std::make_unique<proto::Type::Boolean>();
+    i1Type->set_nullability(
+        Type_Nullability::Type_Nullability_NULLABILITY_REQUIRED);
+
+    auto type = std::make_unique<proto::Type>();
+    type->set_allocated_bool_(i1Type.release());
+    return std::move(type);
```
jpienaar, line 60: Is there a way we could avoid make_unique's here? (well more the memory allocation parts here if this done often).

jpienaar (approved):

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: You are touching upon a question I had pending in my head: what is the best practice to assemble nested protobufs. I searched both internal and external doc and couldn't find any clear answers.

It sound *plausible* that using `std::make_unique` and `set_allocated_*` entails more allocations than necessary. I *suppose* that the better alternative is to use something like:

```c++
Inner* inner = outer->mutable_inner();
inner->set_primite(...);
```

If that is the case, then I'd have to restructure the whole file, so I'd like to be sure what is the best approach before starting. Can you confirm or have a good resource on the top of your head?

ingomueller-net: Resolving conversation here in favor of #826.

PR #818, opened 2024-03-21
[Substrait] Implement first version of FieldReferenceOp.

ingomueller-net:
This PR depends on and, therefor, includes #812 and its dependencies.

### DIFF
--- a/include/structured/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/structured/Dialect/Substrait/IR/SubstraitOps.td
@@ -164,6 +164,33 @@ class Substrait_ExpressionOp<string mnemonic, list<Trait> traits = []> :
     Substrait_ExpressionOpInterface
   ]>;
 
+def Substrait_FieldReferenceOp : Substrait_ExpressionOp<"field_reference", [
+    DeclareOpInterfaceMethods<InferTypeOpInterface>
+  ]> {
+  let summary = "Field reference expression";
+  let description = [{
+    Represents a `FieldReference` message together with all messages it contains
+    and the `Expression` message it is contained in.
+
+    Example:
+
+    ```mlir
+    %0 = ...
+    %1 = field_reference %0[[0]] : tuple<si32>
+    // %1 is of type `si32`
+    ```
+  }];
+  // TODO(ingomueller): extend to other types, map access, and masked references.
+  let arguments = (ins
+    Substrait_ContainerType:$container,
+    I64ArrayAttr:$position
+  );
+  let results = (outs Substrait_AtomicType:$result);
+  let assemblyFormat = [{
+    $container `[` $position `]` attr-dict `:` type($container)
+  }];
+}
+
 def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
     DeclareOpInterfaceMethods<InferTypeOpInterface>
   ]> {
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -72,6 +72,56 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   return success();
 }
 
+/// Computes the type of the nested field of the given `type` identified by
+/// `position`. Each entry `n` in the given index array `position` corresponds
+/// to the `n`-th entry in that level. The function is thus implemented
+/// recursively, where each recursion level extracts the type of the outer-most
+/// level identified by the first index in the `position` array.
+static FailureOr<Type> computeTypeAtPosition(Location loc, Type type,
+                                             ArrayRef<Attribute> position) {
+  if (position.empty())
+    return type;
+
+  // Recurse into tuple field of first index in position array.
+  if (auto tupleType = llvm::dyn_cast<TupleType>(type)) {
+    auto indexAttr = llvm::dyn_cast<IntegerAttr>(position[0]);
+    if (!indexAttr)
+      return emitError(loc) << position[0] << " is not a valid index";
+
+    int64_t index = indexAttr.getInt();
+    ArrayRef<Type> fieldTypes = tupleType.getTypes();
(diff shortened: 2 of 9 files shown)


include/structured/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -128,14 +128,79 @@ def Substrait_PlanRelOp : Substrait_Op<"relation", [
 }
 
 def Substrait_YieldOp : Substrait_Op<"yield", [
-    Terminator, HasParent<"::mlir::substrait::PlanRelOp">
+    Terminator,
+    ParentOneOf<[
+      "::mlir::substrait::PlanRelOp",
+      "::mlir::substrait::FilterOp"
```
jpienaar, line 134: Order?

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -72,6 +72,106 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
+  if (!loc)
+    loc = UnknownLoc::get(context);
+
+  // Extract field type at given position.
+  ArrayAttr position = typedProperties->getPosition();
+  Type inputType = operands[0].getType();
+  FailureOr<Type> fieldType =
+      computeTypeAtPosition(loc.value(), inputType, position.getValue());
+  if (failed(fieldType))
+    return ::emitError(loc.value())
```
jpienaar, line 116: emitOptionalError

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -137,6 +141,115 @@ FailureOr<std::unique_ptr<Rel>> exportOperation(CrossOp op) {
...
+
+  // Build `Rel` message.
+  auto rel = std::make_unique<Rel>();
+  rel->set_allocated_filter(filterRel.release());
+
+  return rel;
+}
+
+FailureOr<std::unique_ptr<Expression>> exportOperation(LiteralOp op) {
+  auto si1 = IntegerType::get(op.getContext(), 1, IntegerType::Signed);
```
jpienaar, line 234: Given you are just using for checking, you could do a dyn\_cast\<IntegerType> and check signed and type (these 2 here would be cached, but probably cheaper check still)

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -137,6 +141,115 @@ FailureOr<std::unique_ptr<Rel>> exportOperation(CrossOp op) {
...
+  auto si32 = IntegerType::get(op.getContext(), 32, IntegerType::Signed);
+
+  // Build `Literal` message.
+  auto value = llvm::cast<TypedAttr>(op.getValue());
+  mlir::Type literalType = value.getType();
+  auto literal = std::make_unique<Expression::Literal>();
+  if (literalType == si1 || literalType == si32)
+    literal->set_boolean(value.cast<IntegerAttr>().getSInt());
```
jpienaar, line 242: So its only 0 or not that matter here?

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -137,6 +141,115 @@ FailureOr<std::unique_ptr<Rel>> exportOperation(CrossOp op) {
...
+  else
+    op->emitOpError("has unsupported value");
```
jpienaar, line 244: type ?

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -106,6 +111,130 @@ static mlir::FailureOr<CrossOp> importCrossRel(ImplicitLocOpBuilder builder,
...
+importExpression(ImplicitLocOpBuilder builder, const Expression &message) {
+  MLIRContext *context = builder.getContext();
+  Location loc = UnknownLoc::get(context);
+
+  Expression::RexTypeCase rex_type = message.rex_type_case();
+  switch (rex_type) {
+  case Expression::RexTypeCase::kSelection: {
+    return importFieldReference(builder, message.selection());
+  }
+  case Expression::RexTypeCase::kLiteral: {
```
jpienaar, line 124: Lets keep it in alphabetical order unless order matters

jpienaar (approved):

include/structured/Dialect/Substrait/IR/SubstraitOps.td
ingomueller-net: (Already raised and addressed in #812, sorry for the double review effort.)

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: (Dito: raised and addressed in #812.)

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: Good point! This is part of #812, so I addressed it there.

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: Very good catch! `set_boolean` does not make sense for `si32`. This passed unnoticed because we can't create plans that use 32-bit literals, so there was no test. This is part of #812, so I addressed everything there.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Done.

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: I think you can argue both ways: the attributed on the op is called `value` and if that attribute has an unsupported *type`, the op has an unsupported *value*, no?

ingomueller-net (commented):
Thanks for the review! Unfortunately, you seem to have re-reviewed #812, onto which this PR was stacked. You did find some new issues there, which came on time before landing that ;)

I rebased this PR now onto `main` with #812 included, so you should see the changes specific to this PR more clearly. Do you want to have another look? Otherwise, just stamp...

PR #821, opened 2024-03-27
[Substrait] Support root case in PlanRel message.

ingomueller-net:
~~This PR depends on and, therefor, includes #820 and all of its dependencies.~~

This essentially consist in adding optional field names to the
`PlanRelOp`. Since verifying if those field names match the yielded type
of the relation consists of the same logic as the one used in the
verifier of the `NamedTableOp`, this commit also factors out that
verification log. While touching that code, the commit also slightly
extends the error messages emitted on verification failure.

### DIFF
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -79,9 +79,9 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
 /// own). Furthermore, the names on each nesting level need to be unique. For
 /// details, see
 /// https://substrait.io/tutorial/sql_to_substrait/#types-and-schemas.
-FailureOr<int> verifyNamedStructHelper(Location loc,
-                                       llvm::ArrayRef<Attribute> fieldNames,
-                                       TypeRange fieldTypes) {
+static FailureOr<int>
+verifyNamedStructHelper(Location loc, llvm::ArrayRef<Attribute> fieldNames,
+                        TypeRange fieldTypes) {
   int numConsumedNames = 0;
   llvm::SmallSet<llvm::StringRef, 8> currentLevelNames;
   for (Type type : fieldTypes) {
@@ -109,9 +109,9 @@ FailureOr<int> verifyNamedStructHelper(Location loc,
   return numConsumedNames;
 }
 
-LogicalResult verifyNamedStruct(Operation *op,
-                                llvm::ArrayRef<Attribute> fieldNames,
-                                TupleType tupleType) {
+static LogicalResult verifyNamedStruct(Operation *op,
+                                       llvm::ArrayRef<Attribute> fieldNames,
+                                       TupleType tupleType) {
   Location loc = op->getLoc();
   TypeRange fieldTypes = tupleType.getTypes();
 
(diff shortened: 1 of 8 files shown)


lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -79,9 +79,9 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
 /// own). Furthermore, the names on each nesting level need to be unique. For
 /// details, see
 /// https://substrait.io/tutorial/sql_to_substrait/#types-and-schemas.
-FailureOr<int> verifyNamedStruct(Location loc,
-                                 llvm::ArrayRef<Attribute> fieldNames,
-                                 TypeRange fieldTypes) {
+FailureOr<int> verifyNamedStructHelper(Location loc,
```
jpienaar, line 82: Is this a local helper? (should this be static)

lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -142,6 +142,22 @@ LogicalResult NamedTableOp::verify() {
...
+  return verifyNamedStruct(getOperation(), fieldNames, tupleType);
+}
+
+LogicalResult PlanRelOp::verifyRegions() {
+  if (!getFieldNames().has_value())
+    return success();
+
+  llvm::ArrayRef<Attribute> fieldNames = getFieldNames()->getValue();
+  auto yieldOp = llvm::cast<YieldOp>(getBody().front().getTerminator());
+  auto tupleType = llvm::cast<TupleType>(yieldOp.getValue().getType());
```
jpienaar, line 157: And just to confirm, the return type of yield would have been verified before verifyRegions is called? (I don't recall the verifications order off the top of my head)

lib/Target/SubstraitPB/Export.cpp
```diff
@@ -226,8 +226,21 @@ FailureOr<std::unique_ptr<Plan>> exportOperation(PlanOp op) {
...
       return failure();
 
+    // Handle `Rel`/`RelRoot` cases depending on whether `names` is set.
     PlanRel *planRel = plan->add_relations();
-    planRel->set_allocated_rel(rel.value().release());
+    if (std::optional<Attribute> names = relOp.getFieldNames()) {
+      auto root = std::make_unique<RelRoot>();
+      root->set_allocated_input(rel->release());
+
+      auto namesArray = names->cast<ArrayAttr>().getAsRange<StringAttr>();
```
jpienaar, line 235: Free standing cast form? (they marked the others as deprecated so it should be emitting warnings now)

jpienaar (approved):

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: Yes. Fixed!

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: I confirm via the [docs](https://mlir.llvm.org/docs/DefiningDialects/Operations/#verification-ordering):

> Note that the second phase [i.e., the region verifiers] will be run after the operations in the region are verified.

lib/Target/SubstraitPB/Export.cpp
ingomueller-net: Thanks! I had grepped for `.cast` so I missed `->cast` 😉

ingomueller-net (commented):
Thanks! Should all be addressed by the latest commits.

PR #822, opened 2024-03-27
[Substrait] Add test that executes exported plan in DuckDB.

ingomueller-net:
~~This PR depends and, therefor, include #821 and its dependencies.~~

This illustrates that plans produced by our compiler can run in
[DuckDB](https://github.com/duckdb/duckdb), an in-process SQL DB system.

### DIFF
--- a/test/python/dialects/substrait/e2e_duckdb.py
+++ b/test/python/dialects/substrait/e2e_duckdb.py
@@ -1,0 +1,52 @@
+# RUN: %PYTHON %s | FileCheck %s
+
+import duckdb
+
+from mlir_structured.dialects import substrait as ss
+from mlir_structured import ir
+
+
+def run(f):
+  print("\nTEST:", f.__name__)
+  with ir.Context(), ir.Location.unknown():
+    ss.register_dialect()
+    f()
+  return f
+
+
+# CHECK-LABEL: TEST: testNamedTable
+@run
+def testNamedTable():
+  # Set up test table.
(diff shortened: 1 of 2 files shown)


test/python/dialects/substrait/e2e_duckdb.py
```diff
@@ -0,0 +1,52 @@
...
+
+  # Execute in duckdb and print result.
+  query_result = con.from_substrait(proto=pb_plan)
+
+  print(query_result.to_arrow_table())
+  # CHECK-NEXT:          pyarrow.Table
+  # CHECK-NEXT:          a: int32
+  # CHECK-NEXT:          b: int32
+  # CHECK-NEXT:          ----
+  # CHECK-NEXT{LITERAL}: a: [[1,2,3]]
```
jpienaar, line 51: Ha, I'd have thought this is array of structs, but it looks more like structs of array like.

jpienaar (approved):

test/python/dialects/substrait/e2e_duckdb.py
ingomueller-net: You are right: this *is* a struct of arrays (data frame) but it doesn't immediately look like one. I guess I could convert it to a pandas [`DataFrame`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) before printing, which would look more like a table, but this form is actually a bit more robust for `CHECK` testing, I think.

PR #823, opened 2024-03-27
[Substrait] Add test that executes exported plan in DataFusion.

ingomueller-net:
This PR adds an end-to-end test with [Apache DataFusion](https://github.com/apache/datafusion), an Arrow-based SQL execution engine that can consume Substrait plans. This illustrates how our Substrait plans can run on different engines and provide a base for more complete compatibility tests, which we may want to do at some point down the road.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,7 @@
 -r third_party/llvm-project/mlir/python/requirements.txt
 
 # Testing.
+datafusion==32.0.0
 lit
 
 # Plotting.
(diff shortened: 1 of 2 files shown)


jpienaar (commented):
Noob question: what is DataFusion? :) (the check result here looks the same as previous one so don't know what roles which do)

ingomueller-net:
> Noob question: what is DataFusion? :) (the check result here looks the same as previous one so don't know what roles which do)

Good question! I did what I probably should have done from the start: write the rationale in the PR description/commit message. (In short: DataFusion is another Substrait consumer, so this tests illustrates the portability of Substrait.)

PR #824, opened 2024-03-27
[Substrait] Add test that imports plan from Ibis.

ingomueller-net:
~~This PR depends on and, therefor, includes #821.~~

This illustrates that our compiler can import plans from
[Ibis](https://ibis-project.org/), a Python-based dataframe library that
can run its computations on a variety of back-ends, including Substrait.

### DIFF
--- a/test/python/dialects/substrait/e2e_ibis.py
+++ b/test/python/dialects/substrait/e2e_ibis.py
@@ -1,0 +1,36 @@
+# RUN: %PYTHON %s | FileCheck %s
+
+import ibis
+from ibis_substrait.compiler import core as ibis_ss
+
+from mlir_structured.dialects import substrait as ss
+from mlir_structured import ir
+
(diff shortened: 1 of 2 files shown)


jpienaar (approved):

test/python/dialects/substrait/e2e_ibis.py
```diff
@@ -0,0 +1,36 @@
...
+from ibis_substrait.compiler import core as ibis_ss
+
+from mlir_structured.dialects import substrait as ss
+from mlir_structured import ir
+
+
+def run(f):
+  print("\nTEST:", f.__name__)
+  with ir.Context(), ir.Location.unknown():
+    ss.register_dialect()
```
jpienaar, line 13: We could probably move this to a test setup section, but also not terrible to do repeatedly here.

PR #825, opened 2024-05-14
[Substrait] Create emit op to represent output mapping.

ingomueller-net:
This PR is based on and, therefore, includes #820 and its dependencies.

The commit introduces handling for the `emit_kind` field of the
`RelCommon` message, which is a common field of all (?) cases of the
`Rel` message. The current design models this field as a dedicated op
such that the output mapping is only ever present in that op, `emit`.

There are at least two alternatives to this IR design. The first one
consists of making the output mapping part of each of the ops the
represnet `Rel` messages and expose it through the `RelOpInterface`.
However, this would mean that (1) the custom assembly of each op would
have to represent the mapping, which is manual effort and a possible
source for inconsistencies, (2) each op would have to implement type
inference in the presence of a mapping, and (3) most rewrites of all ops
would have to take that mapping into account for their semantics. Having
the mapping in one place makes all of this simpler. The downside is that
what is kept in a single place in the Substrait protobuf format is now
spread across two ops in the MLIR representation. However, I believe
that this is the smaller of two evils and the current import and export
seems to work.

Another alternative would be to combine the two: make the mapping part
of all ops but *also* introduce a dedicated `emit` op. Then, two passes
could move the mapping from one to the other depending on which of the
two representations would be more convenient. However, this would not
get rid of Issues (1) and (2) above and lead to more concepts and code.

### DIFF
--- a/lib/Target/SubstraitPB/Export.cpp
+++ b/lib/Target/SubstraitPB/Export.cpp
@@ -39,6 +39,7 @@ namespace {
   static FailureOr<std::unique_ptr<MESSAGE_TYPE>> exportOperation(OP_TYPE op);
 
 DECLARE_EXPORT_FUNC(CrossOp, Rel)
+DECLARE_EXPORT_FUNC(EmitOp, Rel)
 DECLARE_EXPORT_FUNC(ExpressionOpInterface, Expression)
 DECLARE_EXPORT_FUNC(FieldReferenceOp, Expression)
 DECLARE_EXPORT_FUNC(FilterOp, Rel)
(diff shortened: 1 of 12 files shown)


lib/Target/SubstraitPB/Export.cpp
```diff
@@ -96,6 +155,20 @@ FailureOr<std::unique_ptr<proto::Type>> exportType(Location loc,
   return emitError(loc) << "could not export unsupported type " << mlirType;
 }
 
+/// We just forward to the overload for `RelOpInterface`, which will have to
+/// export this op. We can't (easily) do it here because the emit op is
+/// represented as part of the `RelCommon` message of one of the cases of the
+/// `Rel` message but there is no generic way to access the `common` field of
+/// the various cases.
+FailureOr<std::unique_ptr<Rel>> exportOperation(EmitOp op) {
```
ingomueller-net, line 163: I am not 100% happy with the whole structure. It is somewhat dangerous to rely on implementations doing a particular thing without being able to enforce it.

The problem is, I think, that I alluded to here: there is no proper way to the access the `common` field of different message types. Maybe it's worth to consider the following alternative: implement a `accessCommonField` function with a big switch that accesses that field for any of the cases we support. This is ugly but in concentrates the ugliness in one place. Then `exportOperation(EmitOp)` could first export the op that produces its input and then *modify* the `common` field of the resulting message of that export. I suppose something similar could work for the import.

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -47,13 +47,45 @@ namespace {
...
+DECLARE_IMPORT_FUNC(NamedTable, Rel, RelOpInterface)
 DECLARE_IMPORT_FUNC(Plan, Plan, PlanOp)
 DECLARE_IMPORT_FUNC(PlanRel, PlanRel, PlanRelOp)
 DECLARE_IMPORT_FUNC(ReadRel, Rel, RelOpInterface)
 DECLARE_IMPORT_FUNC(Rel, Rel, RelOpInterface)
 
+/// Imports the provided `RelCommon` message by producing an `EmitOp` that
+/// expresses the `Emit` message if it exists.
+///
+/// **This function must be called at the end of the import function of every
```
jpienaar, line 62: Is it possible to have some top level helper function?

say

```c++

template <typename T>
mlir::FailureOr<RelOpInterface> importRel(..., T inputOp) {
  ... importRelImpl<T>(...)
  return importMaybeEmit(....)
}
```

Not sure if it helps that much, but makes a structure where more difficult to forget it.

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -128,6 +163,66 @@ importExpression(ImplicitLocOpBuilder builder, const Expression &message) {
   }
 }
 
+static mlir::FailureOr<FieldReferenceOp>
+importFieldReference(ImplicitLocOpBuilder builder,
+                     const Expression::FieldReference &message) {
+  using ReferenceSegment = Expression::ReferenceSegment;
+
+  MLIRContext *context = builder.getContext();
+  Location loc = UnknownLoc::get(context);
```
jpienaar, line 172: Could we do better here?  even just file name

lib/Target/SubstraitPB/Import.cpp
```diff
@@ -128,6 +163,66 @@ importExpression(ImplicitLocOpBuilder builder, const Expression &message) {
...
+
+  // Emit error on unsupported cases.
+  // TODO(ingomueller): support more cases.
+  if (!message.has_direct_reference())
+    return emitError(loc) << "only direct reference supported";
+
+  // Traverse list to extract indexes.
+  llvm::SmallVector<int64_t> indexes;
```
jpienaar, line 180: indices ?

jpienaar: (i'm not sure which one is used more db side)

jpienaar (approved):

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: Yes, I think that's what I should aim for. I'll have a stab at that. The problem I'll have to solve for that is what I sketched above: I'll need to access the `common` field of the different message types but because they are different message types, there is no one function that does that. I think that the solution is to encapsulate that functionality in a function that does just that (and is implemented with one big switch).

ingomueller-net (commented):
Thanks for the review! I'll address the points shortly.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: OK, that's what I ended up doing. I think that that was an important thing to clean up -- the new version is quite a bit more elegant.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: I think yes, but this would require some code reorganization. A few notes before I forget: I'd need to implement the translation using `llvm::SourceMgr` and access `BufferBuffer:: getBufferIdentifier` to get the filename.

ingomueller-net: I've created #839 to track this effort.

lib/Target/SubstraitPB/Import.cpp
ingomueller-net: This is due to me not being a native speaker, I think. Fixed in #840.

PR #827, opened 2024-05-16
[Substrait] Add folder for EmitOp with identity mapping.

ingomueller-net:
This PR is based on and, therefor, includes #825 and its dependencies.

This folds the result of an `emit` op with identity mapping to its
input. For example, it folds the following snippet:

```mlir
%0 = named_table @t1 as ["a", "b"] : tuple<si1, si32>
%1 = emit [0, 1] from %0 : tuple<si1, si32> -> tuple<si1, si32>
// usages of %1
```

into

```mlir
%0 = named_table @t1 as ["a", "b"] : tuple<si1, si32>
// usages of %0 instead of %1
```

### DIFF
--- a/lib/Dialect/Substrait/IR/Substrait.cpp
+++ b/lib/Dialect/Substrait/IR/Substrait.cpp
@@ -72,6 +72,24 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
   return success();
 }
 
+OpFoldResult EmitOp::fold(FoldAdaptor adaptor) {
+  // Return if the mapping is not the identity mapping.
+  int64_t numFields = cast<TupleType>(getInput().getType()).size();
+  int64_t numIndices = getMapping().size();
+  if (numFields != numIndices)
+    return {};
+  for (int64_t i = 0; i < numIndices; ++i) {
+    auto attr = getMapping()[i];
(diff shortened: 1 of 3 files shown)


lib/Dialect/Substrait/IR/Substrait.cpp
```diff
@@ -72,6 +72,24 @@ CrossOp::inferReturnTypes(MLIRContext *context, std::optional<Location> loc,
...
 
+OpFoldResult EmitOp::fold(FoldAdaptor adaptor) {
+  // Return if the mapping is not the identity mapping.
+  int64_t numFields = cast<TupleType>(getInput().getType()).size();
+  int64_t numIndices = getMapping().size();
+  if (numFields != numIndices)
+    return {};
+  for (int64_t i = 0; i < numIndices; ++i) {
+    auto attr = getMapping()[i];
+    int64_t index = cast<IntegerAttr>(attr).getInt();
```
jpienaar, line 83: Makes me wonder if we should be using dense array attr ...

jpienaar (approved):

lib/Dialect/Substrait/IR/Substrait.cpp
ingomueller-net: Yes! Done in #841.

PR #831, opened 2024-05-22
[Substrait] Initial version of emit deduplication pass.

ingomueller-net:
This PR is based on and, therefor, contains #828 ~~and #829~~.

This pass removes redundant values introduced by duplicated fields in
the `mapping`s of `emit` ops. This commit puts the boilerplate into
place and implements a pattern to handle `emit` ops around `cross` ops.
More work is needed to make the pass work with other ops.

### DIFF
--- a/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
+++ b/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
@@ -149,7 +149,7 @@ struct PushDuplicateThroughCrossJoinPattern : public OpRewritePattern<CrossOp> {
       idx += numLeftIndices;
 
     if (!leftHasDuplicates && !rightHasDuplicates)
-      // Note: if we end up failing here, then both invokations of
+      // Note: if we end up failing here, then both invocations of
       // `createDeduplicatingEmit` returned without creating a new (`emit`) op.
       return rewriter.notifyMatchFailure(
           op, "none of the 'emit' inputs have duplicates");
--- a/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
+++ b/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
@@ -1,0 +1,185 @@
+//===- EmitDeduplication.cpp - Impl. of emit deduplication ------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include "structured/Dialect/Substrait/Transforms/Passes.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "structured/Dialect/Substrait/IR/Substrait.h"
+
+namespace mlir::substrait {
+#define GEN_PASS_DEF_SUBSTRAITEMITDEDUPLICATIONPASS
+#include "structured/Dialect/Substrait/Transforms/Passes.h.inc"
+} // namespace mlir::substrait
+
+using namespace llvm;
+using namespace mlir;
+using namespace mlir::substrait;
+
+namespace {
+
+struct SubstraitEmitDeduplicationPass
+    : substrait::impl::SubstraitEmitDeduplicationPassBase<
+          SubstraitEmitDeduplicationPass> {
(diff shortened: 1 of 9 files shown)


lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -0,0 +1,185 @@
...
+  void runOnOperation() override;
+};
+
+void SubstraitEmitDeduplicationPass::runOnOperation() {
+  mlir::RewritePatternSet patterns(&getContext());
+  populateEmitDeduplicationPatterns(patterns);
+  if (failed(mlir::applyPatternsAndFoldGreedily(getOperation(),
+                                                std::move(patterns)))) {
+    Location loc = getOperation()->getLoc();
+    emitError(loc) << "emit deduplication: pattern application failed";
```
jpienaar, line 41: I beliece this only fails if it fails to reach convergence after N iterations. Is that sufficient to fail the pass pipeline post?

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -0,0 +1,185 @@
...
+/// fields after deduplication and whether the `input` was deduplicated.
+std::tuple<Value, int64_t, bool>
+createDeduplicatingEmit(Value input, SmallVector<int64_t> &reverseMapping,
+                        PatternRewriter &rewriter) {
+  // Handles the bases cases where the input either has no `emit` op or an
+  // `emit` op with no duplicates. In that case, the returned value is just the
+  // `input` and the reverse mapping is just the identity.
+  auto handleNoDuplicates = [&]() {
+    int64_t numInputFields = cast<TupleType>(input.getType()).getTypes().size();
+    for (auto i : iota_range<int64_t>(0, numInputFields,
```
jpienaar, line 61: You can also use llvm::seq

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -0,0 +1,185 @@
...
+    // Right input: the reverse mapping of the right input needs to be adjusted
+    // by the number of deduplicated fields in the left input.
+    int64_t numLeftOriginalindices = reverseMapping.size();
+    auto [newRightInput, numRightIndices, rightHasDuplicates] =
+        createDeduplicatingEmit(op.getRight(), reverseMapping, rewriter);
+    for (int64_t &idx : drop_begin(reverseMapping, numLeftOriginalindices))
+      idx += numLeftIndices;
+
+    if (!leftHasDuplicates && !rightHasDuplicates)
+      // Note: if we end up failing here, then both invokations of
```
jpienaar, line 153: invocations

jpienaar (approved):
So this is rather specific pattern, e.g., this is not through arbitrary ops. It is for emit and cross only to consider here?

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
ingomueller-net: Yes, it is. The function returns `failure` if the pattern application did not converge after the given number of iterations. I am not 100% sure if that's the *only* reason to fail but it is the only one mentioned in the doc and thus seems to be the *typical* reason for failure:

    /// The greedy rewrite may prematurely stop after a maximum number of
    /// iterations, which can be configured in the configuration parameter.
    ...
    /// Returns "success" if the iterative process converged (i.e., fixpoint was
    /// reached) and no more patterns can be matched within the region.

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
ingomueller-net: Oh! So much better! Thanks you :)

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
ingomueller-net: Fixed.

ingomueller-net (commented):
Thanks for the review! All issues addressed with the latest commits :)

> So this is rather specific pattern, e.g., this is not through arbitrary ops. It is for emit and cross only to consider here?

Yes, the *current pattern* is specific. But the pass will eventually cover all ops that can interact with `emit`. The logic exactly how they interact is quite specific to the other op, so I don't currently see a way to implement it in any other way that one pattern or even one set of patterns per op...

**Edit:** #832 and #836 do exactly that for the `filter` and `project` ops, respectively, which covers all ops we currently have. (We can't have an `emit` feeding into a `named_table` because the latter does not have any inputs.)

PR #832, opened 2024-05-24
[Substrait] Extend emit deduplication to FilterOp. 

ingomueller-net:
~~This PR depends on and, therefor, includes #831 and its dependencies.~~

This adds a new pattern that makes emit deduplication to `filter` ops by
adding a new pattern. This pattern is similar to the previous one in
that it adds a "deduplicating" `emit` op in front of the root `filter`
op and one that re-establishes the original order after it but has the
additional complexity of having to update the region to work on the
deduplicated element type as well.

### DIFF
--- a/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
+++ b/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
@@ -230,7 +230,8 @@ struct PushDuplicatesThroughFilterPattern : public OpRewritePattern<FilterOp> {
 
   LogicalResult matchAndRewrite(FilterOp op,
                                 PatternRewriter &rewriter) const override {
-    if (!isa_and_present<EmitOp>(op.getInput().getDefiningOp()))
+    auto emitOp = op.getInput().getDefiningOp<EmitOp>();
+    if (!emitOp)
       return rewriter.notifyMatchFailure(
           op, "input operand is not produced by an 'emit' op");
 
@@ -253,7 +254,6 @@ struct PushDuplicatesThroughFilterPattern : public OpRewritePattern<FilterOp> {
                                 newOp.getCondition().end());
 
     // Update the `condition` region.
-    auto emitOp = op.getInput().getDefiningOp<EmitOp>();
     deduplicateRegionArgs(newOp.getCondition(), emitOp.getMapping(),
                           newInput.getType(), rewriter);
 
--- a/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
+++ b/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
@@ -116,6 +116,60 @@ createDeduplicatingEmit(Value input, SmallVector<int64_t> &reverseMapping,
   return {newEmitOp, newInputMapping.size(), true};
 }
 
+/// Deduplicates the fields of the region with a single `Tuple` argument using
+/// the provided (deduplicating) mapping. This involves changing the type of the
+/// region argument to the provided `newElementType`, which must be the type
+/// obtained by applying deduplication to the argument type of the provided
+/// `region`, as well as changing all `field_reference` ops using the region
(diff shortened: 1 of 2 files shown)


lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -166,14 +220,61 @@ struct PushDuplicatesThroughCrossPattern : public OpRewritePattern<CrossOp> {
...
+/// the `filter` op. This works by introducing a new `emit` op without the
+/// duplicates, creating a new `filter` op updated to work on the deduplicated
+/// element type, and finally a new `emit` op that maps back to the original
+/// order.
+struct PushDuplicatesThroughFilterPattern : public OpRewritePattern<FilterOp> {
+  using OpRewritePattern<FilterOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(FilterOp op,
+                                PatternRewriter &rewriter) const override {
+    if (!isa_and_present<EmitOp>(op.getInput().getDefiningOp()))
```
jpienaar, line 233: Should we dyn_cast here and reuse result below?

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -166,14 +220,61 @@ struct PushDuplicatesThroughCrossPattern : public OpRewritePattern<CrossOp> {
...
+      return rewriter.notifyMatchFailure(
+          op, "input operand is not produced by an 'emit' op");
+
+    // Create input ops for the new `filter` op. These may be the original
+    // inputs or `emit` ops that remove duplicates.
+    SmallVector<int64_t> reverseMapping;
+    auto [newInput, numDedupIndices, hasDuplicates] =
+        createDeduplicatingEmit(op.getInput(), reverseMapping, rewriter);
```
jpienaar, line 241: Is this create a non-op unless hasDuplicates?

test/Transforms/Substrait/emit-deduplication.mlir
```diff
@@ -149,3 +149,48 @@ substrait.plan version 0 : 42 : 1 {
...
+// CHECK-NEXT:        %[[V8:.*]] = field_reference %[[ARG0]]{{\[}}[0]] : [[TYPE]]
+// CHECK-NEXT:        %[[V9:.*]] = field_reference %[[ARG0]]{{\[}}[2]] : [[TYPE]]
+// CHECK-NEXT:        %[[Va:.*]] = func.call @f(%[[V3]], %[[V4]], %[[V5]], %[[V7]], %[[V8]], %[[V9]])
+// CHECK-NEXT:        yield %[[Va]] : si1
+// CHECK-NEXT:      }
+// CHECK-NEXT:      %[[Vb:.*]] = emit [0, 0, 1, 0, 2] from %[[V2]]
+
+func.func private @f(si1, si1, si1, si32, si1, si1) -> si1
+
+substrait.plan version 0 : 42 : 1 {
```
jpienaar, line 177: Could you add a short explanation here?

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
ingomueller-net: Yep -- much better! Done.

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
ingomueller-net: Indeed, see docstring or note below: if `hasDuplicates == true`, then the call to `createDeduplicatingEmit` has no effect.

test/Transforms/Substrait/emit-deduplication.mlir
ingomueller-net: Done.

ingomueller-net (commented):
Good suggestions -- thanks!

PR #836, opened 2024-05-29
[Substrait] Extend emit deduplication to project op. 

ingomueller-net:
~~This PR depends on and, therefor, contains #833 and #832 and their dependencies.~~

This pushes duplicate fields from `emit` ops producing the operand of a
`project` op to an emit op on the result of the `project`.

### DIFF
--- a/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
+++ b/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
@@ -266,6 +266,71 @@ struct PushDuplicatesThroughFilterPattern : public OpRewritePattern<FilterOp> {
   }
 };
 
+/// Pushes duplicates in the mappings of `emit` ops producing the input through
+/// the `filter` op. This works by introducing a new `emit` op without the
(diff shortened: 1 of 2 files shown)


jpienaar (approved):

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -116,6 +116,60 @@ createDeduplicatingEmit(Value input, SmallVector<int64_t> &reverseMapping,
...
+      int64_t pos = indexPositions.size();
+      auto [it, success] = indexPositions.try_emplace(index, pos);
+      oldToNewPositions.push_back(it->second);
+    }
+  }
+
+  // Update field references using the region argument.
+  for (Operation *user : region.getArgument(0).getUsers()) {
+    // We are only interested in `field_reference` ops.
+    if (!isa<FieldReferenceOp>(user))
```
jpienaar, line 153: dyn_cast instead of isa followed by cast

ingomueller-net: Done!

PR #838, opened 2024-05-29
[Substrait] Run CSE in emit deduplication patterns. 

ingomueller-net:
This PR depends on and, therefor, includes #837 and #836 and their dependencies.

This commits adds calls to two patterns that may create common
subexpressions during emit deduplication. By deduplicating their
regions, field references to duplicated fields end up referring to the
same field, thus creating common subexpressions. The patterns, thus
call CSE right away. An alternative would be to call CSE in one or
several dedicated patterns but that would either require (1) to test for
duplicate field references before application or (2) run CSE blindly and
return `success()` iff that made changes.

### DIFF
--- a/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
+++ b/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
@@ -8,7 +8,9 @@
 
 #include "structured/Dialect/Substrait/Transforms/Passes.h"
 
+#include "mlir/IR/Dominance.h"
 #include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/CSE.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "structured/Dialect/Substrait/IR/Substrait.h"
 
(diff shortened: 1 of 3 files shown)


lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -257,6 +259,13 @@ struct PushDuplicatesThroughFilterPattern : public OpRewritePattern<FilterOp> {
     deduplicateRegionArgs(newOp.getCondition(), emitOp.getMapping(),
                           newInput.getType(), rewriter);
 
+    // Deduplicating block args may create common subexpressions. Eliminate
+    // them immediately.
+    {
+      DominanceInfo domInfo;
+      mlir::eliminateCommonSubExpressions(rewriter, domInfo, newOp);
```
jpienaar, line 266: This seems like it could be expensive to do inside a pattern. But not blocking, we can re-evaluate later.

jpienaar (approved):

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
ingomueller-net: I have thought a bit about this: I think that this is the place with the lowest algorithmic complexity. The reason is the following: during duplicate elimination of the ops that produce the inputs of this op, some arguments into this region may be deduplicated, i.e., get fused. This, in turn, may create common sub-expressions since `field_references` that were previously referring to two different fields now refer to the same (deduplicated) field. And that, in turn, may allow to deduplicate the *outputs* of this op by eliminating these common sub-expressions using CSE, which are then interesting for the next op.

If we don't run CSE *inside* the op, we could have to run the CSE and emit deduplication passes as many times as the use-def tree is high because each pass could unlock something that the other pass could clean up. That way, we would have to go over the whole IR *every time* we repeat that cycle. With the current approach, emit deduplication goes over it once and CSE goes over the inner regions (only!) once every time inputs of that op got deduplicated (and only then!).

Let me document that somewhere in the code...

ingomueller-net: (Updated git commit message.)

PR #846, opened 2024-05-31
[Substrait] Add pattern removing duplicate yields in project. 

ingomueller-net:
~~This PR depends on and, therefor, includes #836 and its dependencies.~~

This PR adds a pattern to the emit deduplication pass that removes
duplicates introduced by the `project` op by yielding values more than
once.

### DIFF
--- a/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
+++ b/lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
@@ -172,6 +172,76 @@ void deduplicateRegionArgs(Region &region, ArrayAttr newMapping,
   region.getArgument(0).setType(newElementType);
 }
 
+struct EliminateDuplicateYieldsInProjectPattern
+    : public OpRewritePattern<ProjectOp> {
+  using OpRewritePattern<ProjectOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(ProjectOp op,
+                                PatternRewriter &rewriter) const override {
+    MLIRContext *context = op.getContext();
+    Operation *terminator = op.getExpressions().front().getTerminator();
+    int64_t numOriginalYields = terminator->getNumOperands();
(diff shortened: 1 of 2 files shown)


lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
```diff
@@ -170,6 +170,76 @@ void deduplicateRegionArgs(Region &region, ArrayAttr newMapping,
...
+struct EliminateDuplicateYieldsInProjectPattern
+    : public OpRewritePattern<ProjectOp> {
+  using OpRewritePattern<ProjectOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(ProjectOp op,
+                                PatternRewriter &rewriter) const override {
+    MLIRContext *context = op.getContext();
+    Operation *terminator = op.getExpressions().front().getTerminator();
+    int64_t numOriginalYields = terminator->getNumOperands();
+    auto inputTupleType = cast<TupleType>(op.getInput().getType());
```
jpienaar, line 182: And this is verified for ProjectOp that it has only TupleType inputs?

jpienaar (approved):

lib/Dialect/Substrait/Transforms/EmitDeduplication.cpp
ingomueller-net: Yes. ([`input`](https://github.com/iree-org/iree-llvm-sandbox/blob/5acf08e/include/structured/Dialect/Substrait/IR/SubstraitOps.td#L388) is of type `Substrait_Relation`, which is a placeholder for a to-be-implemented relation type that is [currently](https://github.com/iree-org/iree-llvm-sandbox/blob/5acf08e/include/structured/Dialect/Substrait/IR/SubstraitTypes.td#L58) a (nested) tuple.

In these cases, I do not need to do anything else that I do here, right?

PR #852, opened 2024-07-15
[Substrait] Write README for Substrait dialect.

ingomueller-net:


README-Substrait.md
```diff
@@ -0,0 +1,173 @@
+# Substrait Dialect for MLIR
+
+This project consist of building an input/output dialect in
+[MLIR](https://mlir.llvm.org/) for [Substrait](https://substrait.io/), the
+cross-language serialization format of database query plans (akin to an
+intermediate representation/IR for database queries). The immediate goal is to
+create common infrastructure that can be used to implement consumers, producers,
+optimizers, and transpilers of Substrait; the more transcending goal is to study
```
jpienaar, line 8: The transcending goal feels to local. Using compilation techniques to optimize database queries?

README-Substrait.md
```diff
@@ -0,0 +1,173 @@
...
+the viability of using MLIR to implement database query compilers.
+
+## Motivation
+
+Substrait defines a serialization format for data-intensive compute operations
+similar to relational algebra as they typically occur in database query plans
+and similar systems, i.e., the "intermediate representation" (or IR) of database
```
jpienaar, line 15: It doesn't feel like an IR in that sense, more like an interop, it's a common interchange format.

README-Substrait.md
```diff
@@ -0,0 +1,173 @@
...
+it aims to build an *in-memory* format for the concepts defined by Substrait,
+for which the latter only describe their *serialization format*. MLIR is a
+generic compiler framework providing infrastructure for writing compilers from
+any domain and is part of the LLVM ecosystem. It makes it easy to add new IR
+consisting of domain-specific operations, types, attributes, etc., which are
+organized in dialects (either in-tree and out-of-tree), as well as rewrites,
+passes, conversions, translations, etc. on those dialects. Creating a Substrait
+dialect and a number of common related transformations in such a mature
+framework has the potential to eliminate some of the repeated effort described
+above and, thus, to ease and eventually increase adoption of Substrait. By
```
jpienaar, line 44: I don't know where, but could mention it has an active community with multiple ML frameworks using and usage as varied as homomorphic encryption and abuse detection.

jpienaar (approved):
Very nice thanks

README-Substrait.md
ingomueller-net: You mean "using MLIR to implement" something isn't broad/ambitious/transcendental enough? How about referring to one of the properties of MLIR then? "using a generic compiler framework (like MLIR)" or "using a progressive lowering approach"?

"Using compilation techniques to optimize database queries" makes me think in different directions: one is the old textbook-style "query compilation," which DB systems do basically since they were invented, and the other one is to use code generation (through LLVM), which some DB systems have been doing for the last 10-15 years.

Maybe "using modern, general-purpose compiler infrastructure to ..."? "Modern" referring to generic, modular, based on progressive lowering, ... as pioneered/popularized by MLIR? Also, I think "compiler infrastructure" is minimally more different from "compilation" and thus associates less with the textbook DB compilers, at least in my ears.

README-Substrait.md
ingomueller-net: Yeah, true. Changed here and added an introduction of the term "intermediate representation (IR)" below.

README-Substrait.md
ingomueller-net: Good idea! I added a mention to MLIR community and wide adoption with links to discourse and https://mlir.llvm.org/users/.

ingomueller-net (commented):
Cool, thanks! Comments addressed by latest commit.

PR #855, opened 2024-07-19
[Substrait] Implement `CallOp` for `ScalarFunction` message.

ingomueller-net:
~~This PR depends on and, therefor, includes #853.~~

This allows to use the externally declared functions from #853 as scalar
function calls (which have the semantics of a typical function call).
The current design anticipates that aggregate and window functions can
be modelled with the same op, but future PR will need to show if and how
that is possible.

### DIFF
--- a/include/structured/Dialect/Substrait/IR/SubstraitOps.td
+++ b/include/structured/Dialect/Substrait/IR/SubstraitOps.td
@@ -297,6 +297,43 @@ def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
   let assemblyFormat = "$value attr-dict";
 }
 
+def Substrait_CallOp : Substrait_ExpressionOp<"call", [
+    DeclareOpInterfaceMethods<SymbolUserOpInterface>,
+  ]> {
+  let summary = "Function call expression";
+  let description = [{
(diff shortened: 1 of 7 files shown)


include/structured/Dialect/Substrait/IR/SubstraitOps.td
```diff
@@ -296,6 +296,43 @@ def Substrait_LiteralOp : Substrait_ExpressionOp<"literal", [
...
+    DeclareOpInterfaceMethods<SymbolUserOpInterface>,
+  ]> {
+  let summary = "Function call expression";
+  let description = [{
+    Represents a `ScalarFunction` message (or, in the future, other `*Function`
+    messages) together with all messages it contains and the `Expression`
+    message it is contained in.
+
+    Currently, the specification of the function, which is in an external YAML
+    file,  is not taken into account, for example, to verify whether a matching
```
jpienaar, line 309: single space?

jpienaar (approved):

include/structured/Dialect/Substrait/IR/SubstraitOps.td
ingomueller-net: Done.

PR #862, opened 2024-11-22
[CI] Fix installing apt dependencies by using action.

ingomueller-net:
This PR how CI installs dependencies via apt: the previous code simply ran apt-get install without prio apt-get update. This may make CI break, which happened for some PRs, if the GH runner has a version of a package in cache that just got updated such that the cached URL is invalid and the package cannot be installed. By using a proper GH action, not only is that taken care of but the package is even cached such that installing it may be a bit faster.


ingomueller-net:
@jpienaar: Maybe the notification for this PR also got lost?

